PR #2, opened 2025-05-12
fix for windows node build

jhlee525:
1. Add libssl as a addon's dependency explicitly
2. Fix to copy dependent dll files in electron example


grf53 (approved):
LGTM

PR #3, opened 2025-05-12
Feature: Integrate MCP tools

khj809:
This PR enables using MCP tools on Ailoy via Stdio transport.
- Users can add MCP tools by `add_tools_from_mcp_server`(python) or `addToolsFromMcpServer`(js) with MCP stdio server parameters. See test codes for the detailed usages.
- Consolidated python examples in a single directory. (If it should be also applied in node examples, I can proceed to do it)
- Revised outdated readme. (will be entirely updated though)


jhlee525 (approved):
LGTM 👍 Please fix to new API names

Issue #5, opened 2025-05-13
ld warning related to target macOS version of `libtokenizers_c.a`


grf53:
## Description
The following warning comes out many times during building for macOS machines.
`[build] ld: warning: object file (../_deps/mlc-llm-build/tokenizers/aarch64-apple-darwin/debug/libtokenizers_c.a[255](a445302c6d3dcb51-unicode_fold3_key.o)) was built for newer 'macOS' version (15.2) than being linked (14.4)`

The OS version of tested machine is `14.7`, none of `15.2` or `14.4`.

`14.4` is consistent because the target version of this library for macOS is `14.4` as set at:
https://github.com/brekkylab/ailoy/blob/c52becd169e84a758a30097f075c63dd13c8b6f9/CMakeLists.txt#L38

Maybe there is some problem to set the target macOS version in `tokenizers-cpp` with `CMAKE_OSX_DEPLOYMENT_TARGET`.

jhlee525:
According to [cmake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html), the option `CMAKE_OSX_DEPLOYMENT_TARGET` seems related to minimum version of the OSX.

It makes sense that #6 can depress the warning message.

PR #6, opened 2025-05-13
build: bump `tokenizers-cpp` version as latest

grf53:
## Ticket
#5 

## Description
The latest version of `tokenizers-cpp` can set the target macOS version. (added at https://github.com/mlc-ai/tokenizers-cpp/pull/52)
But the `tokenizers-cpp` version that the `mlc-llm` is using is [4bb7533](https://github.com/mlc-ai/tokenizers-cpp/tree/4bb753377680e249345b54c6b10e6d0674c8af03) which doesn't have that.
So let the `mlc-llm` use the latest `tokenizers-cpp` [here](https://github.com/brekkylab/mlc-llm/commits/3b8165524fdb79fd3255395432ac70a7734a22ef/), and ailoy use the `mlc-llm`.

## Requirements
The macOS version setting merged into `tokenizers-cpp` later than its `huggingface/tokenizers` update to `0.21.0` (at https://github.com/mlc-ai/tokenizers-cpp/pull/62), we need to use `0.21.0` for this.
And the `0.21.0` of `huggingface/tokenizers` includes [`std::option::Option::is_none_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) which is added rust `1.8.20`, so we need rust version `>= 1.82.0`.

### DIFF
--- a/vm/CMakeLists.txt
+++ b/vm/CMakeLists.txt
@@ -70,13 +70,13 @@ if(AILOY_WITH_MLC_LLM)
             FetchContent_Declare(
                 mlc-llm
                 GIT_REPOSITORY https://github.com/brekkylab/mlc-llm.git
-                GIT_TAG 026f91859dddc10ecf96bdaf85f7f3576d73e161
+                GIT_TAG 3b8165524fdb79fd3255395432ac70a7734a22ef
             )
         else()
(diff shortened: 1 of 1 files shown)


jhlee525 (approved):
LGTM

PR #7, opened 2025-05-13
Optimize includes

grf53:
## Description
Removing unused includes, or move includes that are only used in source file into the source file.

### target directories
- core
- broker
- vm

(maybe there are some unoptimized includes in files in binding csrc directories.)

### Note
This change is only tested on macOS, so it needs to be tested on Linux and Windows environments too.
- [ ] Linux
- [ ] Windows

### DIFF
--- a/broker/include/broker_client.hpp
+++ b/broker/include/broker_client.hpp
@@ -1,9 +1,5 @@
 #pragma once
 
-#include <memory>
-#include <optional>
-#include <ostream>
-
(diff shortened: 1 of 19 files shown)


jhlee525 (commented):
It seems nice for me. After verification that msvc can compile this without error, we can merge it

grf53:
Checked in both Linux & Windows environments too.

jhlee525 (approved):
LGTM

PR #8, opened 2025-05-13
Update issue templates

jhlee525:
After reviewing other libraries, it seems that using issue templates has become quite common these days.

For now, we’ve defined a simple issue template. We can gradually improve it as we start using it.


khj809 (approved):

grf53 (approved):
👍 :fire:

PR #14, opened 2025-05-14
Fix model download

khj809:
This PR fixes the following bugs:
- Files marked to be downloaded are not actually downloaded -> Removed the invalid condition
- Files written with larger invalid bytes than it's supposed to be is not properly downloaded -> Force fresh download always


jhlee525 (approved):
👍

PR #17, opened 2025-05-14
Change agent runtime from async to sync

khj809:
As we discussed earlier, for someone who is not familiar with async programming, we decided to provide Agent APIs in a synchronous manner by default. This change is also applied to VectorStore as well.
Async version will be restored after discussing how we should provide it more smoothly.

## Notes
- MCP sdks does not provide any synchronous APIs, so it's mandatory to call them asynchronously. They're synchronously handled via `run_async` function.
- Revised tests by grouping them properly and adding fixtures to reuse resources.


jhlee525 (approved):
👍

PR #18, opened 2025-05-15
fix: minor changes from windows

grf53:
## Description
Fix for some minor issues observed in Windows environment

### DIFF
--- a/bindings/js-node/package.json
+++ b/bindings/js-node/package.json
@@ -18,7 +18,6 @@
         "build": "npm run build:napi && npm run build:ts",
         "clean": "rm -rf dist/",
         "test": "ts-mocha",
-        "test:debug": "node --inspect-brk ./node_modules/.bin/ts-mocha",
         "prepack": "clean-package",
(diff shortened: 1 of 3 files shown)


bindings/js-node/package.json
```diff
@@ -18,6 +18,7 @@
         "build": "npm run build:napi && npm run build:ts",
         "clean": "rm -rf dist/",
         "test": "ts-mocha",
+        "test:debug": "node --inspect-brk ./node_modules/.bin/ts-mocha",
```
grf53: Comment requested: not sure this is needed to be merged.

jhlee525: Currently, we are using `ts-node`(kind of ts interpreter) as a test code. Is it okay to run with just `node`?

Maybe @khj809 have idea about it.

khj809: We don't need this. It was just added for temporary test purpose.

khj809 (approved):

PR #19, opened 2025-05-15
Public api change

jhlee525:
As we discussed, we needed to make some API changes, especially to improve naming consistency between the Python and TypeScript versions.

Here are the changes I’ve worked for.

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -54,18 +54,15 @@ interface MessageDelta {
 
 /** Types for LLM Model Definitions */
 
-type TVMModel = "qwen3-8b" | "qwen3-4b" | "qwen3-1.7b" | "qwen3-0.6b";
-interface TVMModelAttrs {
-  name: TVMModel;
-  quantization?: "q4f16_1";
-  mode?: "interactive" | "local" | "server";
-}
+export type TVMModelName =
+  | "qwen3-8b"
+  | "qwen3-4b"
+  | "qwen3-1.7b"
+  | "qwen3-0.6b";
 
-type OpenAIModel = "gpt-4o";
-interface OpenAIModelAttrs {
-  name: OpenAIModel;
-  api_key: string;
-}
+export type OpenAIModelName = "gpt-4o";
+
+export type ModelName = TVMModelName | OpenAIModelName;
 
 interface ModelDescription {
   modelId: string;
@@ -73,8 +70,7 @@ interface ModelDescription {
   defaultSystemMessage?: string;
 }
 
-type AvailableModel = TVMModel | OpenAIModel;
-const modelDescriptions: Record<AvailableModel, ModelDescription> = {
+const modelDescriptions: Record<ModelName, ModelDescription> = {
   "qwen3-8b": {
     modelId: "Qwen/Qwen3-8B",
     componentType: "tvm_language_model",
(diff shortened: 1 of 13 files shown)


bindings/js-node/src/agent.ts
```diff
@@ -107,48 +103,42 @@ const modelDescriptions: Record<AvailableModel, ModelDescription> = {
...
-    | "output_text"
-    | "tool_call"
-    | "tool_call_result"
-    | "reasoning"
-    | "error";
-  endOfTurn: boolean;
-  role: "assistant" | "tool";
-}
-interface AgentResponse_OutputText extends AgentResponse_Base {
+export interface AgentResponseText {
```
grf53, line 106: Decided to export the detailed types & interfaces?

bindings/js-node/src/agent.ts
```diff
@@ -202,7 +188,7 @@ interface RESTAPIToolDefinition extends BaseToolDefinition {
   };
 }
```
grf53, line 189: Looks good, it became looking more intuitive.

bindings/js-node/src/agent.ts
```diff
@@ -246,60 +232,97 @@ export function bearerAutenticator(
...
   private runtime: Runtime;
-  private modelInfo: {
-    componentType: string;
-    componentName: string;
-    attrs: Record<string, any>;
+
+  private componentState: {
+    name: string;
+    valid: boolean;
   };
```
grf53, line 239: `modelInfo` is replaced by `componentState` in node, but `_model_info` in python still exists.
Would it be the same soon, or is there another plan?

khj809 (approved):

bindings/js-node/src/agent.ts
jhlee525: The rationale behind introducing componentState is to clearly separate the roles of each function:

__init__ (cannot be async): Initializes resources.

define (can be async): Calls define_component on the runtime.

With this in mind, I felt that passing parameters to the constructor and storing them just to be used later in define would result in unnecessary state retention. Therefore, the class should hold only the minimal information needed to communicate with the runtime.

As you pointed out, introducing a `component_state` Python as well seems appropriate. I'll work for it

bindings/js-node/src/agent.ts
jhlee525: I think it's not essential but it can be exported(for convenience of typescript users)

jhlee525:
@grf53 @khj809 I removed kwargs parameters in `Agent` API. Could you check?

bindings/python/ailoy/agent.py
```diff
@@ -322,7 +324,7 @@ def __init__(
         self._tools: List[Tool] = []
 
         # Define can be performed in constructor
-        self.define(model_name, **kwargs)
+        self.define(model_name, api_key, attrs)
```
khj809: I think delivering keyword parameters with their names explicity would make the code less ambiguous.
```
self.define(model_name, api_key=api_key, attrs=attrs)
```
Or just make sure they're keyword parameters.
```
self.define(model_name, *, api_key, attrs)
```

jhlee525: Ok I'll fix it. But please note that this is internal call from `__init__` to `define` so the ambiguous is not the critical problem.

khj809 (approved):

bindings/js-node/src/agent.ts
grf53: Checked that it is in Python too.

examples/simple_rag_electron/src/ipc.ts
```diff
@@ -13,11 +13,7 @@ export const initializeComponents = async (mainWindow: BrowserWindow) => {
   }
 
   if (vectorstore === undefined) {
-    vectorstore = new ailoy.VectorStore(runtime, {
-      type: "faiss",
-      embedding: "bge-m3",
-    });
-    await vectorstore.initialize();
+    vectorstore = await ailoy.defineVectorStore(runtime, "bge-m3", "faiss");
```
grf53, line 16: Here I think the latest changes are not reflected.

grf53 (approved):
LGTM 👍

bindings/python/ailoy/agent.py
```diff
@@ -294,95 +292,103 @@ class Agent:
...
+        # Initialize messages
+        self._messages: List[Message] = []
+        if system_message:
             self._messages.append(SystemMessage(role="system", content=system_message))
 
-        self._tools: List[Tool] = tools if tools is not None else []
+        # Initialize tools
+        self._tools: List[Tool] = []
+
+        # Define can be performed in constructor
```
grf53: This comment looks a little redundant.

jhlee525: Fixed

PR #25, opened 2025-05-16
make cache root & r2 base url configurable

jhlee525:
Previously, hard-coded `CACHE_ROOT` and `R2_BASE_URL` controls remote & local cache roots.

This PR makes the user can control the path of these cache-related directories, via environment variable

- `AILOY_CACHE_ROOT`: Can change local cache root
- `AILOY_REMOTE_CACHE_BASE_URL`: Can change remote file server

### DIFF
--- a/vm/src/mlc_llm/model_cache.cpp
+++ b/vm/src/mlc_llm/model_cache.cpp
@@ -301,8 +301,7 @@ get_model(const std::string &model_name, const std::string &quantization,
   for (auto &[file, sha1] :
        manifest["files"]
            .get<std::vector<std::pair<std::string, std::string>>>()) {
-    // Add to files_to_download if neither file exists nor sha1 checksum is
-    // same
+    // Add to files_to_download if neither file exists nor sha1 checksum is same
     if (!(fs::exists(model_cache_path / file) &&
           sha1 == sha1_checksum(model_cache_path / file))) {
       files_to_download.emplace_back(file);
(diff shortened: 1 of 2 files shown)


khj809 (approved):
Looks good, but I think `AILOY_REMOTE_CACHE_BASE_URL` has broader meaning than just for models. Maybe we can either replace the name of env with `AILOY_MODELS_URL`, or make the R2 bucket to serve not only models but also another assets.

jhlee525:
`AILOY_MODELS_URL` seems good. R2 is the vendor so I think more neutral name would be fine

vm/src/mlc_llm/model_cache.cpp
```diff
@@ -274,7 +301,8 @@ get_model(const std::string &model_name, const std::string &quantization,
   for (auto &[file, sha1] :
        manifest["files"]
            .get<std::vector<std::pair<std::string, std::string>>>()) {
-    // Add to files_to_download if neither file exists nor sha1 checksum is same
+    // Add to files_to_download if neither file exists nor sha1 checksum is
+    // same
```
grf53: Maybe it's because of the auto formatting, but this looks a little weird.
A tiny adjustment would be good here.

grf53 (approved):
Good to have this!
Good to be merged after the tiny adjustment in the comment.

vm/src/mlc_llm/model_cache.cpp
jhlee525: I have no idea why it is changed. I'll restore it.

PR #28, opened 2025-05-16
Deprecated typing

grf53:
## Description
Some typing classes (like `List`, `Tuple`, etc.) are deprecated and replaced by pre-existing classes from python 3.9.
If we don't need to support python 3.8 or below(these versions are in end-of-life.), we had better not to use those classes.
So this PR just replace those with the existing classes

## Note
`typing.Union[T, U]` and `typing.Optional[T] == typing.Union[T, None]` can be replaced with `T | U` and `T | None` from python 3.10. But we still need to python 3.9 and these are not even deprecated, so those would remains.

### DIFF
--- a/bindings/python/ailoy/agent.py
+++ b/bindings/python/ailoy/agent.py
@@ -2,21 +2,14 @@ import json
 import subprocess
 import warnings
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


khj809 (approved):
👍

PR #29, opened 2025-05-16
Pass function as authenticator

grf53:
## Description
During writing the example code for NY Times tool, I realized that there is no handy way to authenticate the REST requests through the query params in python. (in node.js, anonymous class object can cover this.)

So I changed functions to get the proper 'callable' instead of `ToolAuthenticator`. 
I just added `__call__()` into the `ToolAuthenticator` to make it callable, so existing `ToolAuthenticator` is still available.
```python
# Apply authentication
if callable(authenticator):
    request = authenticator(request)
```

To apply this on nodejs, It is just using the function.

Request for comments on this kind of changes on API!

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -190,36 +190,32 @@ export interface ToolDefinitionRESTAPI {
 
 export type ToolDefinition = ToolDefinitionUniversal | ToolDefinitionRESTAPI;
 
-export interface ToolAuthenticator {
-  apply: (request: {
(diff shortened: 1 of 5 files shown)


jhlee525 (approved):
👍 Cool

grf53:
@jhlee525 Re-requested review to have the reviews on nodejs runtime too.

khj809 (approved):

jhlee525 (approved):
Ok let's go with this schema

PR #30, opened 2025-05-17
Align module implementations with API references

khj809:
Some operators and methods do not align with API references written in https://github.com/brekkylab/ailoy-docs/pull/9, so they're fixed in this PR as below.
- Operators and components for debugging purpose (`echo`, `spell`, `accumulator`) were migrated to separate "debug module", and they now expect a map type input with `text` field instead of getting string input directly.
- <del>Vectorstore's `insert_many` operator now expects a map type input with `items` field instead of getting array input directly.</del> Also, in `retrieve` operator, renamed `k` to `top_k` for clarity.
- Fixed tests and examples not aligned with the latest updates.


jhlee525 (approved):
👍 👍

grf53 (approved):

PR #33, opened 2025-05-18
update tokenizer cpp

jhlee525:
Since my patch(about mt/md) merged into `mlc-ai/tokenizers-cpp`, we can use official tokenizers-cpp.

This PR is paired with https://github.com/brekkylab/mlc-llm/pull/3


khj809 (approved):
👍

grf53 (approved):

PR #39, opened 2025-05-19
Capture sigint while downloading models

### LINKED ISSUE #21: Ensure graceful shutdown on SIGINT
Currently, we are frequently encountering hangs when a SIGINT is received.
It seems necessary to ensure that the program terminates gracefully when the console is interrupted (e.g., via Ctrl+C) on Windows, macOS, and Linux.

- [ ] macOS / python
- [ ] macOS / node
- [ ] Linux / python
- [ ] Linux / node
- [ ] Windows / python
- [ ] Windows / node

khj809:
- Added `SigintCatcher` which registers sigint handler and restores it via RAII pattern.
- `download_file_with_progress()` initializes the above `SigintCatcher` at the beginning, and returns the corresponding error message immediately when SIGINT has been detected.
- `get_model` now returns a struct `model_cache_get_result_t` which stores `success` and optional `error_message`. The users of `get_model` (currently including `tvm_model_t` and `mlc_llm_engine_t`) are responsible for handling errors when `success` is false. Components (`tvm_embedding_model` and `tvm_language_model`) may return `error_output_t` if such exception has been catched.
- Added a test case `SigintWhileGetModel` for testing SIGINT situations.
- Decided to show progress bars by default, which looks better for user experiences.

Verified that users can do ctrl+c while downloading models, and this results in `RuntimeError` instead of abort.
```python
>>> rt.define("tvm_language_model", "lm0", {"model": "Qwen/Qwen3-0.6B"})
Downloading params_shard_0.bin [██████▊                                           ] 13% [00m:00s]                                               
^CTraceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    rt.define("tvm_language_model", "lm0", {"model": "Qwen/Qwen3-0.6B"})
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/haejoonkim/Desktop/workspace/BrekkyLab/ailoy/bindings/python/ailoy/runtime.py", line 147, in define
    raise RuntimeError(packet["body"]["reason"])
RuntimeError: Interrupted while downloading the model
```

Note that this was tested on MacOS only, but I believe it works well on Linux too. Not sure for Windows.

### DIFF
--- a/vm/src/mlc_llm/embedding_model.cpp
+++ b/vm/src/mlc_llm/embedding_model.cpp
@@ -235,8 +235,13 @@ create_tvm_embedding_model_component(std::shared_ptr<const value_t> inputs) {
     return outputs;
   };
 
-  auto tvm_embedding_model =
-      create<tvm_embedding_model_t>(model_name, quantization);
+  std::shared_ptr<tvm_embedding_model_t> tvm_embedding_model;
+  try {
+    tvm_embedding_model =
+        create<tvm_embedding_model_t>(model_name, quantization);
+  } catch (const ailoy::runtime_error e) {
+    return ailoy::error_output_t(e.what());
+  }
 
   auto tokenizer = create<tokenizer_t>(tvm_embedding_model->get_model_path() /
                                        "tokenizer.json");
--- a/vm/src/mlc_llm/language_model.cpp
+++ b/vm/src/mlc_llm/language_model.cpp
@@ -103,7 +103,13 @@ create_tvm_language_model_component(std::shared_ptr<const value_t> inputs) {
 #endif
 
   // Get engine
-  auto engine = get_mlc_llm_engine(model, quantization, device, mode);
+  std::shared_ptr<mlc_llm_engine_t> engine;
+  try {
+    engine = get_mlc_llm_engine(model, quantization, device, mode);
+  } catch (const ailoy::runtime_error e) {
+    return error_output_t(e.what());
+  }
+
   if (engine->get_last_error().has_value())
     return error_output_t(engine->get_last_error().value());
 
--- a/vm/src/mlc_llm/mlc_llm_engine.cpp
+++ b/vm/src/mlc_llm/mlc_llm_engine.cpp
@@ -21,14 +21,19 @@ mlc_llm_engine_t::mlc_llm_engine_t(const std::string &model_name,
   auto device_type_str = tvm::runtime::DLDeviceType2Str(device.device_type);
 
   // Download model
-  auto [model_path, model_lib_path] =
+  model_cache_get_result_t get_model_result =
       get_model(model_name, quantization, device_type_str);
-  model_path_ = model_path;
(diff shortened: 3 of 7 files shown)


jhlee525:
Currently we have `sigint` handler in core https://github.com/brekkylab/ailoy/blob/c3b9a2506a16fce06921e1bd661e541bc3cf78bd/core/include/thread.hpp#L210

It is used in broker:
https://github.com/brekkylab/ailoy/blob/main/broker/src/broker.cpp#L53

It also used in vm:
https://github.com/brekkylab/ailoy/blob/c3b9a2506a16fce06921e1bd661e541bc3cf78bd/vm/src/vm.cpp#L18

The rough explanation
SIGINT -> captured by `stop_t::global_stop` -> notify to monitor -> monitor captures event -> hanle stops

Can these be incorporated with previous stop logic?

khj809:
@jhlee525 I replaced it with `global_stop` as you mentioned. PTAL.

jhlee525:
If it works in both Python and Node, we can consider it resolved.

Unfortunately, there seem to be some issues with how signals are handled in Python and Node:

- Both Python and Node have their own signal handlers.
- global_stop overrides these handlers.
- As a result, the original signal handlers in Python and Node are lost.
- Therefore, when a SIGINT actually occurs, Python and Node do not terminate properly and instead stall.

This hasn't been thoroughly tested, so I might be wrong.

jhlee525:
Could you please explain the situation in more detail?

From my understanding, when a SIGINT is received, it terminates the entire process, including Python and the Runtime.
However, the current behavior seems different.  Could you clarify how it behaves (or you desired)?

khj809:
@jhlee525 What I expected is just stopping the model download function, not the entire process. When the function is running on Python REPL, we expect pressing ctrl+c just stops that function with KeyboardInterrupt, not terminating the REPL itself.

jhlee525:
Hmm... Then we should not stop the whole runtime if ctrl+c raised. It must break ailoy's runtime and python will cannot use these functions.

- Running python in REPL: Do not stop the runtime (the command will continue)
- Running python in script: Ctrl+C must stop the runtime

Is it right? It might be complex then I expected.

jhlee525:
This is partial solution only can apply to downloading models.

We'll handle whole sigint handlling in the following works.

This is an outline for handling this requirements:

- System stop flag: when enabled, The whole system will be stopped. `vm` or `broker` should bind this flag and obliged to stop immediately.
- Routine stop flag: Stops long routine (like download, infer): if this flag raised, this job should be terminated with an error.
- In cpp: introduces two kinds of stop system, and attach handler for this. A function can toggle these flags also should be implemented.
- In python or node: determines the situation whether to raise stop system or just a routine. Raises a proper signal.

@khj809 Does it makes sense?

jhlee525:
All right. I'll do it

PR #40, opened 2025-05-19
Improve/Refactor APIs

grf53:
## Changes
Changes are mainly on Python APIs.
- The inner functions for the tools no longer get the `Runtime`, but use `self._runtime` when they need.
- The inner functions for the tools gets 'unpacked' parameters(`**kwargs`) instead of a single input parameter of  `Any` type.
- Test for python function tool is added.
- Add context_manager special functions(`__enter__`, `__exit__`) for Runtime to enable `with` statement.
- Add a little more accurate validations and error message for the parameter of `VectorStore`
- Rename `agent.run()` as `agent.query()`
- The default value of `enable_reasoning` and `ignore_reasoning_message` are `False` instead of `None` for `agent.query()`.

Other changes
- The inner functions for the tools no longer get the `Runtime`, but use `this.runtime` when they need (Node.js).
- Rename `agent.run()` as `agent.query()` (Node.js)
- fix `ex` -> `agent` for `Agent` instances
- some minor fixes

### DIFF
--- a/examples/movie_agent_js_node/src/index.ts
+++ b/examples/movie_agent_js_node/src/index.ts
@@ -44,7 +44,7 @@ async function main() {
     if (query === "" || query === "exit") break;
 
     process.stdout.write(`\nAssistant: `);
-    for await (const resp of agent.run(query, { enableReasoning: true })) {
+    for await (const resp of agent.query(query, { enableReasoning: true })) {
       if (resp.type === "output_text") {
         process.stdout.write(resp.content);
         if (resp.endOfTurn) {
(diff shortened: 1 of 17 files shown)


examples/movie_agent_js_node/src/index.ts
```diff
@@ -44,7 +44,7 @@ async function main() {
     if (query === "" || query === "exit") break;
 
     process.stdout.write(`\nAssistant: `);
-    for await (const resp of ex.run(query, { enableReasoning: true })) {
+    for await (const resp of agent.run(query, { enableReasoning: true })) {
```
jhlee525:
```suggestion
    for await (const resp of agent.query(query, { enableReasoning: true })) {
```

grf53: I only considered Python APIs at first, so I missed this.
Now `agent.run()` became `agent.query()`

jhlee525 (approved):
👍

PR #41, opened 2025-05-19
refactor: rename `universal tool` -> `builtin tool`

grf53:
## Description
As we decided to call the 'universal tool' as 'built-in tool', renaming it.

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -170,8 +170,8 @@ export interface ToolDescription {
   };
 }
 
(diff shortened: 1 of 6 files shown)


khj809 (commented):
I think it's not necessary to put `-` in the name. For example, Python also uses module name `builtins`.

jhlee525:
Why is it closed?

grf53:
> Why is it closed?

Sorry, I misunderstood the message. It is now re-opened and ready for reviews.

jhlee525 (approved):

PR #43, opened 2025-05-20
remove global stop

jhlee525:
While I'm digging into #21 , I found that the default signal handler is enough to handle this issue.

Just keep default signal handler will stop all thread properly.

So I removed global_stop concepts from the library.

One execption is while downloading something(`recv()` call). `sa_restart` is enabled by default, so It'll automatically retries - hanging.

See: https://www.gnu.org/software/libc/manual/html_node/Flags-for-Sigaction.html

@khj809


jhlee525:
Ok then let's go with this solution 💪

PR #45, opened 2025-05-21
Improve printing agent responses

### LINKED ISSUE #22: Providing formatted console print
**Is your feature request related to a problem? Please describe.**
Currently, inference results are returned as a dictionary, which is not easily readable. It would be better if the output were displayed in a human-readable text format.

**Describe the solution you'd like**
The approach currently used in the example would be good.

**Describe alternatives you've considered**
None at the moment.

**Additional context**
It would be nice to support colored output, but I'm concerned that the method may differ across operating systems.

khj809:
Resolves #22 

- Added `print()` method on `AgentResponse*` classes to pretty-print contents using [rich](https://rich.readthedocs.io/en/latest/index.html)


grf53:
- Suggests the format of the `Call Tool` box and the `Tool Result` box be consistent.
- Suggests to append some mark(ellipsis or sth else) to notice the result is trucated.
- Maybe need to notice that the yellow text are the reasoning tokens in the document
  - Thought about a neat better way to show that in the console, but hard to find.

README.md
```diff
@@ -48,7 +48,7 @@ from ailoy import Runtime, Agent
 rt = Runtime()
 agent = Agent(rt, model_name="qwen3-8b")
 for resp in agent.query("When is your cut-off date?"):
-    print(resp)
+    resp.print()
```
grf53, line 51: It seems like these APIs that do the same thing had better to be consistent.

grf53 (approved):
Looks good to me!

PR #47, opened 2025-05-22
Logger env

jhlee525:
We can set `AILOY_LOG_LEVEL=debug` or `AILOY_LOG_LEVEL=DEBUG` to adjust log-level.

Plus, log bug fixed and add more logs. But we need more loggings.

### DIFF
--- a/core/src/logging.cpp
+++ b/core/src/logging.cpp
@@ -1,36 +1,70 @@
 #include "logging.hpp"
 
+#include <spdlog/sinks/stdout_color_sinks.h>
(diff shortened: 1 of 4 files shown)


grf53 (approved):
Look good enough as the feature.
Do you want to merge this first or do it after adding quite an amount of loggings?

jhlee525:
I think the first one seems good.

However, the second one is also fine, so feel free to add log in this branch

khj809 (approved):
Looks good.
This may be urgently needed for Windows debugging.

Issue #20, opened 2025-05-15
Device selection functionality


jhlee525:
**Is your feature request related to a problem? Please describe.**
Currently, we are supporting metal and vulkan. In the case of Vulkan, all of our inference uses device `0`. On machines with multiple GPUs, it's not possible to select a alternative GPU.

**Describe the solution you'd like**
For NVIDIA GPUs, it's possible to select devices using an environment variable like `NVIDIA_VISIBLE_DEVICES`. We'd like to adopt a similar approach by allowing device selection through an environment variable. Something like `AILOY_VISIBLE_GPUS` would be acceptable name.

**Describe alternatives you've considered**
Of course, there are various possible approaches, like input arguments in `define_component`. However, according to our current implementation, I believe this is the simplest way to implement.

**Additional context**
None

PR #49, opened 2025-05-22
Device selection

grf53:
## Ticket
#20 

## Description
Add feature to get device id (as `int_t` or `uint_t` value) as component attribute, then use the device with that device id.
Device type is still set by the target environment, but the device id can be configured.

First, it checks if a device with the given device id exists, and if so, it uses it.
If not, it checks again with id 0, and if so, it uses it(logs this situation with `info` level), otherwise it returns `error_output_t`.

## Changes
- Code for getting the device for embedding models and language models is integrated.
  - `tvm_model_t::prepare_engine()` is removed.
  - `tvm_embedding_model_t` gets device instead of getting it internally.
- Added test to try to get the device with id `1`

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -770,6 +770,8 @@ export async function defineAgent(
   args?: {
     /** Optional system message to set the initial assistant context */
     systemMessage?: string;
+    /** Optional device id to set the device id to run LLM model */
+    device?: number;
     /** A parameter for API key usage.
      * This field is ignored if the model does not require authentication. */
     apiKey?: string;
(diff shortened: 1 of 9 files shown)


jhlee525:
Any python version?

grf53:
> Any python APIs?

Python API just bypass the `attrs` into `define_component`, so no change was needed.

BUT, there was a bug on defining em and vs components.
The `faiss_vector_store` component should get `vector_store_attrs ` but was getting `embedding_model_attrs`, and `embedding_model_attrs` was going nowhere.

Fix is on commit 0193556ca549512f0e80f2bd7930d4358f50339b.

jhlee525 (approved):
I see. Then LGTM 👍

khj809 (approved):
Looks good

PR #50, opened 2025-05-22
refactor: use huggingface model id instead of custom id for models

grf53:
## Description
Use huggingface model ids for Agent, VectorStore APIs just like Runtime APIs, not to confuse users.

### DIFF
--- a/examples/python/movie_agent.py
+++ b/examples/python/movie_agent.py
@@ -12,7 +12,7 @@ async def main():
     if tmdb_api_key is None:
         tmdb_api_key = input("Enter TMDB API Key: ")
 
-    agent = Agent(rt, model_name="qwen3-8b")
+    agent = Agent(rt, model_name="Qwen/Qwen3-4B")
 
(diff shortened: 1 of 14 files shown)


jhlee525 (approved):
Thank you!

examples/python/movie_agent.py
```diff
@@ -12,7 +12,7 @@ async def main():
     if tmdb_api_key is None:
         tmdb_api_key = input("Enter TMDB API Key: ")
 
-    agent = Agent(rt, model_name="qwen3-8b")
+    agent = Agent(rt, model_name="Qwen/Qwen3-4B")
```
khj809: Is this change intended? (8b->4b)

khj809 (approved):

examples/python/movie_agent.py
grf53: Oh wow, maybe I changed it locally to `qwen-4b` to test 4B model and then replaced in bulk.
There could be another. Let me check it.

grf53: Fortunately(?), there seems no more. Let me fix it.

PR #57, opened 2025-05-26
fix: set to use UTF-8 on Windows environment

grf53:
## Description
The default 'code page' is not UTF-8 in Windows environment, so the progress bars were shown improperly in Windows consoles.
- Some characters like `■` or `°`  were shown as broken.
- Progress bars are not overwritten, but printed in new lines for every updates.

So we need to set the code page as `CP_UTF8` explicitly in the code.

Special thanks to @chachaAA!

### Before (CP_949)
image

### After (CP_UTF8)
Photo_2025-05-27-00-54-39

### DIFF
--- a/vm/src/mlc_llm/model_cache.cpp
+++ b/vm/src/mlc_llm/model_cache.cpp
@@ -501,6 +501,10 @@ download_model(const std::string &model_id, const std::string &quantization,
     }
   }
 
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


khj809 (approved):
Wow thanks!!

jhlee525:
I didn’t know that it can be configured in our code.

Thanks again @chachaAA

PR #65, opened 2025-05-28
Update Electron RAG example

khj809:
Updated Electron RAG example app
- Let users be able to copy & paste texts directly, rather than taking it from txt files only.
- Separate "Update Vectorstore" so users can clear and update vectorstore whenever the content is changed.
- Add readme with attached screen captures.

## Additional changes
- Added `engines.node` and `os` in `package.json`
- Fix incorrect agent test code


grf53 (approved):
LGTM

PR #69, opened 2025-05-29
fix(cmake): set minimum version as 3.28, etc.

khj809:
- Fix readme which states an incorrect CMake minimum version.
- Set `CMAKE_OSX_SYSROOT` instead of adding `-isysroot` option on targets
- Set minimum OSX deployment target as 14.0


jhlee525 (approved):
👍

PR #73, opened 2025-05-30
Add new example: Gradio chatbot

khj809:
This PR adds a simple Gradio integration example using Chatbot interfaces, with supports for streaming responses, reasoning, and tool calling capabilities.

## Breaking Changes

* Replaced the `end_of_turn` field in `AgentResponse` with `is_type_switched` to better detect when the response type changes during streaming.
  The `end_of_turn` flag wasn’t working properly for responses of type `reasoning`, and in fact, it doesn’t make sense to set `end_of_turn` on the last reasoning response—since we can’t know it’s the last until a non-reasoning response arrives.
  While working on the example, I realized that simply detecting when the response type switches is sufficient. We can then append a new `gr.ChatMessage` only when the type changes.
  **Since this is a breaking change to the public API, I can separate it into its own PR, and also update the JS bindings accordingly.**

### DIFF
--- a/examples/gradio_chatbot/pyproject.toml
+++ b/examples/gradio_chatbot/pyproject.toml
@@ -1,0 +1,9 @@
+[project]
+name = "ailoy-gradio-chatbot-example"
+version = "0.1.0"
+description = "Ailoy + Gradio integration example"
+readme = "README.md"
+requires-python = ">=3.10"
+dependencies = [
+    "gradio>=5.31.0",
+]
--- a/examples/gradio_chatbot/pyproject.toml
+++ b/examples/gradio_chatbot/pyproject.toml
@@ -10,4 +10,5 @@ dependencies = [
 ]
 
 [tool.uv.sources]
+# TODO(@khj809): change to next version (probably 0.0.2)
 ailoy-py = { path = "../../bindings/python", editable = true }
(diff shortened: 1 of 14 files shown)


examples/gradio_chatbot/pyproject.toml
```diff
@@ -0,0 +1,13 @@
...
+description = "Ailoy + Gradio integration example"
+readme = "README.md"
+requires-python = ">=3.10"
+dependencies = [
+    "ailoy-py",
+    "gradio>=5.31.0",
+]
+
+[tool.uv.sources]
+ailoy-py = { path = "../../bindings/python", editable = true }
```
grf53, line 14: This will probably need to be changed to version-wise before the 0.0.2 release.

grf53 (approved):
Example looks flawless! And the other changes look good to me.

examples/gradio_chatbot/pyproject.toml
khj809: Yes indeed. I'll add todo comment here.

khj809:
> Do you think the isTypeSwitched field is actually useful for users? We could consider removing it.
> 
> If you think it's necessary, I'm happy to agree with that as well.

Yes it was definitely helpful while creating the demo app.
Although users can manage the state and detect when the state changes, we can also provide this feature out-of-the-box and let users be beneficial.

PR #36, opened 2025-05-18
Implementing Python packaging CI/CD pipelines

khj809:

PR #75, opened 2025-06-02
Github Action to deploy documents

grf53:
## Description 
Add Github Action to test to build document & build and deploy document

- Test document build
  - triggered on the **_pull request_** into `main` branch with the changes in `docs/**`.
  - just try to build document.
- Build and Deploy document
  - triggered on the **_push_** into `main` branch with the changes in `docs/**`.
  - build document and deploy to https://brekkylab.github.io/ailoy

## Things to consider
These workflows are only for documents. It might be wanted to integrate these with any other CI workflows like packaging or unit testing.

## Reference
https://docusaurus.io/ko/docs/deployment#triggering-deployment-with-github-actions


grf53:
I need to consider those API references(js, python) to be built along with the main document.

jhlee525:
One thing that came to mind is that the documentation should be based on the latest release, not the main branch.

This is because users who install the package through a package manager, probably use latest released version.

We’ll probably need to update the PR trigger conditions and change them from push accordingly.

.github/workflows/deploy-docs.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Build Typescript API Document
+        run: npx typedoc ../bindings/js-node/src/index.ts --tsconfig ../bindings/js-node/tsconfig.json --favicon static/img/favicon.ico --out static/tsdocs
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+          cache: pip
+
+      - name: Install ailoy-python
+        run: pip install ../bindings/python
```
grf53, line 34: This step would be better to install the packaged python wheel instead of sources after integration with #36.

PR #76, opened 2025-06-02
Fix: Agent loop behavior on tool calls

khj809:
Previously, we exited the `infer` loop as soon as tool calls were encountered and their results were evaluated. However, it turns out we should only break the loop when the infer request returns a finish reason of `"stop"`.

This PR addresses the issue by introducing an `is_tool_called` flag and rerunning `infer` if it's set. As a result, the condition for completing the `query` generator has been updated: it now checks not only for a `"stop"` finish reason but also ensures that no tool calls were triggered.

## Additional Fixes
* Added `await client.close()` in the JS Agent's MCP functions to ensure the MCP server process terminates correctly.

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -514,6 +514,7 @@ export class Agent {
           }
         }
       });
+      await client.close();
 
       return parsedContent;
     };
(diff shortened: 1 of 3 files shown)


jhlee525 (approved):
👍

grf53 (approved):
👍

Issue #38, opened 2025-05-19
Add 'callable' itself as tool with generated tool description from docstring


grf53:
**Is your feature request related to a problem? Please describe.**
It is not a problem solving feature, but a convenience feature.

**Describe the solution you'd like**
When you use LLMs with transformers, you just can [add python functions](https://huggingface.co/docs/transformers/v4.51.3/en/chat_extras#tools) with docstrings as the tools.
It is very convenient to use custom tool because we can omit to write the tool description if we have google-style docstrings on the functions.
That is already provided in many tools, so it would be good to have this feature to provide the user experience similar to the others.

**Describe alternatives you've considered**
We can refer to the transformers [code](https://github.com/huggingface/transformers/blob/40a493c7ed4f19f08eadb0639cf26d49bfa5e180/src/transformers/utils/chat_template_utils.py#L213) to implement this feature.
The code has a recursive logic to parse the type hints, but is not really complicated overall.

**Additional context**

jhlee525:
P.S. The title is not intuitive. I can't figure out at once. Could you rename this issue?

grf53:
We don't plan on supporting other types of docstring in the near future, so Let me close this for now.

PR #77, opened 2025-06-02
Docstring-based `add_py_function_tool()`

grf53:
## Ticket
#38

## Description
Add feature to put python callable to the `add_py_function_tool()` without extra tool description JSON document, the function's docstring is parsed into the tool description(JSON Schema format).

Now it only supports google-style docstring just like the transformers.

### A Breaking Change
The parameter order of the `add_py_function_tool()` is changed
from: `def add_py_function_tool(self, desc: dict, f: Callable[..., Any]):`
to: `def add_py_function_tool(self, f: Callable[..., Any], desc: Optional[dict] = None):`
in order to make the tool description optional.

This had better to be merged quickly not to confuse users.

### DIFF
--- a/bindings/python/ailoy/agent.py
+++ b/bindings/python/ailoy/agent.py
@@ -16,12 +16,13 @@ from urllib.parse import urlencode, urlparse, urlunparse
 import jmespath
 import mcp
 import mcp.types as mcp_types
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, ValidationError
 from rich.console import Console
 from rich.panel import Panel
 
 from ailoy.ailoy_py import generate_uuid
 from ailoy.runtime import Runtime
+from ailoy.tools import DocstringParsingException, TypeHintParsingException, get_json_schema
 
 __all__ = ["Agent"]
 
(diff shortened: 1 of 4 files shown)


bindings/python/ailoy/agent.py
```diff
@@ -576,14 +577,27 @@ def add_tool(self, tool: Tool) -> None:
...
+            try:
+                tool_description = ToolDescription.model_validate(desc)
+            except ValidationError:
+                pass
+
+        if tool_description is None:
+            try:
+                tool_description = ToolDescription.model_validate(get_json_schema(f).get("function"))
+            except (TypeHintParsingException, DocstringParsingException) as e:
+                ValueError("Failed to parse docstring", e)
```
khj809:
```suggestion
                raise ValueError("Failed to parse docstring", e)
```

bindings/python/tests/test_2_agent.py
```diff
@@ -49,32 +58,33 @@ def get_current_temperature(location: str, unit: str):
...
+    #                 },
+    #             },
+    #             "required": ["location", "unit"],
+    #         },
+    #         "return": {
+    #             "type": "number",
+    #             "description": "The current temperature at the specified location in the specified units, as a float.",
+    #         },
+    #     },
+    # )
```
khj809: unused?

grf53: Now we have the alternative for this explicit tool description(parsed one from docstring), we don't need to give one like this.(but still possible to give those things to override.)
So yeah, we could remove this.

jhlee525 (approved):
👍 

Note that we have to update our reference API (until the next release.)

grf53:
> Note that we have to update our reference API (until the next release.)

Just checked that pdoc generates document with modified APIs

grf53:
Changed the validation logic for tool description in `add_py_function_tool()`
1. When the tool description(`desc`) is explicitly given as parameter and the validation is failed for that description, it just raises the `ValidationError` instead of trying to use parsed description.
2. To make the code clearer and easier to see where the errors raised, I split up the parts that call `get_json_schema()` and `ToolDescription.model_validate()`.
3. (Added docstring describing possible errors from the function.)

PR #78, opened 2025-06-02
Validate runtime client before using

### LINKED ISSUE #63: Unexpected exception on terminating python without deleting `Agent` instance
**Describe the bug**
Exception raises when the python running an `Agent` instance is terminated without deleting the `Agent` instance.

```
Exception ignored in: <function Agent.__del__ at 0x106d13b00>
Traceback (most recent call last):
  File "/Users/woorak/.pyenv/versions/ailoy/lib/python3.12/site-packages/ailoy/agent.py", line 372, in __del__
  File "/Users/woorak/.pyenv/versions/ailoy/lib/python3.12/site-packages/ailoy/agent.py", line 425, in delete
  File "/Users/woorak/.pyenv/versions/ailoy/lib/python3.12/site-packages/ailoy/runtime.py", line 157, in delete
  File "/Users/woorak/.pyenv/versions/ailoy/lib/python3.12/site-packages/ailoy/runtime.py", line 68, in _send_type2
AttributeError: 'NoneType' object has no attribute 'send_type2'
```

**System Environment**
 - Version: 0.0.1
 - OS: Every OS
 - Accelerator [e.g. metal, vulkan]: Every

**To Reproduce**
Run python and type:
```python
>>> from ailoy import Runtime, Agent
>>> rt = Runtime()
>>> agent = Agent(rt, "Qwen/Qwen3-0.6B")
>>> exit()
```

**Expected behavior vs. Actual behavior**
Expected: termimated without Exception
Actual: `AttributeError` raises

**Screenshots**

**Additional context**

grf53:
## Ticket
resolves #63 

## Description
Not only an agent can be deleted after the runtime has stopped when the program terminates, but an agent can also use an invalid runtime anytime in some circumstances. (See the following code.)
So we should raise an error properly when the agent try to use it instead of preventing that situation.

```python
rt = Runtime()
rt.stop()
agent = Agent(rt, ...)  # AttributeError: 'NoneType' object has no attribute 'send_type2'
```

However, in `Agent.delete()`, it could be better to just skip the Runtime task rather than raising errors.

### DIFF
--- a/bindings/js-node/src/runtime.ts
+++ b/bindings/js-node/src/runtime.ts
@@ -44,6 +44,8 @@ export class Runtime {
    */
   private listener: Promise<void> | null;
 
+  alive: boolean = false;
+
   constructor(url: string = "inproc://") {
     this.url = url;
     this.resolvers = new Map();
@@ -62,18 +64,23 @@ export class Runtime {
     return new Promise<void>(async (resolve, reject) => {
       this.registerResolver(txid, resolve, reject);
       while (this.resolvers.has(txid)) await this.listen();
+      this.alive = true;
     });
   }
 
   stop(): Promise<void> {
-    const txid = generateUUID();
-    if (!this.client.send_type1(txid, "disconnect"))
-      throw Error("Disconnection failed");
-    return new Promise<void>(async (resolve, reject) => {
-      this.registerResolver(txid, resolve, reject);
-      while (this.resolvers.has(txid)) await this.listen();
-      stopThreads(this.url);
-    });
+    if (this.alive) {
+      const txid = generateUUID();
+      if (!this.client.send_type1(txid, "disconnect"))
+        throw Error("Disconnection failed");
+      return new Promise<void>(async (resolve, reject) => {
+        this.registerResolver(txid, resolve, reject);
+        while (this.resolvers.has(txid)) await this.listen();
+        stopThreads(this.url);
+        this.alive = false;
+      });
+    }
+    return Promise.resolve();
   }
 
   async call(funcName: string, inputs: any = null): Promise<any> {
--- a/bindings/js-node/src/runtime.ts
+++ b/bindings/js-node/src/runtime.ts
@@ -44,7 +44,7 @@ export class Runtime {
    */
   private listener: Promise<void> | null;
 
-  alive: boolean = false;
+  private alive: boolean = false;
 
   constructor(url: string = "inproc://") {
     this.url = url;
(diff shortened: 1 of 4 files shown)


jhlee525 (commented):
👍  Should this logic also be applied to JS?

bindings/python/ailoy/agent.py
```diff
@@ -388,6 +388,9 @@ def define(self, model_name: ModelName, api_key: Optional[str] = None, **attrs)
         if self._component_state.valid:
             return
 
+        if self._runtime._client is None:
+            raise ValueError("Runtime is not valid.")
+
```
khj809: I think it would be better to have a dedicated method or property (such as `.is_alive()`) rather than accessing to the private member outside the runtime class.

grf53: Added property `Runtime.alive` into both python and js.

grf53:
Added `Runtime.alive` both into Python and JS as 'property' that shows the validity of `Runtime`. The way to implement it is different between Python and JS, but way to use is same.

And also, the checks for validity of component state are added into the `Agent.query()` both in Python and JS.
With those, users can get more coherent errors instead of kinda `low-level` errors in the situation like:
```python
rt = Runtime()
agent = Agent(rt, "Qwen/Qwen3-8B")
agent.delete()  # unexpected delete
for resp in agent.query("..."): 
    agent.print(resp)
```

#### Before
```
    raise RuntimeError(packet["body"]["reason"])
RuntimeError: There is no channel can handle this request
```
#### After
```
ValueError: Agent is not valid. Create one or define newly.
```

bindings/js-node/src/runtime.ts
```diff
@@ -44,6 +44,8 @@ export class Runtime {
    */
   private listener: Promise<void> | null;
 
+  alive: boolean = false;
```
khj809: Unlike in Python, there's a possibility of assigning values to this field outside the class, which is not an expected behavior. To prevent this in advance, how about making this as private (such as `#_alive`) and creating the method (`.alive()`) explicitly?

bindings/python/ailoy/agent.py
khj809: Continued on the below review.

bindings/js-node/src/runtime.ts
grf53: Good point, thanks! I made it as private

khj809 (approved):
Looks good!

Issue #79, opened 2025-06-02
Unified Message Format Proposal for LLM Input/Output


jhlee525:
## Introduction

Currently, the message flow from user input to final output consists of four distinct stages, each with slightly different formats. this forces adaptation logic for each routines, These are:

- User Input
  - The user provides input either Python or JavaScript, through the `Agent` class
  - Internally, this input is passed to C++ following a specific schema.
  - For multi-turn conversations, the input must also be reserved internally.
- LLM Input
  - On the C++ side, we are using `minja` for chat template generation.
  - Input messages must be converted to a format compatible with the `minja` template.
- LLM Output
  - The model generates responses as next tokens.
  - These tokens are converted into dictionary-style deltas and streamed from C++ back to Python/JS.
- User Output
  - Final output shown to the user is a stream, which requires merging multiple intermediate structures.
  - This results in another transformation layer.

Each stage uses a slightly different format, so manual format adaptation is required at every boundary. Moreover, the chat templates vary by model (e.g., Gemma vs Qwen vs chatml), so the input messages must be restructured according to model-specific conventions. This process takes up a significant portion of the current implementation.

## References: HF Transformers

https://huggingface.co/docs/transformers/chat_templating
https://huggingface.co/docs/transformers/chat_templating_multimodal
https://huggingface.co/docs/transformers/chat_extras

### Core Parameters

In Hugging Face Transformers, the entry point for templating is the apply_chat_template function. This function converts message dictionaries into prompts that conform to the expected format of each model. The main parameters are as follows:

- messages: A list of dictionaries with "role" and "content" keys
- tools: A list of tool specifications (functions)
- documents: (RAG-related) – we omit this for now

### Plain Conversation
```
[
  {"role": "user", "content": "Hello, how are you?"},
  {"role": "assistant", "content": "I'm doing great. How can I help you today?"},
  {"role": "user", "content": "I'd like to show off how chat templating works!"},
]
```

### Tool Use
```
tools = [
  {
    "type": "function",
    "function": {
      "name": "current_time",
      "description": "Get the current local time as a string.",
      "parameters": { "type": "object", "properties": {} }
    }
  },
  {
    "type": "function",
    "function": {
      "name": "multiply",
      "description": "Multiply two numbers",
      "parameters": {
        "type": "object",
        "properties": {
          "a": { "type": "number", "description": "First number" },
          "b": { "type": "number", "description": "Second number" }
        },
        "required": ["a", "b"]
      }
    }
  }
]

[...]

PR #80, opened 2025-06-02
Language model inference revised

jhlee525:
1. mlc-llm-engine & mlc-llm dependency has been removed
2. introduces top-p & temperature in arguments
3. introduces grammar in LLM (using `xgrammar`). it is intended to use paired with "streaming_mode" concepts
4. New message format has been applied, intruduced in #79 

Currently, legacy codes are not deleted. If we complete migrations, these should be deleted.

TODO
- OpenAI adaptation
- Py / JS logic

### DIFF
--- a/core/include/exception.hpp
+++ b/core/include/exception.hpp
@@ -112,9 +112,13 @@ template <typename T = runtime_error>
   requires is_exception_reason<T>
 class exception_t : public std::exception {
 public:
-  template <typename... TArgs>
-  exception_t(TArgs... args)
-      : std::exception(), errstr(build_errstr(T(args...).what())) {}
+  template <typename... TArgs> exception_t(TArgs... args) : std::exception() {
+    if constexpr (requires { T::what(); }) {
+      errstr = build_errstr(T::what());
+    } else {
+      errstr = build_errstr(T(std::forward<TArgs>(args)...).what());
+    }
+  }
 
   const char *what() const noexcept override { return errstr.c_str(); }
 
@@ -124,7 +128,7 @@ private:
 
 template <typename T = runtime_error, typename... TArgs>
 exception_t<T> exception(TArgs... args) {
-  return exception_t{args...};
+  return exception_t<T>(std::forward<TArgs>(args)...);
 }
 
 } // namespace ailoy
--- a/vm/CMakeLists.txt
+++ b/vm/CMakeLists.txt
@@ -157,6 +157,22 @@ if(AILOY_WITH_FAISS)
     target_link_libraries(ailoy_vm_obj PUBLIC faiss)
 endif()
 
+# # xgrammar
+# FetchContent_Declare(
+#     xgrammar
+#     URL https://github.com/mlc-ai/xgrammar/archive/refs/tags/v0.1.19.tar.gz
+#     EXCLUDE_FROM_ALL
+# )
+# # Since xgrammar overrides settings via a config file, calling `set` in here has no effect.
+# # Instead, create a `config.cmake` file to set the configuration.
+# if(NOT EXISTS "${CMAKE_BINARY_DIR}/config.cmake")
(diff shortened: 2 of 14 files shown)


bindings/python/ailoy/agent.py
```diff
@@ -29,58 +29,59 @@
 ## Types for OpenAI API-compatible data structures
```
khj809: These are no longer OpenAI API-compatible. We may need to remove or change the comment. (same in JS)

vm/tests/test_tvm_language_model.cpp
```diff
@@ -1,333 +1,310 @@
...
+                     ->operator std::string();
+    if (out.finish)
+      break;
+  }
+  return agg_out;
 }
 
-TEST(LanguageModelTest, TestInfer) {
-  std::shared_ptr<ailoy::component_t> language_model;
+TEST(TestLangModelV2, TestSimple) {
```
khj809:
```suggestion
TEST(TestLangModel, TestSimple) {
```
Same for another test cases

vm/tests/test_tvm_language_model.cpp
```diff
@@ -1,333 +1,310 @@
...
-        std::cout << "(" << i << " tokens)" << std::endl;
-        break;
-      }
+  std::shared_ptr<ailoy::component_t> model = get_model();
+  auto messages = ailoy::decode(messages_str, ailoy::encoding_method_t::json);
+  auto tools = ailoy::decode(tools_str, ailoy::encoding_method_t::json);
+
+  // model->set_builtin_grammar("tool_call", "json");
+  auto answer = infer(model, messages, tools);
+  // std::cout << answer << std::endl;
```
khj809: How about changing this to debug logging?

vm/src/tvm/language_model.hpp
```diff
@@ -0,0 +1,213 @@
...
+                  const std::string &open_indicator,
+                  const std::string &close_indicator);
+
+    /**
+     * @brief Whether the currently generated token (history) matches the
+     * indicator
+     * @param indicator_type "open" or "close"
+     * @param history History of generated token
+     */
+    bool check_indecator(const std::string &indicator_type,
```
khj809:
```suggestion
    bool check_indicator(const std::string &indicator_type,
```

vm/src/tvm/language_model.hpp
```diff
@@ -0,0 +1,213 @@
...
+  bool is_eor(const std::string &tok) const;
+
+  /** End of reasoning */
+  bool is_eor(int32_t tok) const;
+
+  /** Begin of sentence */
+  bool is_bos(const std::string &tok) const;
+
+  /** End of sentence */
+  bool is_eos(const std::string &tok) const;
```
khj809: Is it okay not to have `bool is_eos(int32_t tok) const;` similar to another?

bindings/python/ailoy/agent.py
jhlee525: Ok, we can remove it since the rule of thumb is our internal rule. But note that it is still "OpenAI-compatible", except for the reasoning fields.

vm/src/tvm/language_model.hpp
jhlee525: yes it's because probably it is not used. but it's ok to put it for future purpose.

vm/tests/test_tvm_language_model.cpp
jhlee525: 👍

vm/tests/test_tvm_language_model.cpp
jhlee525: Thank you applied

khj809 (approved):
I have checked everything as far as I can, and seems there's no problem now.
Thanks for the huge work!

grf53 (approved):
👍👍👍

PR #81, opened 2025-06-02
Persistent MCP Server Lifecycle Management

khj809:
Previously, MCP servers were launched and terminated on-demand each time the agent attempted to list or call tools. This PR introduces a persistent lifecycle for MCP servers, ensuring they remain active until explicitly removed or when the agent itself is deleted.

### Changes:

* **Python**

  * `MCPServer` now spawns a **dedicated subprocess** that acts as a bridge between the main process and the MCP stdio server.
  * Initially it was implemented in `threading`-based approach, but it was unreliable in certain environments (e.g., `pytest`).
  * Communication between processes is handled via `multiprocessing.Pipe` using structured, type-safe Pydantic messages (`RequestMessage` and `ResponseMessage`).
  * The `stdio_client` async context is maintained throughout the subprocess lifecycle and is properly cleaned up via `cleanup()`.

* **JavaScript**

  * `MCPServer` internally manages a single `Client` instance and maintains the connection using `StdioClientTransport`.
  * The connection persists until `cleanup()` is explicitly called, preventing repeated startup overhead.

* **Agent APIs**

  * Each MCP server instance is now given a **unique `name`**, which is used as a prefix for its tools.
  * The `Agent.add_tools_from_mcp_server()` method is still used as before, with the addition of a required `name` argument.
  * Agents can now cleanly remove tools from an MCP server using `remove_mcp_server()`.

### DIFF
--- a/bindings/js-node/src/mcp.ts
+++ b/bindings/js-node/src/mcp.ts
@@ -1,0 +1,84 @@
+import * as MCPClient from "@modelcontextprotocol/sdk/client/index.js";
+import * as MCPClientStdio from "@modelcontextprotocol/sdk/client/stdio.js";
+
+class MCPServer {
+  readonly name: string;
+  #params: MCPClientStdio.StdioServerParameters;
+  #client: MCPClient.Client;
+  #connected: boolean = false;
+
+  constructor(name: string, params: MCPClientStdio.StdioServerParameters) {
+    this.name = name;
+    this.#params = params;
+    this.#client = new MCPClient.Client({
+      name,
+      version: "dummy-version",
+    });
+  }
+
+  async start() {
+    if (this.#connected) return;
+    const transport = new MCPClientStdio.StdioClientTransport(this.#params);
+    await this.#client.connect(transport);
+    this.#connected = true;
+  }
+
+  async cleanup() {
+    if (!this.#connected) return;
+    await this.#client.close();
+    this.#connected = false;
+  }
+
+  async listTools() {
+    if (!this.#connected) {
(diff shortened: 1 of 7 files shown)


jhlee525:
Cool. Can we switch our current MCP examples(github API) to playwright API?

Please let me know your opinions.

bindings/js-node/src/mcp.ts
```diff
@@ -0,0 +1,93 @@
...
+ * - It manages a single `MCPClient.Client` instance and connects via `StdioClientTransport`.
+ * - The connection lifecycle is handled through the `start()` and `cleanup()` methods.
+ * - Tools can be discovered with `listTools()`, and invoked using `callTool()`.
+ * - Unlike in Python, this class does not use subprocesses; it runs entirely within the calling Node.js process.
+ */
+class MCPServer {
+  readonly name: string;
+  #params: MCPClientStdio.StdioServerParameters;
+  #client: MCPClient.Client;
+  #connected: boolean = false;
```
jhlee525, line 17: just of curious, is there any reason to prefix `#` for the variables?

khj809: `#` makes the properties as private.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

khj809:
> Cool. Can we switch our current MCP examples(github API) to playwright API?
> 
> Please let me know your opinions.

I'm planning to make Playwright example in a separate PR after this is merged.

jhlee525 (approved):

PR #82, opened 2025-06-03
Clear chat history

grf53:
## Description
Add feature to clear chat history to the `Agent` to prevent LLM to unintentionally refer to the previous chat history.

## Changes
- `Agent` has member `system_message`(or `systemMessage`) explicitly, that is set as model's default system message or overridden system message with the parameter.
  - If there is no available value, it is set as `None`(or `undefined`).
- When `clear_history()`(or `clearHistory()`) is called, the member `messages` reset considering its system message.

### DIFF
--- a/bindings/python/ailoy/agent.py
+++ b/bindings/python/ailoy/agent.py
@@ -360,7 +360,7 @@ class Agent:
 
         # Initialize messages
         self._messages: list[Message] = []
-        self._system_message: str = system_message
+        self._system_message: Optional[str] = system_message
 
         # Initialize tools
         self._tools: list[Tool] = []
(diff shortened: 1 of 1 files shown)


bindings/js-node/src/agent.ts
```diff
@@ -240,6 +240,8 @@ export class Agent {
 
   private messages: Message[];
 
+  systemMessage?: string;
+
```
khj809: If this is not intended to be assignable outside the class, it would be better to make it as private.

grf53: Oh, I thought wrong a little. It should be private. Thanks. I'll fix this.

khj809 (approved):
Looks good!

jhlee525:
Looks good.

Probably it would be good to merge after #80

jhlee525 (approved):

PR #84, opened 2025-06-03
Fix Rich Console output for Jupyter streaming by forcing ANSI terminal mode

isingmodel:
This PR updates the _consol initialization in agent.py to use force_jupyter=False and force_terminal=True for the Rich.Console.
This change disables the Rich HTML renderer in Jupyter and forces ANSI output, which resolves issues with broken or inconsistent streaming output when running in Jupyter notebooks.

Motivation
* Fixes garbled or unexpected LLM streaming output in Jupyter environments.
* Ensures consistent console rendering across both terminal and Jupyter.

### DIFF
--- a/bindings/python/ailoy/agent.py
+++ b/bindings/python/ailoy/agent.py
@@ -141,7 +141,11 @@ class ComponentState(BaseModel):
 
 ## Types for agent's responses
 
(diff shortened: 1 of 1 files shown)


khj809 (approved):
Thank you for your valuable first-time contribution!

PR #85, opened 2025-06-04
Set RPATH so that the target loads dependency from the same directory

jhlee525:
This resolves `dlopen` problems in node.js.

I've tested it in local machine. However, I'm not sure that it really fixed.


jhlee525:
checked in my device(mac) and linux

PR #89, opened 2025-06-04
Multiple Runtime instances

### LINKED ISSUE #37: Instantiating runtime twice results in abort
**Describe the bug**
In Python, the below code results in abort error.
```python
from ailoy import Runtime

rt = Runtime()
rt = Runtime()
libc++abi: terminating
[1]    2808 abort      python
```
I know that instantiating runtime twice with same address would raise the error, but users may frequently encounter this situation when they're on REPL. At least we need to show proper error messages without aborting the process.

**System Environment**
 - Version: 0.0.1
 - OS: tested on MacOS 15.4, but would be same for any other environments
 - Accelerator: Metal

**To Reproduce**
```python
from ailoy import Runtime

rt = Runtime()
rt = Runtime()
```

**Expected behavior vs. Actual behavior**
It should return a proper error message such as: `Failed to create Runtime: the address "inproc://" is already in use`

**Screenshots**
none

**Additional context**
none

grf53:
## Ticket
resolves #37 

## Description
Currently, the broker thread and the VM thread are required to be unique in a context, and lifetime of those threads are highly bound to the lifetime of `Runtime` instance of bindings. So creating multiple `Runtime` instances derives creating multiple broker & VM threads, which is not supported scenario.
And the URL of the broker and VM is designed to be various for some purposes. But this feature is also not being properly supported.
In many cases, creating multiple `Runtime` instances is a common requirement regardless of whether the URL is unique or multiple.

This PR resolved this by allowing multiple broker & vm threads along with the URLs, and managing the lifetime of those threads by the counts of `Runtime` instances by the URLs.

### DIFF
--- a/bindings/js-node/tests/index.spec.ts
+++ b/bindings/js-node/tests/index.spec.ts
@@ -1,7 +1,6 @@
 import { expect } from "chai";
 
 import { Runtime, NDArray } from "../src/index";
-import "./agent.spec";
 
 describe("JSVM", () => {
   it("run-echo", async () => {
--- a/bindings/js-node/csrc/binding.cpp
+++ b/bindings/js-node/csrc/binding.cpp
@@ -1,5 +1,7 @@
-#include <optional>
+#include <memory>
+#include <string>
 #include <thread>
+#include <unordered_map>
 
 #include <napi.h>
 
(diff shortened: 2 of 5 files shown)


bindings/js-node/tests/index.spec.ts
```diff
@@ -1,7 +1,6 @@
 import { expect } from "chai";
 
 import { Runtime, NDArray } from "../src/index";
-import "./agent.spec";
```
grf53, line 4: also, removed this line cuz there target spec files are set as `*.spec.{...}` in `.mocharc.json`.
(then you can adjust the test range by modifying `.mocharc.json`.)

grf53:
> (Not a critical issue), Is this problem related to this PR?

I think so... let me see.

PR #109, opened 2025-06-09
Fix tool name prefix for MCP tools

khj809:
OpenAI accepts tool names with pattern `^[a-zA-Z0-9_-]+$` only. If violated, it returns the following error.
```json
{
  "error": {
    "message": "Invalid 'tools[0].function.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.",
    "type": "invalid_request_error",
    "param": "tools[0].function.name",
    "code": "invalid_value"
  }
}
```

But our MCP tool names are prefixed with `/` separator, which is not included in the pattern. Hence this PR replaces the separator with `-`.
Verified that it works with openai now.

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -512,7 +512,7 @@ export class Agent {
         continue;
 
       const desc: ToolDescription = {
-        name: `${name}/${tool.name}`,
+        name: `${name}-${tool.name}`,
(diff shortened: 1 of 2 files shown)


grf53 (approved):
Looks clear and consistent.

jhlee525 (approved):

PR #113, opened 2025-06-10
Enable Gemini model

khj809:
This PR adds support for the Gemini models listed [here](https://ai.google.dev/gemini-api/docs/models?hl=ko#model-variations) (focusing on text generation models only).

* Since Gemini offers OpenAI-compatible APIs, this PR reuses existing OpenAI component logic wherever possible. With the refactored structure, it's easy to extend support to any models compatible with the OpenAI API.
* There is duplicated model description code across the Python and JS bindings. We should consider a better way to manage this before adding more API models from other providers.
* Applied minor fixes related to tool call id.


grf53 (approved):
👍

PR #115, opened 2025-06-17
Enable Claude Models

khj809:
This PR adds support for Claude models listed at [https://docs.anthropic.com/en/api/models-list](https://docs.anthropic.com/en/api/models-list).

* Claude requires different headers compared to OpenAI, so a virtual `headers()` method was added and overridden accordingly.
* Since Claude expects the "tool result" content as a plain string instead of an array of `{"type": "text", "text": "..."}` objects, we need to handle tool results separately. The overridden `convert_request_body()` method does such processing based on the result of base method.
* Anthropic states that their OpenAI-compatible APIs are not intended for long-term or production use in most cases, and they recommend using their native API to access the full feature set. The same applies to Gemini. However, due to limited resources, we’re currently unable to implement support for each provider's native API individually. Long term, we should consider integrating native APIs into our components.

### DIFF
--- a/bindings/js-node/src/agent.ts
+++ b/bindings/js-node/src/agent.ts
@@ -67,7 +67,21 @@ export type GeminiModelName =
   | "gemini-1.5-flash"
   | "gemini-1.5-pro";
 
-export type ModelName = TVMModelName | OpenAIModelName | GeminiModelName;
+export type ClaudeModelName =
+  | "claude-sonnet-4-20250514"
+  | "claude-3-7-sonnet-20250219"
+  | "claude-3-5-sonnet-20241022"
+  | "claude-3-5-sonnet-20240620"
+  | "claude-opus-4-20250514"
+  | "claude-3-opus-20240229"
+  | "claude-3-5-haiku-20241022"
+  | "claude-3-haiku-20240307";
+
+export type ModelName =
+  | TVMModelName
+  | OpenAIModelName
+  | GeminiModelName
+  | ClaudeModelName;
 
 interface ModelDescription {
   modelId: string;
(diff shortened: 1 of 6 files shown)


bindings/js-node/src/agent.ts
```diff
@@ -67,7 +67,21 @@ export type GeminiModelName =
...
-export type ModelName = TVMModelName | OpenAIModelName | GeminiModelName;
+export type ClaudeModelName =
+  | "claude-sonnet-4-20250514"
+  | "claude-3-7-sonnet-20250219"
+  | "claude-3-5-sonnet-20241022"
+  | "claude-3-5-sonnet-20240620"
+  | "claude-opus-4-20250514"
+  | "claude-3-opus-20240229"
+  | "claude-3-5-haiku-20241022"
+  | "claude-3-haiku-20240307";
```
grf53, line 78: Since we reduced the options for models of Gemini, so maybe we'd better to apply same standard here.

Considering [model overview page](https://docs.anthropic.com/en/docs/about-claude/models/overview), older versions of Opus, Sonnet might be good to be omitted, as they offer lower performance for the same cost. The only exception is Haiku, which seems to make sense to keep the lower version due to the costs(but Imo, the provider seems to recommend 3.5 more than 3, so omitting the Haiku 3 can also be a good idea).

There may also be a debate over whether to use the full name or an alias to specify the model, but it is not a big issue, and it is even more meaningless if it is modified to receive an arbitrary string in the future.

grf53 (approved):
Applying header looks nice to me.

bindings/js-node/src/agent.ts
khj809: As you mentioned, we can remove the model ids that might be unlikely used.
Moreover, while I'm working on image input supports, I did some refactoring on model definitions and now it can receive any model ids with preserving autocompletes for predefined model ids. I didn't push this work yet, but let me show and discuss about this tomorrow.

PR #116, opened 2025-06-18
Show total progress on model download

khj809:
To improve the user experience, this PR displays the total download progress—i.e., how many files have been downloaded so far—above the progress bar for the current file. This is achieved by using a block progress bar solely to update its label, without rendering an actual progress bar.
스크린샷 2025-06-18 오후 3 10 43


grf53 (approved):
🚀

PR #117, opened 2025-06-18
openai models added & can accept custom model

jhlee525:
1. More openai models are added
2. Now, we can accept custom model, by accepting direct `ModelDescription`


jhlee525:
@khj809 Now gpt reasoning models are enabled officially (o4-mini, o3 series).

Can we include `reasoning` responses to our response in `openai.cpp`?

Probably we have to use newer API rather than `chat_completion` API.

khj809:
@jhlee525 This PR is considered as the subset of https://github.com/brekkylab/ailoy/pull/120. Could we close this PR if you don't mind?

Issue #114, opened 2025-06-17
Model file versioning


khj809:
TVM recently underwent a major refactoring of its FFI (Foreign Function Interface), which introduced breaking changes that are incompatible with model libraries compiled using older versions of `mlc-llm`. Since similar breaking changes may occur in the future, it's important to establish a robust model versioning strategy to ensure each Ailoy version works reliably with its corresponding model version.

# Proposed Structure

## Model File Versioning

Currently, model files are organized as follows:

```
Qwen--Qwen3-8B (model_id)
 └─ q4f16_1 (quantization)
      ├── manifest-arm64-Darwin-metal.json (manifest)
      ├── lib-arm64-Darwin-metal.dylib (model library)
      ├── chat-template-config.json
      ├── mlc-chat-config.json
      ├── ndarray-cache.json
      ├── params_shard_0.bin
      ├── params_shard_1.bin
      ├── ...
      ├── params_shard_N.bin
      ├── Qwen--Qwen3-8B.j2
      ├── tokenizer_config.json
      └── tokenizer.json
```

To handle versioning, we propose placing files affected by the TVM version (e.g., `lib-arm64-Darwin-metal.dylib` and `manifest-arm64-Darwin-metal.json`) into subdirectories named after the **minimum supported Ailoy version**. For example:

```
Qwen--Qwen3-8B (model_id)
 └─ q4f16_1 (quantization)
      ├── 0.1.0
      │    ├── lib-arm64-Darwin-metal.dylib
      │    └── manifest-arm64-Darwin-metal.json
      ├── 0.2.0
      │    ├── lib-arm64-Darwin-metal.dylib
      │    └── manifest-arm64-Darwin-metal.json
      ├── lib-arm64-Darwin-metal.dylib (fallback or default)
      ├── manifest-arm64-Darwin-metal.json (fallback or default)
      ├── ...
```

* For Ailoy versions below `0.1.0`, the default manifest and library files are used.
* For versions `>= 0.1.0` and `< 0.2.0`, files under `0.1.0/` are used.
* For versions `>= 0.2.0`, files under `0.2.0/` are used.
* All other shared files remain unchanged and are reused across versions.

## Manifest as an API

Alternatively, we can run a lightweight API server that returns the appropriate manifest data based on the requested Ailoy version. The manifest data would be stored in a separate database. For example:

```
GET /models/manifest?model_name=Qwen--Qwen3-8B&quantization=q4f16_1&ailoy_version=0.1.0
```

[...]

PR #119, opened 2025-06-19
Update TVM codes

khj809:
TVM has recently updated its FFI-related APIs, so this PR updates the corresponding code to align with the latest changes.

## Breaking Changes

After this PR is merged, previously built model libraries will no longer be compatible.
To support older model libraries, we would need to implement model versioning as discussed in #114. However, this requires a more robust architecture and will take time to design and implement.

Therefore, we've decided to drop backward compatibility with versions `<= 0.0.2` for now and encourage users to upgrade Ailoy to the latest version.
We plan to merge this PR just before releasing version `0.0.3`.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -9,7 +9,7 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(nlohmann_json)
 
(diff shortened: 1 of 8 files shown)


grf53 (approved):
Good enough to be merged. Let's look for the best time to merge.

jhlee525 (approved):

PR #120, opened 2025-06-20
Add support for multimodal inputs

khj809:
This PR introduces support for multimodal inputs—such as images and audio—for API-based model providers (OpenAI, Gemini, Claude).

* **Revised OpenAI API schemas** to better align with the official specification by separating request messages (system, user, assistant, tool calls) from response messages (assistant). This change enables user messages to contain multiple content types, including text, images, and audio.
* **Updated `agent.query` API** to handle image and audio inputs:

  * **Python**: Not only a plain string, the input can now be a list of strings, Pillow `Image`, or instances of `AgentInputImageUrl`, `AgentInputImagePillow`, or `AgentInputAudioBytes`. This allows support for both base64 from [Pillow](https://pillow.readthedocs.io/en/stable/index.html) images and public URLs.
  * **Node.js**: Similarly, inputs can now be either a string, or an array of strings, or instances of `AgentInputImageUrl`, `AgentInputImageSharp`, or `AgentInputAudioBytes`, supporting both base64 from [Sharp](https://sharp.pixelplumbing.com/) images and public URLs.
* **Added test cases** for agents using API model backends.

---

### Breaking Changes

* Previously, the `Agent` class accepted a `model_name` parameter, supporting both local (TVM) and API-based models. It now requires an explicit `model` instance, created via specific model classes such as `TVMModel`, `OpenAIModel`, `GeminiModel`, or `ClaudeModel`. As a result, model-specific parameters (e.g., `api_key`) must now be passed directly to the corresponding model class, making the required parameters clearer and more type-safe for each case.

---

### TODO

* Update all example usages and documentation to reflect the new model instantiation pattern and multimodal support.
- [x] Update examples
- [x] Update documentations


khj809:
Following an internal review, we decided to consolidate all API-based models into a single `APIModel` class and rename `TVMModel` to `LocalModel`.

Additionally, there was feedback that the new Agent input interfaces were quite unnecessary. As a result, we will use the existing internal content types directly, while providing static helper methods to simplify their instantiation.

FYI @jhlee525 @grf53

jhlee525 (approved):

grf53 (approved):
LGTM

PR #130, opened 2025-07-02
Update ailoy minimum version on examples

khj809:
Updated the minimum ailoy package version for examples.

## Todo
- We need to add docs that ailoy-node has `sharp` as a dependency so users may have to update their bundler configs as described in [here](https://sharp.pixelplumbing.com/install/#bundlers). (See [examples/simple_rag_electron/vite.main.config.mts](https://github.com/brekkylab/ailoy/compare/update/examples-update-ailoy-version?expand=1#diff-ff6e38f5abda533f0fc1904419281376be24b4b954de7237c4ff39e4f0bd43f4) for vite example).


grf53 (approved):
:+1:

jhlee525 (approved):

PR #133, opened 2025-07-09
Remove `Self` type for Python3.10 compat

khj809:
The latest ailoy-py package has import error on Python3.10 due to the usage of `Self` type which is not present in 3.10.

### DIFF
--- a/bindings/python/ailoy/models/api_model.py
+++ b/bindings/python/ailoy/models/api_model.py
@@ -1,4 +1,4 @@
-from typing import Literal, Optional, Self, get_args
+from typing import Literal, Optional, get_args
 
 from pydantic import model_validator
 from pydantic.dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


grf53 (approved):

bindings/python/ailoy/models/api_model.py
```diff
@@ -44,7 +44,7 @@ class APIModel:
     provider: Optional[APIModelProvider] = None
 
     @model_validator(mode="after")
-    def validate_provider(self) -> Self:
+    def validate_provider(self):
```
grf53, line 47: Would it be better to be `'APIModel'`?
```suggestion
    def validate_provider(self) -> 'APIModel':
```

khj809: This also works, but now we have duplicated names across class name and method's return type. (that was why I tried to use `Self`)
Of course the duplication is negligible, but I think it would be better to try to minimize such duplication whenever possible.

PR #134, opened 2025-07-15
Abstract HTTP Request

khj809:
As the first step of enabling Emscripten runtime, this PR aims to implement abstract http request function which internally uses either cpp-httplib on native or Emscripten fetch on browsers.

## Browsers (Emscripten)
- The Emscripten's native fetch function from `<emscripten/fetch.h>` lacks features to get response headers, so we used `EM_ASYNC_JS` to use js fetch directly.
- Since it's not easy to link openssl on emscripten, we reimplemented calculating SHA1 checksum without using openssl in `model_cache.cpp` in order to totally remove openssl dependency.
- Added `test_http.html` to test http requests on browsers, which loads the `test_http.js` and `test_http.wasm` built with emcc.

## Native (cpp-httplib)
- We need to use `httplib::SSLClient` to handle requests to https, but it cannot handle requests to http. So we use variant of `httplib::Client` and `httplib::SSLClient` to use required one for each scheme.
- The codes previously using httplib have been replaced with the new `ailoy::http::request`.

### DIFF
--- a/core/tests/test_http.cpp
+++ b/core/tests/test_http.cpp
@@ -20,7 +20,7 @@ void test_get_request() {
   auto result = ailoy::http::request(req);
 
   if (!result) {
-    std::cout << "ERROR: " << result.error() << std::endl;
+    std::cerr << result.error() << std::endl;
     return;
   }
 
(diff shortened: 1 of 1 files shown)


grf53:
Checked that this works both on native / web. Great!
I just modified a little bit of codes to see that the `printErr` works well.

jhlee525 (approved):

grf53 (approved):
🚀

PR #136, opened 2025-07-17
Abstract Filesystem across Native and Browser(OPFS)

khj809:
As the second step for enabling Emscripten runtime, this PR aims to implement an abstract filesystem API layer available on either native filesystem or browser OPFS(Origin Private File System).


core/include/filesystem.hpp
```diff
@@ -0,0 +1,392 @@
...
+
+  result_t(error_code_t c = error_code_t::Success, const std::string &msg = "")
+      : code(c), message(msg) {}
+
+  bool success() const { return code == error_code_t::Success; }
+  operator bool() const { return success(); }
+};
+
+template <typename T> struct result_value_t {
+  result_t result_;
```
grf53, line 220: (Just a first-look opinion!)
`result_t` and `result_value_t` are looking commonly usable.
Perhaps there are other situations where they could be used, or perhaps there are already structures that exist for similar purposes. (actually, nothing came to mind specifically now.)
If then, those could be better to be somewhere more commonly referrable not only from filesystem-related, I think.

core/tests/test_http.cpp
```diff
@@ -100,9 +98,7 @@ void test_error_handling() {
 
 #ifdef EMSCRIPTEN
 
-extern "C" {
-EMSCRIPTEN_KEEPALIVE
-void run_tests() {
+int main() {
```
grf53, line 101: This looks great to have the same function signature with the native one.
I am just curious that we recently got a change that make we can use `int main()` instead of another signature like `void run_tests()`?

Maybe, if we use `main`, then we can run that function only by the `<script>` tag? I could see there removed the part using `Module._run_tests()`, but could not see any alternative is added.

grf53 (approved):
Looks great.
I left some comments, but regardless of that, thes changes are good enough to merge, as I see.
Just leave some opinions on those comments, please!

core/include/filesystem.hpp
khj809: I agree with that. I'll try to modularize them later for usages other than filesystem.

core/tests/test_http.cpp
khj809: We don't need any special handling for `int main()` when running emcc, it's just executed immediately after loaded in the browser via `<script>` tag. So we don't need to add `_main` to `EXPORTED_FUNCTIONS` as we did  for `run_tests()` previously.

PR #138, opened 2025-07-24
Support `typing.Literal` as type hint of python function tools

grf53:
### Support `typing.Literal` as type of tool parameters
Details are described in https://github.com/huggingface/transformers/pull/39633.

### Bug fixes in 'python function tools'
1. pydantic validation error
  When a type hint of a parameter is `int`, its tool description is generated including `{"type": "integer"}` but, there was no `"integer"` in possible type. So added it.
2. When a python function tool returns a string as the result, that was wrapped by `""`, because all the results were processed with `json.dumps()`. Now omitting it when the result output is a string.

### DIFF
--- a/bindings/python/ailoy/agent.py
+++ b/bindings/python/ailoy/agent.py
@@ -217,7 +217,7 @@ class ToolParameters(BaseModel):
 
 
 class ToolParametersProperty(BaseModel):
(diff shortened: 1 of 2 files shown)


grf53:
> It would be better to have a pytest case for this change, but it's not urgent.
the test and some more changes are added. Please take a look.

PR #140, opened 2025-07-25
Add test for tokenizers

khj809:
- We suffered from slow initialization of tokenizer class in browser environment, and the reason was sanitizer link options (`-fsanitize=undefined -fsanitize=address`) added in test executables. We fixed it by removing these options on Emscripten build time.
- We found that there's `-DCOMPILE_WASM_RUNTIME` compile definition in official mlc-ai/web-tokenizers build script (https://github.com/mlc-ai/tokenizers-cpp/blob/main/web/build.sh#L8). This is for preventing thread parallelism which leads to error on WASM runtime (see https://github.com/mlc-ai/tokenizers-cpp/pull/42). So we provide the same compile definition on Emscripten build time.
- Added the test executable (and html) for testing tokenizers only.


grf53 (approved):
LGTM

PR #142, opened 2025-07-28
Revise nodejs build scripts

khj809:
- We don't need to build packages for each node versions anymore, by replacing "node" runtime to "napi" runtime in prebuild. Every node version supporting specified N-API target "9" is available with a single package (see https://nodejs.org/docs/latest/api/n-api.html#node-api-version-matrix).
- Consolidated build time scripts into `scripts` directory.
- Removed matrices for each node version in workflows.
- Found that there're lots of `_.*` files included in package on MacOS environment, so removed them by applying the solution described in https://unix.stackexchange.com/questions/282055/a-lot-of-files-inside-a-tar.
- Fixed some build errors on Windows.


jhlee525 (approved):
👍

grf53 (approved):

PR #149, opened 2025-08-05
Run LLM on web with tvmjs

grf53:
## Description
This PR enables using `LocalModel` in `js-web` binding.
We can run LLM on web with a custom version of tvmjs and web-llm.

### tvmjs modification
- Add `ArtifactOPFSCache` as an implementation of `ArtifactCacheTemplate` interface.
  - It uses OPFS as the cache storage for the files. 
  - Unlike other ArtifactCache implementations, its base URL can be a filesystem-like path rather than an HTTP URL. This assumes a cached copy exists on a local (OPFS) path.
- Fix other typescript restricts such as 'isolated modules'.
- To build the necessary files from the source, TVM is now fetched via FetchContent even if it is not included in the main build. It used to build a js file for tvmjs through `npm run build:tvm`.
- `cc` files are not modified, so just build `tvmjs_runtime_wasi.js` in the fetched tvm source directory, and copy the result into the `src` directory.

### web-llm modification
In principle, I aim to include the minimum content required to run LLM and Embedding for Ailoy's purpose from web-llm, but some content may remain as comments or code to allow for future expansion.


vm/include/vm.hpp
```diff
@@ -49,6 +49,6 @@ void vm_start(const std::string &url,
  */
 void vm_stop(const std::string &url, const std::string &name = "default_vm");
 
-std::atomic_bool vm_ready{false};
+extern std::atomic_bool vm_ready;
```
grf53: to prevent link error: duplicated symbol

vm/src/vm.cpp
```diff
@@ -15,6 +15,8 @@ using namespace std::chrono_literals;
 
 namespace ailoy {
 
+std::atomic_bool vm_ready{false};
+
```
grf53: to prevent link error: duplicated symbol

vm/src/chat_manager.cpp
```diff
@@ -322,15 +322,22 @@ create_chat_manager_component(std::shared_ptr<const value_t> inputs) {
...
+      } else if (inputs_map->at("tools")->is_type_of<array_t>()) {
+        tools = inputs_map->at<array_t>("tools");
+      } else {
         return error_output_t(type_error("ChatManager: apply", "tools",
-                                         "array_t | string_t",
-                                         tools_val->get_type()));
-      tools = tools_val->as<array_t>();
+                                         "string_t | array_t",
+                                         inputs_map->at("tools")->get_type()));
+      }
```
grf53, line 340: This allows `tools` to be `string_t` or `array_t` just like the one in the language model.

bindings/js-web/src/agent.ts
```diff
@@ -302,13 +312,51 @@ export class Agent {
...
+      };
+    } else {
+      // Call runtime to define model component
+      const result1 = await this.runtime.define(
+        model.componentType,
+        this.componentState.name,
+        model.toAttrs()
+      );
+      if (!result1) throw Error(`model component define failed`);
+    }
```
khj809: I think we can abstract the model initialization logics into a common usage, by migrating these under `APIModel` and `LocalModel` respectively. As a result, the agent code would become much simpler.

bindings/js-web/src/agent.ts
```diff
@@ -613,17 +677,27 @@ export class Agent {
...
+          this.messages,
+          this.tools.map((v) => {
+            return { type: "function", function: v.desc };
+          }),
+          options?.reasoning,
+          this.genConfig ?? {}
+        ))!;
+      }
+
+      for await (const result of it) {
```
khj809: Ditto.

bindings/js-web/src/agent.ts
khj809: @grf53  I abstracted it. PTAL later.

grf53: I checked the refactored structure. Looks great!
Thank you for filling my void.

Let me merge this if you think it's ready, or you just can do.

PR #156, opened 2025-08-09
[WASM] Add VectorStore

khj809:
This PR adds the VectorStore supports on WASM environments using tvmjs embedding pipeline and Ailoy vector store components.
- The `Engine` class is used for both language models and embedding models, but its constructor takes ChatManager, which is not present in embedding models. So I concluded not to provide ChatManager to Engine as long as it's also used in embedding models. (@grf53 correct me if I'm wrong)
- In `EmbeddingPipeline`, there are some validation logics related to context window size and prefill chunk size. But for bge-m3 those values are set as 0, leads to throwing validation errors. In order to make it working, I just commented out the related logics, but not sure if it's safe to ignore. Please review this when you have time. @jhlee525 @grf53 
- bge-m3 model returns float16 dtype array, which is currently not considered in tvmjs `toArray()`. So I modified it to return `Uint16Array` if dtype is `float16`, and convert it to `Float32Array` by using `convertFloat16ArrayToFloat32Array()` implemented under `utils/float.ts`.


grf53:
> @grf53 Thanks for the clarification. Then I think `Engine` should not be used for embedding purpose. Since it corresponds to `tvm_language_model`, we may need another class that corresponds to `tvm_embedding_model`, which does not use chat templates at all. But problem is that `Engine` class uses its own locking mechanism to prevent running models multiple times at the same time. We may need to extend this lock to be used in our new embedding engines too.

Let me modify to remove `Engine`'s dependency on `ChatManager` in my changes. It would be more straight-forward, and easier.

bindings/js-web/src/llm/engine.ts
```diff
@@ -119,9 +109,9 @@ export class Engine {
         }
       }
     }
-    gpuDetectOutput.device.lost.then((info: any) => {
-      throw new DeviceLostError();
-    });
+    // gpuDetectOutput.device.lost.then((info: any) => {
+    //   throw new DeviceLostError();
+    // });
```
grf53, line 114: Was there a special reason to comment out this?

khj809: When I call `.dispose()` of Engine, it always throw the above error. But I found that it's already handled in `initWebGPU` in `tvmjs/runtime.ts`, so it seems redundant.

PR #158, opened 2025-08-09
docs: add WASM supports page

khj809:
This PR adds the documentation for WASM supports.


grf53:
Here is my suggestion for more readability and considering readers as more beginners.
: https://github.com/brekkylab/ailoy/compare/docs/wasm...docs/wasm-suggest?expand=1

khj809:
> I needed to add this configurations in my `vite.config.ts`.
> 
> ```
>   server: {
>     fs: {
>       allow: [
>         searchForWorkspaceRoot(process.cwd()),
>         searchForWorkspaceRoot(
>           "/Users/woorak/Documents/Github/ailoy/bindings/js-web/dist/"
>         ),
>       ],
>     },
> ```
> 
> Maybe its because I tested the example with `ailoy-web` as `"../ailoy/bindings/js-web/"`. But I am not really sure of that, so I am just reporting this.

Yes it's because you have installed via relative path. You can avoid this issue by creating package (`npm pack`) and install it (`npm install /path/to/ailoy-web-0.0.6.tgz`).


> And I had an issue that my assistant message contents are aligned weirdly in the middle. So I modified `text-align` from `center` to `justify` in the `App.css` Would this need to be in the guide too?
> 
> ```css
> #root {
>   ...
>   text-align: justify;
> }
> ```

Yes it looks better, but I tried to fix the really necessary parts as possible, in order to make the docs concise. Moreover, if we apply this css we need to create all the screen captures again..

khj809:
> Here is my suggestion for more readability and considering readers as more beginners. : https://github.com/brekkylab/ailoy/compare/docs/wasm...docs/wasm-suggest?expand=1

Please feel free to apply your update on this PR.

grf53:
> Yes it looks better, but I tried to fix the really necessary parts as possible, in order to make the docs concise. Moreover, if we apply this css we need to create all the screen captures again..

I've noticed that CSS is applied differently depending on the environment. I think it would be better not to provide separate guidance on how to change this. It would be better to leave it up to individuals to modify the CSS or frontend components if necessary. Thank you.

jhlee525 (approved):
👍

PR #159, opened 2025-08-11
[WASM] Support `Agent` with local models

grf53:
## Description
Supports `Agent` using local models on web environment.
- add output token processing
- fix issues
  - In web, the previous infer process must be completely completed before the next infer can be performed, so the logic is modified to process the loop without breaking it in the middle. (Maybe due to its own locking mechanism.)
  - add missed type validation(Make sure that no problem occurs even if the `"reasoning"` value is `undefined(=null_t)`.)


bindings/js-web/src/llm/config.ts
```diff
@@ -209,7 +209,7 @@ export const prebuiltAppConfig: AppConfig = {
       vram_required_MB: 1403.34,
       low_resource_required: true,
       overrides: {
-        context_window_size: 4096,
+        context_window_size: 40960,
```
khj809: Should we also update the overrided `context_window_size` for another models too?

grf53: If you mean, the other models in `prebuiltAppConfig` which are not currently in use but going to be used, I would say 'yes'. (For other environments, they are already `40960` as far as I know.)
Maybe we could adjust the value considering environment that could be run on, but just `4096` is too small to to something more meaningful than demos.

grf53: I set 0.6B and 1.7B models as `low_resource_required: true` and set their `context_window_size: 10240`,
and I set 4B and 8B models as `low_resource_required: false` and set their `context_window_size: 40960`.

`low_resource_required` is explained as below.
`@param low_resource_required: whether the model can run on limited devices (e.g. Android phone)`
We are not officially support Android phone or other 'limited devices', but this distinguishment could be useful someday. 

```javascript
[
  {
    model_id: "Qwen/Qwen3-0.6B",
    low_resource_required: true,
    overrides: {
      context_window_size: 10240,
    },
  },
  {
    model_id: "Qwen/Qwen3-1.7B",
    low_resource_required: true,
    overrides: {
      context_window_size: 10240,
    },
  },
  {
    model_id: "Qwen/Qwen3-4B",
    low_resource_required: false,
    overrides: {
      context_window_size: 40960,
    },
  },
  {
    model_id: "Qwen/Qwen3-8B",
    low_resource_required: false,
    overrides: {
      context_window_size: 40960,
    },
  },
]
```

PR #163, opened 2025-08-14
[WASM] Add utility to check webgpu supported

khj809:
- Before downloading local model files, the LocalModel and VectorStore now checks for WebGPU availability to avoid downloading models in environments where WebGPU is not supported.
- Added a new utility function, `isWebGPUSupported`, allowing users to quickly verify WebGPU compatibility.
- Updated the relevant docs.


grf53 (approved):
LGTM

PR #167, opened 2025-08-15
Add MCP logics

khj809:
This PR implements MCP functionalities including `MCPClient` and `MCPTool`.
- `MCPClient` is created via stdio, streamable-http, sse transport, and provides `.list_tools()` to get the tools available in the client.
- `MCPTool` implements `.run()` in `Tool` trait by calling the tool with provided arguments. The raw contents are converted to our Parts.

## Additional fixes
- Added `Audio` Part type which takes base64 encoded data and format (mp3 or wav).
- A tool can return multiple contents, so I changed the return type of `.run` in `Tool` trait to return `Vec<Part>`.
- Added MCP-related methods in `Agent` as we did in bindings before.
- Fixed message aggregator related logics by @jhlee525, which is temporary.

### DIFF
--- a/v2/src/agent.rs
+++ b/v2/src/agent.rs
@@ -1,7 +1,5 @@
-use std::collections::HashMap;
 use std::sync::Arc;
 
-use anyhow::anyhow;
 use futures::{Stream, StreamExt};
 use tokio::sync::Mutex;
 
@@ -15,7 +13,6 @@ pub struct Agent {
     lm: Arc<dyn LanguageModel>,
     tools: Vec<Arc<dyn Tool>>,
     messages: Arc<Mutex<Vec<Message>>>,
-    mcp_clients: HashMap<String, Arc<MCPClient>>,
 }
 
 impl Agent {
@@ -24,7 +21,6 @@ impl Agent {
             lm: Arc::new(lm),
             tools: tools.into_iter().collect(),
             messages: Arc::new(Mutex::new(Vec::new())),
-            mcp_clients: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


v2/src/cache/filesystem.rs
```diff
@@ -16,7 +16,15 @@ mod native {
...
 
-    pub async fn write(path: impl AsRef<Path>, data: impl AsRef<[u8]>) -> Result<(), String> {
+    pub async fn write(
+        path: impl AsRef<Path>,
+        data: impl AsRef<[u8]>,
+        create_parent: bool,
+    ) -> Result<(), String> {
+        let parent_dir = path.as_ref().parent().unwrap();
+        if !parent_dir.exists() && create_parent {
+            tokio_create_dir_all(parent_dir).await.unwrap();
```
jhlee525:
```suggestion
            tokio_create_dir_all(parent_dir).await.map_err(|e| format!("tokio::create_dir_all failed: {}", e.to_string()))?;
```

v2/src/agent.rs
```diff
@@ -1,18 +1,21 @@
...
-    tool::Tool,
+    tool::{MCPClient, Tool},
     value::{Message, MessageAggregator, MessageDelta, Part, Role, ToolCall},
 };
 
 pub struct Agent {
     lm: Arc<dyn LanguageModel>,
     tools: Vec<Arc<dyn Tool>>,
     messages: Arc<Mutex<Vec<Message>>>,
+    mcp_clients: HashMap<String, Arc<MCPClient>>,
```
jhlee525: Does the MCP client bind the MCP server in one-to-one way? (In other words, one MCP server per client.)

If so, could we consider putting the MCP client struct inside the tool?

v2/Cargo.toml
```diff
@@ -20,8 +21,9 @@ napi = { version = "3", features = ["tokio_rt"], optional = true }
 napi-derive = { version = "3", optional = true }
 pyo3 = { version = "0.25", features = ["extension-module"], optional = true }
 pyo3-async-runtimes = { version = "0.25", features = ["attributes", "tokio-runtime"], optional = true }
+regex = "1.11.1"
```
jhlee525: `tokenizers`(by HF) uses `onig`(non-wasm) or `fancy-regex`(wasm) for it's regex processing.

Honestely, I didn't tried this library, but these can be an option, since it is already imported in this project by `tokenizers`.

khj809: Verified that it works well with `onig`.

v2/src/agent.rs
khj809: MCP tools already have a corresponding client in `service` field.
It seems your intention is to minimize the states in agent we need to handle, so I removed `mcp_clients` and just adds the tools from it.

jhlee525 (approved):
👍

PR #171, opened 2025-08-20
update build system

jhlee525:
This PR aims to simplify the previous complex build system:

1. Removed the use of cxx-build on the Rust side. Instead, all CXX code generation is unified under CMake using `cxxbridge-cmd`.
2. Configured the build to produce `tvm_runtime.a`, so that Rust, Python, and Node no longer need to link against `tvm_runtime.so` explicitly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,7 +49,6 @@ dependencies = [
  "async-stream",
  "cmake",
  "cxx",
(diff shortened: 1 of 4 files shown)


khj809 (approved):
We need to verify that it works on Linux and Windows too, but let's believe our big brother for now.

PR #172, opened 2025-08-21
Vectorstore for V2

grf53:
## Description
Add vector store feature in v2.
- Add `DLPackTensor` as alternative of `DLManagedTensorVersioned`, opaque type from C++ 
  - `DLManagedTensorVersioned`(Raw C type) -> `ManagedTensor`(C++ interface wrapper type for cxx bridge) -> `DLPackTensor`(Rust Wrapper type)
  - `DLPackTensor` can be generated from the `DLManagedTensorVersioned` pointer address got from TVM outputs(NDArray)
- Add `VectorStore` trait as interface of various vector stores
  - API
    - Chromadb: Implement chromadb vector store with 'chromadb' crate
  - Local
    - FAISS
      - ~Implement FAISS vector store with cxx bridged types and functions.~
      - Add cxx bridge for 'FAISS Index'(`faiss::Index`)
        - `faiss::index_factory()` will work as contructor


khj809:
@grf53 Could you change the base branch to `v0.2.0`?

grf53:
> @grf53 Could you change the base branch to `v0.2.0`?

Done, my bad!

v2/src/knowledge_base/vector_store.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Debug)]
+pub struct RetrieveResult {
+    pub id: String,
+    pub document: String,
+    pub metadata: Option<Metadata>,
+    pub distance: f32,
+}
+
+#[async_trait]
+pub trait VectorStore: MaybeSend + MaybeSync {
```
grf53: `VectorStore` trait needs to be Send + Sync for the chromadb compatibility.
This can make the APIs not compatible with wasm environment, so maybe I need to replace the chromadb part to use reqwest directly or something.
I need some opinions on this.

grf53: Maybe I could apply a similar changes of [anthropic-ai-sdk](https://github.com/brekkylab/anthropic-sdk-rs/tree/main/anthropic-ai-sdk) to the [chromadb crate](https://github.com/Anush008/chromadb-rs/tree/master).

grf53:
@jhlee525 @khj809 Done merging the core. Please review this!

v2/src/utils/float.rs
```diff
@@ -0,0 +1,63 @@
...
+            let f32_exponent = (exponent as u32) + (127 - 15);
+            // 가수 비트 확장 (f16: 10비트 -> f32: 23비트)
+            let f32_fraction = (fraction as u32) << 13;
+
+            (sign as u32) << 31 | (f32_exponent << 23) | f32_fraction
+        };
+
+        f32::from_bits(result_bits)
+    }
+}
```
khj809: Could you translate all Korean comments into English?

v2/src/knowledge_base/local/faiss.rs
```diff
@@ -0,0 +1,421 @@
...
+pub struct FaissStore {
+    index: ffi::FaissIndex,
+    doc_store: DocStore,
+}
+
+impl FaissStore {
+    pub fn new(dim: i32) -> Result<Self> {
+        let index = ffi::FaissIndexBuilder::new(dim)
+            // .description(description)
+            // .metric(metric)
```
khj809: unused?

v2/src/knowledge_base/local/faiss.rs
```diff
@@ -0,0 +1,421 @@
...
+    //                         id,
+    //                         document: doc_entry.document.clone(),
+    //                         metadata: doc_entry.metadata.clone(),
+    //                         distance,
+    //                     })
+    //                 })
+    //                 .collect::<Vec<_>>()
+    //         })
+    //         .collect::<Vec<_>>())
+    // }
```
khj809: Why is this commented?

v2/shim/src/faiss_bridge.cpp
```diff
@@ -0,0 +1,230 @@
...
+//     std::copy(reconstructed_vector.begin(), reconstructed_vector.end(),
+//               std::back_inserter(result));
+
+//     return result;
+
+//   } catch (const std::exception &e) {
+//     throw std::runtime_error("Failed to get vector by ID " +
+//                              std::to_string(id) + ": " + e.what());
+//   }
+// }
```
khj809: Is this unused in favor of `get_by_ids`?

v2/src/knowledge_base/local/faiss.rs
grf53: Currently yes, unused.
I had plan to provide the way to use other types of Indexes and Metrics, but I couldn't find its necessity enough to add complexity of uses.
I added this comment because I wanted to let someone know that there's a way to do this.
Let me remove it for the sake of brevity in the merged code, if you think it is needed.

v2/src/knowledge_base/local/faiss.rs
grf53: This implementation is for a specific struct within the VectorStore trait, so adding it requires implementing it for ChromaStore, which implements the same trait, and I missed it.
The Chroma search feature also natively supports batching multiple embeddings, so I think it's worth adding it.
Let me do that.

v2/shim/src/faiss_bridge.cpp
grf53: Yes, this is an initial implementation before we have that. Now we don't need that in the `FaissIndexWrapper`, so let me remove it.

v2/src/knowledge_base/local/faiss.rs
grf53: Added.

PR #174, opened 2025-08-22
Enable Gemini

khj809:
This PR adds Gemini functionalities with the same coverage as in cpp.

In addition, some utilities are included to make multi-platform tests more simpler.
- To make a test runnable on both native and browser, just wrap it with `multi_platform_test!` macro.
- Use `crate::utils::log` to print logs in both native and browser.
- Reads `GEMINI_API_KEY` env var from `.env` at compile time, and forward it on both native and browser tests.

## Notes
It requires https://github.com/flachesis/gemini-rust/pull/7 to be merged in order to run on wasm. Until then, it's downloaded from our forked repo.

### DIFF
--- a/v2/src/utils/test.rs
+++ b/v2/src/utils/test.rs
@@ -1,0 +1,24 @@
+#[cfg(test)]
+#[cfg(not(target_arch = "wasm32"))]
+#[macro_export]
+macro_rules! multi_platform_test {
+    ($(#[$meta:meta])* $vis:vis async fn $name:ident($($args:tt)*) $body:block) => {
+        $(#[$meta])*
+        #[tokio::test]
+        $vis async fn $name($($args)*) {
+            let _ = env_logger::builder().is_test(true).try_init();
+            $body
+        }
+    };
+}
+
+#[cfg(test)]
+#[cfg(target_arch = "wasm32")]
+#[macro_export]
+macro_rules! multi_platform_test {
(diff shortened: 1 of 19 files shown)


v2/src/value/part.rs
```diff
@@ -59,11 +60,11 @@ pub enum Part {
     /// ```
     ImageURL(String),
 
-    /// Inline base64-encoded image bytes (no decoding/validation is performed).
+    /// Inline base64-encoded image bytes with MIME type (no decoding/validation is performed).
     /// ```json
     /// { "type": "image", "data": "<base64>" }
     /// ```
-    ImageData(String),
+    ImageData(String, String),
```
grf53, line 67: It would be great to apply this into the `PyPart`!
(could not build with `--features python`)

v2/src/utils/test.rs
```diff
@@ -0,0 +1,24 @@
+#[cfg(test)]
+#[cfg(not(target_arch = "wasm32"))]
+#[macro_export]
+macro_rules! multi_platform_test {
```
grf53, line 4: Really cool!

v2/src/value/part.rs
khj809: I fixed it and regenerated stubs. PTAL.

grf53 (approved):
LGTM

PR #175, opened 2025-08-22
Enable Anthropic

khj809:
This PR enables AnthropicLanguageModel.

It requires our forked [anthropic-ai-sdk](https://github.com/brekkylab/anthropic-sdk-rs/tree/main/anthropic-ai-sdk) to run on wasm32-unknown-unknown.

### DIFF
--- a/v2/src/utils/async_trait.rs
+++ b/v2/src/utils/async_trait.rs
@@ -1,0 +1,17 @@
+#[cfg(not(target_arch = "wasm32"))]
+#[macro_export]
+macro_rules! async_trait {
+    ($item:item) => {
+        #[async_trait::async_trait]
+        $item
+    };
+}
+
+#[cfg(target_arch = "wasm32")]
+#[macro_export]
+macro_rules! async_trait {
+    ($item:item) => {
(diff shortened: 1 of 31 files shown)


v2/src/utils/async_trait.rs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+#[cfg(target_arch = "wasm32")]
+#[macro_export]
+macro_rules! async_trait {
+    ($item:item) => {
+        #[async_trait::async_trait(?Send)]
+        $item
+    };
+}
```
grf53, line 17: This doesn't seem to be used. What is it for?

khj809: Ah this is not directly related to this PR. I was just testing async-trait available on both environments.
I think we can use this for your vectorstore PR, which already uses async-trait.

grf53: I was looking into this cuz it can be applied, but I couldn't find one.
Anyway, that's good news!

jhlee525 (approved):

PR #176, opened 2025-08-25
Js inferencer

jhlee525:


grf53:
I can't find any test code or configuration for this. How did you test this?

v2/src/model/local/local_language_model.rs
```diff
@@ -319,3 +394,61 @@ mod tests {
         println!("Assistant message: {:?}", assistant_msg);
     }
 }
+
+#[cfg(all(test, target_arch = "wasm32"))]
+#[cfg(test)]
+mod tests {
```
jhlee525, line 400: @grf53 I've tested with this example

```bash
wasm-pack test --chrome --release
```

v2/shim_js/src/index.ts
```diff
@@ -0,0 +1 @@
+export { init, LanguageModel } from "./language_model";
```
khj809: `init` looks quite ambiguous. We may have various initialization for such as embedding models, faiss, etc.
How about renaming it as distinguishable?

jhlee525: ok i'll do for it

v2/src/ffi/web/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::value::Part;
+
+#[wasm_bindgen]
+extern "C" {
+    // Import a JavaScript function named 'alert'
+    fn alert(s: &str);
+
+    // Import a JavaScript function from a specific module
+    #[wasm_bindgen(js_namespace = ailoy)]
+    fn log(s: &str);
```
grf53, line 12: Not sure for this. Do we have a module function named 'log' to import from JS?

v2/src/ffi/web/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+#[derive(Clone)]
+#[wasm_bindgen]
+pub struct WasmPart {
+    inner: Part,
+}
```
grf53: Do we still need this file?

jhlee525: It should be implemented. For now, we can remove these web binding files

v2/src/cache/claim.rs
```diff
@@ -0,0 +1,27 @@
+use std::any::Any;
+
+use crate::{cache::CacheEntry, dyn_maybe_send};
+
+pub struct CacheClaim {
+    pub entries: Vec<CacheEntry>,
+    pub ctx: Option<Box<dyn_maybe_send!(Any)>>,
```
grf53, line 24: I could understand that `CacheClaim` implies `CacheEntry`s.
But what is the role of `ctx` here?

jhlee525: Usually ctx is None, but in case of additional information should be delivered, (from `claim_files` to `try_from_cache`) it can contains free-form information.

It is used in `LocalLanguageModel`, which should remember the member's claims (chat_template + tokenizer + inferencer).

grf53: Through verbal discussion, we confirmed the following context:
1. `try_from_contents()` needs to be converted to async (returning a `BoxFuture`) for the wasm runtime.
2. `try_from_contents()` has been changed to use owned typed (moved) instead of mutable ref type of `CacheContents`. (This allows for a clearer representation of the data transfer and destruction flow without using Arc instead of mutable ref.)
3. `ctx` was created to allow `try_from_contents()` to use owned typed instead of mutable ref type of `CacheContents`. This requires passing a list of required files for each element to reconstruct `CacheContents` for each child element.
4. Since the TryFromCache trait is not limited to a specific situation such as 'LocalLanguageModel', the content that needs to be passed from `claim_files()` to `try_from_contents()` cannot be fixed, so a ctx containing the `Any` type is used.

grf53: I think this _context_ should be written as the docstring for `CacheClaim` to deliver the intention of this.
Would you please add that docstring?

khj809 (approved):
LGTM!

PR #180, opened 2025-09-01
change language model trait to mutable ref

jhlee525:
previous trait spec:
```
pub trait LanguageModel: MaybeSend + MaybeSync + 'static {
    fn run(
        self: Arc<Self>,
        msg: Vec<Message>,
        tools: Vec<ToolDesc>,
    ) -> BoxStream<'static, Result<MessageOutput, String>>;
}
```

to

```
pub trait LanguageModel: MaybeSend + MaybeSync {
    fn run<'a>(
        self: &'a mut Self,
        msg: Vec<Message>,
        tools: Vec<ToolDesc>,
    ) -> BoxStream<'a, Result<MessageOutput, String>>;
}
```

So that it is not necessary to force Arc<dyn LanguageModel> as a container type(+ do not force stream lifetime to static)

### DIFF
--- a/v2/src/agent.rs
+++ b/v2/src/agent.rs
@@ -12,15 +12,18 @@ use crate::{
 };
 
 pub struct Agent {
-    lm: Arc<dyn LanguageModel>,
+    lm: Box<dyn LanguageModel>,
     tools: Vec<Arc<dyn Tool>>,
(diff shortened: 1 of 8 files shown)


grf53 (approved):
Looks good, but env var part would cause conflicts with another PR(#182).

PR #181, opened 2025-09-01
Add Faiss JS shim

khj809:
This PR adds Faiss JS shim of Faiss prebuilt libs -> cpp bridge(Emscripten) -> JS wrapper -> Rust bridge(wasm_bindgen) -> Rust wrapper (combines cxx and js interfaces into single interface).

## Build Emscripten bridge
- For prebuilt libs (`libfaiss.a`, `libblas.a`, `liblapack.a` and headers), we use the asset from https://github.com/brekkylab/faiss-web/releases/tag/v1.12.0-wasm.
- For linking `libblas.a` and `liblapack.a`, we need `libFortranRuntime.a` from `flang-wasm` image, so we run a build command inside docker container with mounting csrc and destination directory.
- `csrc/faiss_bridge.cpp` includes similar interfaces as defined in `shim/faiss_bridge.hpp`.
- After build, `faiss_bridge.js` and `faiss_bridge.wasm` are generated under `shim_js/src/faiss` with type definition `faiss_bridge.d.ts`. Note that `faiss_bridge.js` and `faiss_bridge.wasm` are gitignored.
- Build script is located under `shim_js/scripts/build_faiss_bridge.sh`, and this is executed via `npm run build:faiss`.

## Build JS wrapper
- `shim_js/src/faiss/index.ts` implements a simple `FaissIndexWrapper` class which wraps operations of `FaissIndex`, and async initialization function `init_faiss_index_wrapper`. This file should include minimal logics to keep the binding chain as simple as possible.
- Bundled `index.js` is built via `npm run build:ts`, and it also copies `src/faiss/faiss_bridge.wasm`. (See `rollup.config.js`)
- Added simple tests in `shim_js/tests/faiss.test.ts`.

## Build Rust bridge (via wasm_bindgen)
- `src/ffi/js_bridge.rs` includes bindings for `FaissIndexSearchResult`, `FaissIndexWrapper`, and `init_faiss_index_wrapper`.
- The above bindings are consumed in `FaissIndex` at `src/ffi/faiss_wrap.rs`, which is a unified interface across cxx and js.
- Originally I tried to reuse `FaissMetricType` and `FaissIndexSearchResult` defined in `cxx_bridge.rs`, but found it's not possible. So I managed to redefine them in `faiss_wrap.rs` and implemented `From` and `Into` for conversion between them.
- On the top of `js_bridge.rs`, I used `raw_module` instead of `module`, since using `module` cannot load `faiss_bridge.wasm` properly.

## Tests
- Now `src/knowledge_base/local/faiss.rs` is available on both native and wasm environments.
- Verified that all test cases in `src/knowledge_base/local/faiss.rs` are passed.


v2/shim_js/src/faiss/faiss_bridge.d.ts
grf53, line 1: As I understood, this file is generated by adding `--emit-tsd faiss_bridge.d.ts` to `emcc`.
if then, what is the reason why this file is git-managed?

khj809: This is used to ensure that `src/faiss/index.ts` is type-safe, regardless of the actual generated js and wasm.
The same reason applies to `binding/python/ailoy/_core.pyi` and any type stubs generated in the future.

v2/shim_js/src/faiss/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  get_metric_type(): keyof FaissMetricType {
+    const index_metric = this.index.get_metric_type().value;
+    const metric_types = Object.keys(
+      FaissIndexWrapper.wasm!.FaissMetricType
+    ) as (keyof FaissMetricType)[];
+    return metric_types.find(
+      (key) =>
+        FaissIndexWrapper.wasm!.FaissMetricType[key]?.value === index_metric
+    )!;
+  }
```
grf53: Here's my understanding. Please verify this:
We can get the faiss index's metric type only as its value(as `uint8_t`), so we need to dereference here to get key from the value.

khj809: Yes that's correct. We have a mapping from metric key to value, and this is for figuring out the key from value reversely.

v2/shim_js/src/index.ts
```diff
@@ -3,3 +3,5 @@ export {
   EmbeddingModel,
 } from "./embedding_model";
 export { init as init_language_model, LanguageModel } from "./language_model";
+export { init as init_faiss_index_wrapper, FaissIndexWrapper } from "./faiss";
```
grf53: I think the naming scheme is different with the C++ -> Rust shim.

### C++ -> Rust
`faiss::index`(cpp; FAISS)
-> `FaissIndexWrapper`(cpp; shim)
-> `FaissIndexWrapper`(rs; cxx_bridge)
-> `FaissIndex`(rs; API)

### C++ -> Typescript
`faiss::index`(cpp; FAISS)
-> `FaissIndex`(cpp; shim_js)
-> `FaissIndex`(ts; embind)
-> `FaissIndexWrapper`(ts; API)

I wanted the type name used in the user API to be `FaissIndex`, so I used `FaissIndexWrapper` as the 'inner' name.

I think this should be unified anyway, and I still think it's preferable for the type in the user API to be `FaissIndex` rather than `FaissIndexWrapper`.

Please let me know your opinion.

grf53: Hmm, maybe it's because the ts API `FaissIndexWrapper` is still an 'inner' name... as used like `js_bridge::FaissIndexWrapper` similar to `cxx_bridge::FaissIndexWrapper`, then it makes sense I think.

v2/shim_js/tests/faiss.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    expect(searchResults.indexes[1]).to.be.equal(1n);
+
+    // distance of 0 should be less than 1
+    expect(searchResults.distances[0]).to.be.lessThan(
+      searchResults.distances[1]
+    );
+
+    // clear
+    vs.clear();
+    expect(vs.get_ntotal()).to.be.equal(0);
```
grf53:
```suggestion
    expect(vs.get_ntotal()).to.be.equal(0n);
```
Also here.

v2/shim_js/tests/faiss.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // add two vectors
+    vs.add_vectors_with_ids(
+      new Float32Array([
+        0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0,
+      ]),
+      2,
+      new BigInt64Array([0n, 1n])
+    );
+    expect(vs.get_ntotal()).to.be.equal(2);
```
grf53:
```suggestion
    expect(vs.get_ntotal()).to.be.equal(2n);
```
`vs.get_ntotal()` returns `bigint` so it needs to be `2n`.

v2/shim_js/src/index.ts
khj809: Yes this was a big headache.. Since JS Faiss has one more level than C++, it was inevitable to use the term "Wrapper" one more time. If there's better naming than current, I'm happy to apply that.

grf53: we could choose some other names
- `faiss_index_t` (as C++ class)
- `FaissIndexInner` (as type of `inner` in `struct FaissIndex`)

AI suggested some names with suffix `Cxx`, `Js`, `Embind`, but I don't like them a lot.

grf53 (approved):
I think this PR is enough to be merged, regardless of changing the names of types of `FaissIndex`.

v2/shim_js/src/index.ts
khj809: Based on the above suggestion, I aligned the name of classes directly facing Rust bridge as `FaissIndexInner`.
For Emscripten (more inner) class, I named it as `FaissIndexHandle`.

PR #182, opened 2025-09-01
fix: use `LazyLock` and `option_env!` instead of `env!` for the tests

grf53:
## Description
We're currently using the `env!` macro to provide an API key at compile time for use in tests.
`env!` causes a compile error if the corresponding environment variable doesn't exist at compile time. However, since the code is compiled when building with the test target, the corresponding environment variable must be present, even if empty, even when building tests unrelated to the environment variable. This negatively impacts the unit testability of modules.

Therefore, we're retrieving the environment variable at compile time, but using `std::sync::LazyLock` to generate an error when attempting to use it, and maintaining it static.


khj809 (approved):
Thanks for fixing this!

jhlee525 (approved):
👍

PR #185, opened 2025-09-04
Python binding

khj809:
This PR defines Python binding via pyo3, with as many type hints as possible in stub(pyi) file.

### Primitive types
- [x] Part
- [x] Message
- [x] MessageOutput
- [x] MessageAggregator
- [x] ToolDesc
### Cache Progress
- [x] CacheResult (base class used for generic type)
- [x] CacheProgress
### Models
- [x] LocalLanguageModel
- [x] OpenAILanguageModel
- [x] GeminiLanguageModel
- [x] AnthropicLanguageModel
- [x] XAILanguageModel
- [x] LocalEmbeddingModel
### Tools
- [x] BuiltinTool
- [x] PythonFunctionTool
- [x] PythonAsyncFunctionTool
- [x] MCPTool
### VectorStore
- [x] FaissStore
- [x] ChromaStore
### Agent
- [x] Agent

## Future works
- Add pytest cases for each binding
- Add more docstrings
- Add rich `__repr__` for each binding


v2/src/model/api/openai_legacy.rs
```diff
@@ -251,7 +251,7 @@ mod tests {
                     (
                         "unit",
                         ToolDescArg::new_string()
-                            .with_enum(["Celcius", "Fernheit"])
+                            .with_enum(["Celsius", "Fernheit"])
```
grf53:
```suggestion
                            .with_enum(["Celsius", "Fahrenheit"])
```

v2/src/value/tool_desc.rs
```diff
@@ -1,505 +1,78 @@
...
+            "Get current temperature".into(),
+            parameters,
+            Some(returns),
+        )
+        .unwrap();
+
         let serialized = {
-            let expected = r#"{"type":"function","function":{"name":"temperature","description":"Get current temperature","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city name"},"unit":{"type":"string","description":"Default: Celcius","enum":["Celcius","Fernheit"]}},"required":["location"]},"return":{"type":"number"}}}"#;
-            let actual = serde_json::to_string(&original).unwrap();
+            let expected = r#"{"name":"temperature","description":"Get current temperature","parameters":{"type":"object","properties":{"location":{"type":"string","description":"The city name"},"unit":{"type":"string","description":"Default: Celsius","enum":["Celsius","Fahrenheit"]}},"required":["location"]},"returns":{"type":"number"}}"#;
```
grf53, line 70: According to transformers, the json schema of function return values ​​comes with the key `"return"`, not `"returns"`.

https://github.com/huggingface/transformers/blob/c81f426f9a1a6bb860532201d3efb2f5c0bd087c/src/transformers/utils/chat_template_utils.py#L370

In this PR, there's a change to using `"returns"` instead of `"return"` across the board, including using `returns` instead of `r#return` in `ToolDesc`. Is there a specific reason for this?

v2/bindings/python/tests/test_multimodal.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@pytest.mark.parametrize(
+    "agent", ["openai", "gemini", "anthropic", "xai"], indirect=True
+)
+async def test_image_base64_input(agent: ai.Agent):
+    with urlopen(image_url) as resp:
+        image_data = resp.read()
+        b64_data = base64.b64encode(image_data).decode("utf-8")
+        img_part = ai.Part.ImageData(b64_data, "image/jpeg")
```
grf53, line 78: This may have nothing to do with Python bindings.(maybe based on Rust code)
we can create a part with `ai.Part.ImageData(..., ...)`(`Part::ImageData{..., ...}` in Rust) like this line, and also with `ai.Part.new_image_data(..., ...)`(`Part::new_image_data(..., ...)` in Rust) using class method.

This looks redundant to me. Why we have these two way to do this? Is there any difference I couldn't find?

v2/src/agent.rs
```diff
@@ -106,18 +106,16 @@ impl Agent {
...
             !tool_name.starts_with(format!("{}--", client_name).as_str())
         });
-
         Ok(())
     }
 
     pub fn run<'a>(
         &'a mut self,
-        user_message: impl Into<String>,
+        contents: Vec<Part>,
```
grf53, line 124: Now `agent.run()` gets vector of parts instead of just a string.
With the abstraction at the binding level, you can still pass a string to `agent.run()` in Python.(and also vector of parts)
Would we keep this abstraction in the binding level, regardless of binding languages?

v2/src/value/message.rs
```diff
@@ -112,69 +112,38 @@ pub struct Message {
...
 
-    pub fn with_reasoning(self, reasoning: impl Into<String>) -> Self {
-        Message {
-            role: self.role,
-            reasoning: reasoning.into(),
-            contents: self.contents,
-            tool_calls: self.tool_calls,
-            tool_call_id: self.tool_call_id,
-        }
+    pub fn with_role(&mut self, role: Role) -> Self {
```
grf53: `with_role()` no longer generates a message, and it now just modifies a message like other functions. I prefer this.

v2/src/value/message.rs
```diff
@@ -97,64 +99,61 @@ pub enum Role {
...
-        Message {
-            role: Some(role),
-            reasoning: String::new(),
-            contents: Vec::new(),
-            tool_calls: Vec::new(),
-        }
+    pub fn with_role(&mut self, role: Role) -> Self {
+        self.role = Some(role);
+        self.clone()
     }
```
grf53:
```suggestion
    pub fn with_role(mut self, role: Role) -> Self {
        self.role = Some(role);
        self
    }
```

I'm concerned that these modified `with_xxx()` functions clone the Message objects. These functions are designed to be used chained multiple times, as needed, and since they involve heap-allocated variables like Vector, I believe continuous duplication could lead to inefficiencies.
Instead of borrowing a mutable reference, it would be better to obtain an owned mutable object, modify it, and return it, thus making it move not clone.

v2/bindings/python/tests/test_multimodal.py
khj809: That's a good catch. I think we don't need to have `new_*` factories since we already have constructor for each part type. I'll remove them.

v2/src/agent.rs
khj809: IMO, the goal of v2 binding is to cover as many use cases in v1 as possible, and that's why I kept `agent.run()` to accept both str and list of parts. I'm also planning to apply this to another language bindings if applicable.

grf53: The question was whether the part that takes a string or a list of parts doesn't need to be made at the Rust level (or whether it would be very difficult to do this).
If Rust always takes a list of parts and the binding code wraps the string as a vector of parts, then every binding would have to implement it differently. This might be an easier approach, but I'm not familiar with it, so I'm asking.

v2/bindings/python/tests/test_multimodal.py
grf53: The `new_*()` methods are removed in bindings, but there are still `new_*()` in Rust.
As I see, `StyledPart` variants seem to need `new_*()` methods to abstract its style, while `Part` don't need to.
But it looks like a good idea to keep `new_*()` in `Part` too, as it allows us to maintain consistency in how `StyledPart` and `Part` are created.
(you can just resolve this conversation again after reading this.)

v2/bindings/python/README.md
```diff
@@ -0,0 +1,90 @@
...
+pip install maturin
+
+# This generates `_core.cpython-3xx-darwin.so` under `ailoy/`
+maturin develop
+```
+
+### Generate wheel
+
+```bash
+python -m build -w
```
grf53:
```suggestion
maturin build --out ./dist
```
This command requires to install another build dependency `build`(like `pip install build`), so suggest to use just `maturin`(which is already required in former step) also for building wheels.
`--out` option designates the directory to locate the built wheel files and its default is `<build_dir>/wheels/`(`<root>/target/wheels/`), so it can be modified or removed as wanted.

.vscode/settings.json
```diff
@@ -142,5 +143,8 @@
     "rust-analyzer.procMacro.enable": true,
     "rust-analyzer.runnables.extraEnv": {
         "RUST_LOG": "debug"
-    }
+    },
+    "rust-analyzer.rustfmt.extraArgs": [
+        "+nightly"
```
grf53: This looks accidentally added.

v2/src/model/local/local_language_model.rs
```diff
@@ -14,14 +14,14 @@ use crate::{
     value::{FinishReason, Message, MessageOutput, Part, Role, ToolDesc},
 };
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
jhlee525: Could you clear why `LocalLanguageModel` to be Clonable?

.vscode/settings.json
khj809: I applied it on rustfmt PR, but didn't pull it on this PR yet. Now I merged and resolved the conflicts.

v2/bindings/python/tests/test_multimodal.py
khj809: Okay I'll keep the `new_*()` factories in Rust part.

v2/src/agent.rs
khj809: In Rust, maybe we can make it accepting enum which holds either a plain string or a vector of parts, but it seems quite cumbersome for now, and not considered as in the scope of this PR.
For another strong-typed binding languages such as Java or Go, there should be a language-specific approach to accomplish it, but not sure for now (that's why I put "if applicable" at the end).

v2/bindings/python/README.md
khj809: Okay I'll update it.

v2/src/model/local/local_language_model.rs
khj809: Now I removed `Clone` from all LanguageModels and Agent.

v2/src/value/tool_desc.rs
grf53: ~~Statement of the 'returning value' is only in the mentioned code of transformers.
I couldn't find any mention of a return value in any other examples or description about tools, including the [document of transformers](https://huggingface.co/docs/transformers/v4.51.3/en/chat_templating_writing#tool-definitions).
So I suggest to remove statement of returning values in the tool descriptions of ours.~~

It's hard to argue that there are never cases where you need to provide a description for a return value (such as when a tool returns a complex nested object, and the LLM needs to correctly interpret the return value and use it for subsequent operations).
I've seen a few additional attempts to include the returning value in the tool description, often using "returns" as well (though overall, the cases to include them aren't very common).
So, including the description of return value with the "returns" key seems to be more usable than previously thought.

grf53:
I believe the changes in this PR are now sufficiently complete to be merged.
However, given the anticipated significant fundamental changes related to Message, I'm not sure if merging it before then is a good idea.
A decision on this will likely require discussion.

v2/Cargo.toml
```diff
@@ -19,18 +19,21 @@ async-trait = "0.1.89"
 base64 = "0.21"
 chromadb = { git = "https://github.com/brekkylab/chromadb-rs.git", branch = "master" }
 derive_builder = "0.20.2"
+downcast-rs = "2.0.2"
+dyn-clone = "1.0.20"
```
jhlee525: Do we still need this crate?

jhlee525 (commented):
In fact, too many ideas have been applied, and among them there are parts that are difficult to review.

Including ideas in a PR that have not been sufficiently discussed in issues, or that cannot be reasonably justified, makes the review process difficult.

Since the definition of message is currently changing, many parts will likely be revised again.

However, this PR on its own could be merged to some extent.

Although many parts will change again later, I ask for your understanding.

v2/Cargo.toml
khj809: Removed it.

jhlee525 (approved):

PR #187, opened 2025-09-08
Formatted log

grf53:
## Description
Add macros wrapping log functions can be used like `print!()`.
macro names could be changed.


v2/src/utils/log.rs
```diff
@@ -12,6 +12,13 @@ pub fn debug(s: impl Into<String> + std::fmt::Display) {
     web_sys::console::debug_1(&s.to_string().into());
 }
 
+#[macro_export]
+macro_rules! debug_fmt {
```
khj809: I think it's better to remove `_fmt` suffix, assuming that users won't get confused with external `log` crate.

PR #188, opened 2025-09-10
Fix bugs in parts

grf53:
## Description
There seemed to be a mismatch between the `Part` codes and its tests.
To sync these up, modified those a little.

Please take a look at this, and let me know if there are some problems or misunderstandings.

### DIFF
--- a/v2/src/value/part.rs
+++ b/v2/src/value/part.rs
@@ -518,6 +518,7 @@ impl<'a> Serialize for PartFunctionRef<'a> {
     }
 }
 
+#[derive(Debug)]
 struct PartFunctionOwned {
     name: String,
     arguments: String,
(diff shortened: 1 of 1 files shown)


v2/src/value/part.rs
```diff
@@ -753,14 +754,14 @@ impl<'de> de::Visitor<'de> for PartVisitor {
             }
             (ty, Some((k, v)), _, _) if k == "function" => {
                 if let Some(ty) = ty {
-                    style.text_type = ty;
+                    style.function_type = ty;
```
grf53, line 757: Please double check this should be interpreted as `function_type`.

v2/src/value/part.rs
```diff
@@ -753,14 +754,14 @@ impl<'de> de::Visitor<'de> for PartVisitor {
...
                 }
                 style.function_field = k;
                 Ok(StyledPart::new_function_string(v).with_style(style))
             }
             (ty, _, id, Some((k, v))) if k == "function" => {
                 if let Some(ty) = ty {
-                    style.text_type = ty;
+                    style.function_type = ty;
```
grf53, line 764: Please double check this should be interpreted as `function_type`.

PR #192, opened 2025-09-24
fix: bug on post-processing of local embedding model result

grf53:
## Description
During the post-processing of the local embedding model results in C++, an error occurred that caused the f16 value to be interpreted as f32, resulting in a malformed f16 value returned in the C++ stage. Therefore, because the incorrect f16 value was being used, the actual value was problematic.

This PR checks the DataType of the model result and returns a result in the appropriate format, allowing Rust to use valid f16 or f32 values. The conversion process for actually using the f16 values ​​is performed in Rust.


khj809 (approved):
It works well. Thanks!

PR #193, opened 2025-09-24
message schema updated

jhlee525:


v2/src/value/part.rs
```diff
@@ -391,94 +328,73 @@ impl<'de> Deserialize<'de> for Part {
...
-    FunctionToolCall {
+    Text(String),
+    VerbatimFunction(String),
+    Function {
         id: Option<String>,
         name: String,
         arguments: String,
     },
-    TextRefusal(String),
+    ParsedFunction {
```
grf53, line 338: To maintain consistency with `Part::Function`, how about renaming it as follows?
Function -> RawFunction(or sth else)
ParsedFunction -> Function

v2/src/value/message.rs
```diff
@@ -20,29 +20,137 @@ pub enum Role {
...
+        } = self;
+        Self {
+            role,
+            id,
+            think,
+            contents,
+            tool_calls,
+            signature,
+        }
+    }
```
grf53: When using the wither pattern, it's more intuitive and widely used to accept `self` as `mut self`, mutate a specific member (`self.member = member`), and then return that `self`.

Using a temporary `self` as a mutable value within the function seems like an acceptable choice, as it doesn't affect external immutability.

Even if `self` remains immutable, `Self {member, ..self}` is more concise.

v2/src/value/message.rs
```diff
@@ -20,29 +20,137 @@ pub enum Role {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Message {
     pub role: Role,
-    pub parts: Vec<Part>,
+    pub id: Option<String>,
+    pub think: String,
```
grf53, line 24: Just an opinion. For me, `thinking` looks more suitable for this then `think`.

v2/src/value/message.rs
```diff
@@ -20,29 +20,137 @@ pub enum Role {
...
+    pub contents: Vec<Part>,
+    pub tool_calls: Vec<Part>,
+    pub signature: Option<String>,
```
grf53, line 27: Signature is a concept bound for thinking, so how about make `think`(or `thinking`) as `(String, Option<String>)`(or `{content: String, signature: Option<String>}`)?

v2/src/value/part.rs
```diff
@@ -41,46 +41,28 @@ impl TryFrom<String> for PartImageColorspace {
...
-        signature: Option<String>,
+    Text(String),
+    Function {
+        id: Option<String>,
+        name: String,
+        arguments: Value,
     },
-    TextContent(String),
-    ImageContent {
+    Value(Value),
```
grf53, line 50: `Value` as `Part` seems like it could be interpreted in quite various ways, so I'm not sure exactly what this was intended for. What is the purpose of this?

grf53 (commented):
I left a few reviews, PTAL.

(Using `part_to_value` makes the code much more readable. Great!)

v2/src/value/message.rs
jhlee525: Yep. I already applied in following changes.

v2/src/value/message.rs
jhlee525: I'm not sure but if we want to use builder-like pattern, using destructive self would be helpful.

For example, destructive self (`fn with_a(self, a: A) -> Self`) can use API like this
```
let config = Config::new()
    .with_a(10)
    .with_b("hello");
```

However, if using reference self, declaration(holding ownership) and configuration must be split
```
let config = Config::new();
config.with_a(10).with_b("hello");
```

It’s a matter of preference, if you believe later syntax is more intuitive, than please tell me

v2/src/value/part.rs
jhlee525: In following works, I've changed it like this

```
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
#[serde(untagged)]
pub enum PartDeltaFunction {
    Verbatim(String),
    WithStringArgs {
        name: String,
        #[serde(rename = "arguments")]
        args: String,
    },
    WithParsedArgs {
        name: String,
        #[serde(rename = "arguments")]
        args: Value,
    },
}
```

v2/src/value/message.rs
jhlee525: I agree, but it related to serde logic

If we want to serialize this struct like this, it should be placed as an independent field.
```
{"thinking": "...", "signature": "..."}
```

Otherwise, it is placed under "think" field
```
{"thinking": ["...", "..."]
```
or
```
{"thinking": {text: "...", signature: "..."}
```

v2/src/value/part.rs
jhlee525: In some cases such as with the Gemini API, tool calls and responses need to be provided as a Value, not as a JSON-serialized string. Although it is possible to store them as a "JSON-serialized" string, I chose to define a type instead.

In summary, it’s for object-like values, where one doesn’t want to just holding a JSON string.

v2/src/value/message.rs
jhlee525: I've changed logic to `mut self` pattern. PTAL.

v2/src/value/message.rs
grf53: I see. It makes sense.

v2/src/value/part.rs
grf53: In cases like Gemini, which accept objects as they are, using the value itself is a good approach. Having the `Part` retain the `Value` itself seems like a good approach.
The current marshaled form of the `Part::Value` variant is identical to the `Part::Text` variant, which made me confuse about its intention a little.

This feature seems to stem from the use of object-type values ​​as arguments for Tool Calls or Tool Responses.(Otherwise, it is unlikely that a complete object-type value will be included as part of a user message.) While Tool Calls are already sufficiently covered by `Part::Function`'s `"arguments"` being of type `Value`, Tool Responses are unlikely to be covered by a single `Value`. This is because most APIs also require specifying the call ID (or function name).
e.g.)
```
{
"type": "function_call_output",
"call_id": "call_DF3wZtLHv5eBNfURjvI8MULJ",
"output": "{\"temperature\": 25,\"unit\": \"celsius\"}"
}
```

Therefore, instead of `Part::Value`, I suggest defining `Part::ToolResponse` or `Part::FunctionResponse` as a combination of the call id and the result `Value`, instead of `Part::Value(Value)`.

```
pub enum Part {
    ...
    FunctionResponse {
        id: Option<String>,
        result: Value,
    },
    ...
}
```

grf53: And then, maybe `Part::Function` would be better to be `Part::FunctionCall`...?

v2/src/value/dialect/gemini.rs
```diff
@@ -244,46 +195,33 @@ mod tests {
...
-        ));
+        let msg = Message::new(Role::User).with_contents([
+            Part::text("Explain me about Riemann hypothesis."),
+            Part::text("How cold brew is different from the normal coffee?"),
+        ]);
         let marshaled = Marshaled::<_, GeminiMarshal>::new(&msg);
         assert_eq!(
             serde_json::to_string(&marshaled).unwrap(),
-            r#"{"role":"user","contents":[{"parts":[{"text":"Explain me about Riemann hypothesis"},{"text":"How cold brew is different from the normal coffee?"}]}]}"#
+            r#"{"role":"user","contents":[{"parts":[{"text":"Explain me about Riemann hypothesis."},{"text":"How cold brew is different from the normal coffee?"}]}]}"#
```
grf53, line 205: I apologize that I mistook to set up this 'expected' result. It should be like one of #194.
Let me modify this on #194, after merging this PR.

grf53: And if there are any mistakes for other dialects, I'll fix them as I test them against the actual API in #194.

v2/src/value/part.rs
jhlee525: I understand your concern or confusion. However, there are some difficulties.

### API function responses cannot include an id

In the case of tool calls, multiple calls can be contained in a single assistant message, for example:

```json
{
  "role": "assistant",
  "tool_calls": [
    { "type": "function", "id": "tc1", "function": { "name": "...", "args": "..." } },
    { "type": "function", "id": "tc2", "function": { ... } }
  ]
}
```


However, for tool messages, a single message can only carry one id. Therefore, tool calls must be resolved by supporting multiple messages, like this:

```json
{ "role": "tool", "id": "tc1", "content": ... }
{ "role": "tool", "id": "tc2", "content": ... }
```

Of course, there are APIs like Gemini that allow multiple tool responses in a single message. While it is disappointing that each API uses a different approach, we still need to implement support for these differences. That is why the current Message struct has an id field at the root level, which is only used when the role is tool.

Since we don’t necessarily need to be 100% compatible with the chat-completion API, this is something we can consider.

### Parts are designed to represent only data types (not intention)

As discussed earlier, each part of a message carries two dimensions of information:

- Data type: text, image, etc.
- Intention: reasoning, tool call, tool response, contents, etc.

In the current design, the data type is handled by the Part enum, while intention is represented by struct fields of Message (such as thinking, content, etc.).

Based on this principle, “tool calling” and “tool response” are closer to intention rather than data type. That’s why these are represented as a Value rather than as a separate Part variant.

---

It's really painful to choose something to fulfill every APIs... I think we need further discussion to how to implement it.

grf53: I understood that the Parts represents the data types.

Maybe this difference of opinion came from considering target API.
The implementation of this PR seems to consider OpenAI ChatCompletion API at most, meanwhile I was considering the other ones.

Since I don't think the scope of this PR extends beyond the chatcompletion API, it seems like this PR is good enough to merge.
I understood the pain point and the design principle now, so let me try to apply those into the APIs in the following works.

PR #194, opened 2025-09-25
Add Config

grf53:
## Description
This PR suggests `Config` to be `Marshal`ed to construct configurations for LLM inference.
And added `FinishReason` which represents normalized 'finish reasons' of various models and APIs.
Additionally, the result type of `LanguageModel` became `MessageOutput` again, that contains a `MessageDelta` and optional `FinishReason`.

To see definitions of `Config` and `FinishReason`, take a look at [`config.rs`](https://github.com/brekkylab/ailoy/pull/194/files#diff-2e46c25e798b2ae4a9b9475e6e188f21642b07b3873a9adfb37b21f150418e51).
And for the example of using those, see [`value/dialects/openai.rs`](https://github.com/brekkylab/ailoy/pull/194/files#diff-4b579ffc613b0ad0106288b9f4c6a8ac1589b619e06d6f1685414d3f76bb8461) and [`model/api/openai.rs`](https://github.com/brekkylab/ailoy/pull/194/files#diff-96dbf2b3b9558c834b4b23b99044edfd42221f0c7306798e6953f671315f8a0a).

> Note that `model/api/openai.rs` now contains OpenAI Responses API, while `model/api/chat_completion.rs` contains ChatCompletion API.

### DIFF
--- a/v2/src/model/api/chat_completion.rs
+++ b/v2/src/model/api/chat_completion.rs
@@ -4,7 +4,6 @@ use crate::{
 };
 
 pub fn make_request(
-    // model_name: &str,
     api_key: &str,
     msgs: Vec<Message>,
     tools: Vec<ToolDesc>,
@@ -57,7 +56,7 @@ pub fn handle_event(evt: ServerSentEvent) -> Vec<MessageDelta> {
 mod tests {
     use crate::{
         model::{LanguageModel as _, sse::SSELanguageModel},
-        value::{LMConfigBuilder, Delta},
+        value::{Delta, LMConfigBuilder},
     };
 
     const OPENAI_API_KEY: &str = "";
--- a/v2/src/model/api/gemini.rs
+++ b/v2/src/model/api/gemini.rs
@@ -7,7 +7,6 @@ use crate::{
 };
 
 pub fn make_request(
-    // model_name: &str,
     api_key: &str,
     msgs: Vec<Message>,
     tools: Vec<ToolDesc>,
@@ -52,7 +51,6 @@ pub fn handle_event(evt: ServerSentEvent) -> MessageOutput {
     let Ok(j) = serde_json::from_str::<serde_json::Value>(&evt.data) else {
         return MessageOutput::default();
     };
-    // println!("j: {:?}", j);
 
     let Some(candidate) = j.pointer("/candidates/0") else {
         return MessageOutput::default();
--- a/v2/src/model/api/sse.rs
+++ b/v2/src/model/api/sse.rs
@@ -5,7 +5,7 @@ use futures::StreamExt as _;
 use crate::{
     model::LanguageModel,
     utils::{BoxStream, MaybeSend, MaybeSync},
-    value::{LMConfig, FinishReason, Message, MessageOutput, ToolDesc},
+    value::{FinishReason, LMConfig, Message, MessageOutput, ToolDesc},
 };
 
 #[derive(Clone, Debug)]
@@ -81,7 +81,6 @@ impl SSELanguageModel {
             // Anthropic models
             todo!()
         } else if model.starts_with("gemini") {
-            // todo!()
             // Gemini models
             SSELanguageModel {
                 make_request: Arc::new(
--- a/v2/src/model/api/openai.rs
+++ b/v2/src/model/api/openai.rs
@@ -1,13 +1,12 @@
 use crate::{
     model::sse::ServerSentEvent,
     value::{
-        LMConfig, FinishReason, Marshaled, Message, MessageDelta, MessageOutput, OpenAIMarshal,
+        FinishReason, LMConfig, Marshaled, Message, MessageDelta, MessageOutput, OpenAIMarshal,
         OpenAIUnmarshal, ToolDesc, Unmarshaled,
     },
 };
 
 pub fn make_request(
-    // model_name: &str,
     api_key: &str,
     msgs: Vec<Message>,
     tools: Vec<ToolDesc>,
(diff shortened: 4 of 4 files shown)


v2/src/value/config.rs
```diff
@@ -0,0 +1,176 @@
...
+    #[default]
+    Disable,
+    Enable,
+    Low,
+    Medium,
+    High,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize)]
+pub enum OutputSchema {
```
jhlee525: How about using `Grammar`?

`OutputSchema` can be ambiguous (e.g. MCP tool output schema)

v2/src/value/config.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+// #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "python",
+    pyo3_stub_gen_derive::gen_stub_pyclass_complex_enum
+)]
+#[cfg_attr(feature = "python", pyo3::pyclass(eq))]
+pub struct Config {
```
khj809: The name `Config` seems too broad. How about giving a more specific name such as `LMConfig`?

v2/src/model/api/chat_completion.rs
```diff
@@ -0,0 +1,124 @@
...
+                .iter()
+                .map(|v| Marshaled::<_, ChatCompletionMarshal>::new(v))
+                .collect::<Vec<_>>()
+        );
+    }
+
+    reqwest::Client::new()
+        .request(
+            reqwest::Method::POST,
+            "https://api.openai.com/v1/chat/completions",
```
khj809, line 31: Assuming that this is used by various chat-template-compatible providers, I think this url should be configurable.

v2/src/model/api/sse.rs
```diff
@@ -45,40 +49,60 @@ fn drain_next_event(buf: &mut Vec<u8>) -> Option<ServerSentEvent> {
...
+    handle_event: Arc<dyn Fn(ServerSentEvent) -> MessageOutput + MaybeSend + MaybeSync>,
 }
 
 impl SSELanguageModel {
     pub fn new(model: impl Into<String>, api_key: impl Into<String>) -> Self {
         let model = model.into();
         let api_key = api_key.into();
-        let (make_request, handle_event) = if model.starts_with("gpt") || model.starts_with("o") {
+
+        let ret = if model.starts_with("gpt") || model.starts_with("o") {
```
khj809, line 65: Even though it's not significantly problematic for now, I'm really worried about determining API provider based on model names only. For example, if we need to support Azure OpenAI, it's impossible to decide which provider to use between original OpenAI or Azure.
How about using enums explicitly, such as `ModelProvider::OpenAI("gpt-4o")` or `APIModel::Gemini("gemini-2.5-flash")`?

v2/src/model/api/chat_completion.rs
grf53: Makes sense! Let me consider this in the following works.

v2/src/value/config.rs
grf53: Applied!

v2/src/model/api/sse.rs
grf53: I agree with that we need some better way to distinguish the model providers.
Let me think about this along with the service providers through Chat Completion API.

v2/src/model/api/gemini.rs
```diff
@@ -0,0 +1,183 @@
...
+    use crate::{
+        debug,
+        model::{LanguageModel as _, sse::SSELanguageModel},
+        value::{Delta, LMConfigBuilder},
+    };
+
+    // static GEMINI_API_KEY: LazyLock<&'static str> = LazyLock::new(|| {
+    //     option_env!("GEMINI_API_KEY")
+    //         .expect("Environment variable 'GEMINI_API_KEY' is required for the tests.")
+    // });
```
khj809, line 95: Should this be restored?

v2/src/model/api/gemini.rs
```diff
@@ -0,0 +1,183 @@
+use crate::{
+    model::sse::ServerSentEvent,
+    value::{
+        FinishReason, GeminiMarshal, GeminiUnmarshal, LMConfig, Marshaled, Message, MessageDelta,
+        MessageOutput, ToolDesc, Unmarshaled,
+    },
+};
+
+pub fn make_request(
+    // model_name: &str,
```
khj809: If there's no plan to use `model_name`, how about removing them?

v2/src/value/config.rs
grf53: Applied!

v2/src/model/api/gemini.rs
grf53: Yes, I prefer using environment variables instead of manual modification, it is too cumbersome to check the commits not to include this kind of secrets.

v2/src/model/api/gemini.rs
grf53: Right, `model_name` is now included in the `Config` as `model`, so there would be no need to use that argument.
There still were some code dregs like this. So I just clean those up.

khj809 (approved):

v2/src/model/api/gemini.rs
grf53: I didn't even test this yet, but I have a will to restore this. Maybe this would be restored in the following work.

PR #196, opened 2025-09-26
Enable API models

grf53:
## Description
Enable API LLM models entirely based on the revised message format

### ~~Note about ChatCompletion API~~
~~Within this PR, the ChatCompletion API is used exclusively for XAI grok.~~

### `APISpecification`
`APIProvider` is now `APISpecification` which represents the ways(embracing data schemas, methods, endpoints) to use an API. It's like all you see in a Swagger page.
To mention specially, 'base URL' of APIs are not included in the API specification. (Imagine a situation where the same server is deployed to multiple domains.)

Now `ChatCompletion`, `OpenAI`, `Gemini`, `Claude` are the variants of `APISpecification`, and there can be some 'aliases' of those as variants(`Responses`->`OpenAI`, `Grok`->`ChatCompletion`(with different default URL)).

Based on this, `xai.rs` can be removed because grok API can be regarded as alias of `ChatCompletion` spec.

### Additional
- Added: `impl From<i32> for Value` to support `i32` in `to_value!` (use `3` instead of `3 as i64`)
- Added: use `option_env!()` to give API keys in API tests

## Remaining works
- Add more tests for APIs in various situation
- (Not sure) Extend ChatCompletion API for more API Providers


v2/src/model/api/anthropic.rs
```diff
@@ -0,0 +1,148 @@
...
+        );
+    }
+
+    reqwest::Client::new()
+        .request(reqwest::Method::POST, api_model.endpoint())
+        .header("x-api-key", api_model.api_key.clone())
+        .header("anthropic-version", "2023-06-01")
+        .header(reqwest::header::CONTENT_TYPE, "application/json")
+        .header(reqwest::header::ACCEPT, "text/event-stream")
+        .body(body.to_string())
```
khj809: Anthropic requires `anthropic-dangerous-direct-browser-access: true` header when it's used in browser.
Please see https://github.com/brekkylab/anthropic-sdk-rs/commit/60fef3dcaf50c02124a5a9de834b37300c3d94a9 for my code used before.

v2/src/model/api/mod.rs
```diff
@@ -25,3 +25,37 @@ struct RequestConfig {
...
+#[derive(Debug, Clone)]
+pub enum APIProvider {
+    OpenAI,
+    Google,
+    Anthropic,
+    XAI,
+}
+
+#[derive(Debug, Clone)]
+pub struct APIUsage {
```
jhlee525: Can we rename the name? `APIUsage` can be ambiguous

grf53: Decided to use it destructed.(`APIUsage` doesn't exist now.)

jhlee525 (approved):

jhlee525:
Miscellaneous comment: The name `APIProvider` might cause some confusion.

When a user is using a 3rd-party API (for example, the OpenRouter API with an OpenAI-compatible schema), our recommendation is to set APIProvider to APIProvider::ChatCompletion and simply override the URL. From this perspective, the name APIProvider might cause some confusion (it could be mistaken as meaning that the API is provided directly by OpenAI). So, renaming it to something like `APISchema` or `APIStyle` might be a good idea.

That said, it’s not strictly necessary.

v2/src/model/api/anthropic.rs
```diff
@@ -319,8 +414,71 @@ impl Unmarshal<MessageDelta> for AnthropicUnmarshal {
...
+                .map(|v| Marshaled::<_, AnthropicMarshal>::new(v))
+                .collect::<Vec<_>>()
+        );
+    }
+
+    reqwest::Client::new()
+        .request(reqwest::Method::POST, url)
+        .header("x-api-key", api_key)
+        .header("anthropic-version", "2023-06-01")
+        .header("anthropic-dangerous-direct-browser-access", "true")
```
khj809: I think it's not a good idea to always use this header (which is mentioned as "dangerous"), even for native environments.

v2/src/model/api/stream.rs
```diff
@@ -92,38 +81,44 @@ impl StreamAPILangModel {
...
+                handle_event: Arc::new(super::anthropic::handle_event),
+            },
+            APIProvider::XAI => StreamAPILangModel {
+                name: model.into(),
+                make_request: Arc::new(
+                    move |msgs: Vec<Message>, tools: Vec<ToolDesc>, req: RequestConfig| {
+                        super::chat_completion::make_request(&url, &api_key, msgs, tools, req)
+                    },
+                ),
+                handle_event: Arc::new(super::chat_completion::handle_event),
```
khj809: XAI has a slightly different `finish_reason` compared to the standard chat completion. It contains `end_turn` which is considered same as `stop`, but using the existing unmarshal logic may not handle it correctly.
So I think we should make chat completion unmarshal to be customizable, and create a separate unmarshal for XAI.

v2/src/model/api/anthropic.rs
grf53: could you check this https://github.com/brekkylab/ailoy/pull/196/commits/76631269b63197bcb489602cc78cec3fe96b3b56 out?

khj809: Looks good

v2/src/model/api/stream.rs
grf53: I tested the APIs with three `mod api_tests` test functions which contain the finish reason checks.

I checked the grok API [documentation](https://docs.x.ai/docs/api-reference#chat-completions) as you mentioned, and it is as follows.
> Finish reason. `"stop"` means the inference has reached a model-defined or user-supplied stop sequence in 
`stop`. `"length"` means the inference result has reached models' maximum allowed token length or user defined value in `max_tokens`. `"end_turn"` or `null` in streaming mode when the chunk is not the last.

According to the explanation, in the case of `end_turn`, it seems like it should be processed as `None` since the stream is not finished and needs to continue, but it is difficult to understand because the term does not match well and even I've never seen any response with that. (while `stop` is clearly explained and I checked that it comes out as a response.)

I couldn't find another explanation for `end_turn`, so I don't know in what case `end_turn` comes out. Are you suggesting that it should be processed like `"end_turn" => FinishReason::Stop()`?

khj809: Yes I understood it as you mentioned, but the description looks strange for me too.
If there's no critical failure in several tests, then it's okay to skip this for now.

Not only focusing on XAI, I wanted to say that any providers added in the future may not strictly follow the chat completion format as we expected. For example, [OpenRouter also says](https://openrouter.ai/docs/api-reference/overview) that "with a few small differences". So I thought it would be better to prepare for such situations.

grf53: I've been thinking about how to accommodate those differences.
I don't think it's a good practice to "standardize" the many APIs that claim to be "ChatCompletion-like" to handle such minor differences. Instead, it seems more appropriate to incorporate these differences into the existing ChatCompletion API handling code so that they can be handled as they are discovered.

As the first practice of that, Let me add a part that makes the grok API's `"finish_reason": "end_turn"` connect to `FinishReason::Stop`.

And maybe the `APIProvider` need a little modification based on that principle.

khj809: Okay that seems also a decent option, assuming that the differences are really subtle.

grf53: Please check the latest commits with the description of `APISpecification` in the [comment](https://github.com/brekkylab/ailoy/pull/196#issuecomment-3355929415).

v2/src/model/language_model.rs
```diff
@@ -106,9 +106,13 @@ impl LangModel {
         })
     }
 
-    pub fn new_stream_api(model_name: impl Into<String>, api_key: impl Into<String>) -> Self {
+    pub fn new_stream_api(
+        provider: APISpecification,
```
khj809:
```suggestion
        spec: APISpecification,
```

PR #198, opened 2025-09-28
Revised

jhlee525:
This PR builds on top of #197 and includes additional implementations made before it was merged.

### Introducing `InferConfig`
The existing `LMConfig` struct contained parameters that were not used by `LocalLanguageModel`. To address this, I defined a `InferConfig`, which is a subset of `LMConfig`, includes only the configuration strictly required for inference. `LMConfig` is renamed to `RequestInfo`, concentrates to "serializing" API requests. Any new naming is welcomed.

### Python bindings for the value module
I restored the Python bindings that had previously been implemented for the value module. Unlike the older approach - which placed them under a separate `ffi::py` module - the bindings are now integrated directly into the native implementation modules. This also allows us to leverage macros more extensively, thereby minimizing boilerplate.

### Unifying dialects and API requests
Previously, the `value::dialect` module handled marshaling / unmarshaling for every dialects, while API requests were implemented under model::api. Since these are closely related, I consolidated them into `model::api` so that related code resides in the same place, easy for read.


v2/src/model/language_model.rs
```diff
@@ -2,19 +2,53 @@ use std::{any::TypeId, sync::Arc};
...
+    #[default]
+    Plain,
+    JSON,
+    JSONSchema(String),
+    Regex(String),
+    CFG(String),
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct InferenceConfig {
```
khj809, line 33: What I understood from your description is that `InferenceConfig` is corresponding to local lang models and `RequestInfo` is corresponding to API lang models.
Then I think the better name is a distinguishable one between API and Local, so how about naming it as `LocalLMInferConfig` and `ApiLMInferConfig`?

v2/bindings/python/ailoy/__init__.py
```diff
@@ -1,36 +1,37 @@
...
-    Part,
-    Role,
-    ToolDesc,
-    VectorStoreAddInput,
-    VectorStoreGetResult,
-    VectorStoreRetrieveResult,
-    XAILanguageModel,
-)
-from ailoy._patches import PythonAsyncFunctionTool, PythonFunctionTool
+from ._core import *;
```
khj809: Star imports is generally considered as a bad practice. How about restoring explicit imports?

v2/src/model/local/local_language_model.rs
```diff
@@ -270,159 +353,158 @@ mod tests {
...
+        let mut strm = model.run(msgs, Vec::new(), InferenceConfig::default());
         while let Some(out) = strm.next().await {
             let out = out.unwrap();
             println!("{:?}", out);
-            if let Some(msg) = agg.update(out) {
-                println!("{:?}", msg);
-            }
+            // if let Some(msg) = agg.update(out) {
+            //     println!("{:?}", msg);
+            // }
```
khj809, line 362: Unused?

v2/src/model/language_model.rs
```diff
@@ -2,19 +2,53 @@ use std::{any::TypeId, sync::Arc};
...
+    pub think_effort: Option<ThinkEffort>,
+
+    pub temperature: Option<ordered_float::OrderedFloat<f64>>,
+
+    pub top_p: Option<ordered_float::OrderedFloat<f64>>,
+
+    pub max_tokens: Option<i32>,
+
+    pub grammar: Option<Grammar>,
```
khj809, line 42: I think this field can also be used in API models supporting structured outputs (Regex and CFG might not be supported though).
How about adding this to API model's config and mark the features related to this option as todo?

v2/src/model/language_model.rs
jhlee525: In exact,

* InferenceConfig: Configurations common in inference (regardless of local & API), since the nature of LM inference
* RequestInfo: A dataclass that facilitates marshaling/unmarshal of API models.

v2/src/model/local/local_language_model.rs
```diff
@@ -1,85 +1,169 @@
...
+    ) -> BoxFuture<'static, Result<CacheClaim, String>> {
+        LocalLanguageModelImpl::claim_files(cache, key)
     }
 
-    pub fn disable_reasoning(&self) {
-        self.chat_template.disable_reasoning();
+    fn try_from_contents(contents: CacheContents) -> BoxFuture<'static, Result<Self, String>> {
+        Box::pin(async move {
+            let mut body = LocalLanguageModelImpl::try_from_contents(contents).await?;
+            let (tx, mut rx) = mpsc::channel(8);
```
khj809: Is it safe to use size 8 channel for requests? It seems there's no mutex in LocalLanguageModelImpl's inferencer, so I'm worried about multiple inferences being executed at the same time.

v2/bindings/python/ailoy/__init__.py
jhlee525: According to [PEP guildeline](https://peps.python.org/pep-0008/#imports), it says:

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn’t known in advance).

I considered this case as a "one defensible use case", which means republish module to public APIs. So I think it does not violate PEP guildline.

However I'm open to manual re-export if it has clear benefits.

v2/src/model/language_model.rs
khj809: Thanks for the explanation. Then it seems `RequestInfo` is not an entity exposed to users so it does not matter what the name is. That said, I prefer `Params` or `Config` rather than `Info` in this case, which seems more "configurable" options.

v2/src/model/local/local_language_model.rs
jhlee525: Yep. It can be removed. I'll fix it to the latest logic

v2/src/model/language_model.rs
jhlee525: It is intended behavior. It can be used both local & API, that's why it is in `InferenceConfig`.

v2/src/model/local/local_language_model.rs
jhlee525: The current logic flow is as follows:

```
let (tx, mut rx) = mpsc::channel(8); 
// A new inference request is sent from `LanguageModel` to `LanguageModelImpl`. 
// Each token generated by `LanguageModelImpl` uses the `tx_resp` and `rx_resp` channels.
```

Thus, the channel size here can be defined as the

> maximum number of inference requests that can be sent to `LanguageModelImpl` simultaneously.

In a typical scenario, only one request should be sent at a time. However, for some reason, if the `LanguageModel` is cloned and multiple handles send requests concurrently.

In this case, the following issue arises:

> separate requests(whose message histories are not shared) are processed concurrently, leading to KV cache collisions. Execution is still possible, but it causes performance degradation and is therefore not a recommended usage pattern.

If 8 requests are sent simultaneously, this clearly indicates an improper practice.

v2/src/model/language_model.rs
jhlee525: Ok. `RequestConfig` seems reasonable (and maybe more intuitive).

However, this struct is originally introduced by @grf53 , we should ask him for whether it changes original intention.

v2/bindings/python/ailoy/__init__.py
khj809: Okay that makes sense. But could you remove the semicolon?

v2/src/model/language_model.rs
khj809: Yes. What I said is that it should also exist in `RequestConfig` so we can marshal it to e.g. [`response_format`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format).

jhlee525: Oh, What I do was just renaming. It might be active work by @grf53 , but not high priority.

v2/bindings/python/ailoy/__init__.py
jhlee525: 👌

v2/src/model/api/sse.rs
```diff
@@ -49,30 +49,56 @@ fn drain_next_event(buf: &mut Vec<u8>) -> Option<ServerSentEvent> {
 
 #[derive(Clone)]
 pub struct SSELanguageModel {
+    name: String,
```
grf53: Just a negligible comment:

`name` is set as 'model name' of APIs when `SSELanguageModel` is created.
and `name` is used to fill `model` of `RequestInfo`.
I think, it would be better to just use `model` in `SSELanguageModel` then.

v2/src/model/local/local_language_model.rs
khj809: If the execution is still possible but it's improper situation, then why not explicitly prevent such situation by using length 1 channel or similar approaches?

v2/src/model/api/sse.rs
grf53: Maybe it was: "`SSELanguageModel` is considered as a `model` itself, so another attribute named `model` in that structure is inconsistent."

which kinda makes sense, I think...

Just in case this struct is no longer called 'model', then consider using `model` instead of `name`.

v2/src/value/part.rs
```diff
@@ -332,6 +328,9 @@ pub enum PartDelta {
         #[serde(rename = "function")]
         f: PartDeltaFunction,
     },
+    Value {
+        value: Value,
+    },
```
grf53, line 333: Just a curiosity: We don't need to define behaviors of `PartDelta::Value` in `aggregate`?

v2/src/model/local/local_language_model.rs
jhlee525: Great idea. channel with single length can be a trick for preventing this situation. I'll try for it.

grf53 (approved):
For me, it's good enough to be merged.
If other issues are resolved, you may merge it without my separate approval.

v2/src/model/api/sse.rs
jhlee525: model may be better naming. I'll change for it.

v2/src/value/part.rs
jhlee525: Currently, it is not implemented since there is no use case for it.

But we can implement dictionary merge logic, like `dict.update` in python (however, it can only be applied if both lhs and rhs are "object(dictionary) typed `Value`".

We can later choose a strategy and implement it.

PR #199, opened 2025-09-28
Langmodel revised

jhlee525:


v2/src/model/language_model.rs
```diff
@@ -19,71 +22,298 @@ pub enum ThinkEffort {
...
+            Py::new(
+                py,
+                LangModel {
+                    inner: LangModelInner::Local(inner),
+                },
+            )
+        }
+
+        #[classmethod]
+        #[pyo3(name = "CreateSSE", signature = (model_name, api_key))]
```
khj809: Using the term `SSE` for public API seems quite unintuitive.
Using `Stream` would be better, or assuming it's used for API models only, just using `API` seems enough too.

v2/src/tool/mcp2/mod.rs
```diff
@@ -0,0 +1,144 @@
+use std::sync::Arc;
```
khj809, line 1: Are you planning to integrate `mcp2` module into existing `mcp` in the next PR?

v2/src/model/language_model.rs
jhlee525: Ok I'll change it to `StreamAPILangModel`

v2/src/tool/mcp2/mod.rs
jhlee525: Yes. I tried. I think consolidating `MCPTransports` & `MCPClient`. Making `MCPClient` as public and providing static API(constructors) may be more intuitive to python users. (Enum automatically translated to nested class, which can be not intuitive for non-rust users)

However, I realized that local version and wasm version uses different logics. I was looking for how to make them available in wasm also...

pushing this file is accidental. You can ask me to remove this file.

khj809: I understood. It's okay to keep this file for now.

PR #200, opened 2025-10-01
Change MCP public API, etc.

khj809:
Previously we used `MCPTransport` as a public interface to initialize MCP clients and get tools, but there was an opinion that the term "Transport" is just a matter of configs and it's not natural for it to have a client initialization method. Hence, this PR changes to explicitly use the struct `MCPClient`, and initialize it via `MCPClient::from_stdio()` or `MCPClient::from_streamable_http()`.

This PR also includes:
- Fix wasm32 build by removing `jsonschema` default features and updating some obsolete test cases.
- Local LangModel used `tokio::spawn` for running inference handling loop, but it's not available on wasm32. So replaced it to `wasm_bindgen_futures::spawn_local` for wasm32 environment.
- Introduce `maybe_send_sync` proc macro to use on the situation where `MaybeSend` and `MaybeSync` are unavailable (they're only available on defining new traits).
- Remove unnecessary `stringify` from `KnowledgeTool`.


jhlee525 (approved):
👍

jhlee525:
A question: when we should use ` + MaybeSend + MaybeSync` or `[#[maybe_send_sync]`? It's about our policy.

khj809:
> A question: when we should use ` + MaybeSend + MaybeSync` or `#[maybe_send_sync]`? It's about our policy.

Use `+ MaybeSend + MaybeSync` for defining new "traits":
```rust
pub trait Knowledge: std::fmt::Debug + MaybeSend + MaybeSync {}
```

Use `#[maybe_send_sync]` for defining new "types":
```rust
#[maybe_send_sync]
type ToolFunc = dyn Fn(Value) -> Value;
```

jhlee525:
It can cause confusion for several reasons:

1. Usually, a derive macro can be applied not only to types but also to traits, yet there is no clear explanation or justification for "why it is recommended to `types` only derives macro for implementing `Send`" for other contributors.
2. We already have the `dyn_maybe_sync` macro, but there is no clear reason or explanation of why it should be abandoned.

While this PR is still acceptable to merge, I would appreciate it if you could understand my difficulties. I heard that the reason is your aptitude, but it is difficult for me to explain why "a trait uses `MaybeSend + MaybeSync`, while a function or type uses `#[dyn_maybe_sync]`", solely because of someone's aptitude.

A clearer policy will be needed in the future.

khj809:
@jhlee525 I agree that there's no reason to use `#[maybe_send_sync]` only for type aliases, so I tried to make it available for trait declarations too, and verified that it works well.

So the policy would be:
1. Think of using `#[maybe_send_sync]` always if possible.
2. For some complicated cases where it's not easy to use proc macro, it's allowed to use `MaybeSend` and `MaybeSync` directly.

PR #203, opened 2025-10-01
Model manager CLI

khj809:


jhlee525 (commented):
I think we need another command something like "generate", generating "remote cache-compatible" directory from the local directory.

khj809:
> I think we need another command something like "generate", generating "remote cache-compatible" directory from the local directory.

I added `--to-local-path` option in `upload` command, which does exactly what you mentioned. PTAL.

PR #204, opened 2025-10-02
Value node binding

jhlee525:
In this PR, I added Node bindings to the value module, as a first job for the Node.js integration.

The implementation was written to be as close as possible to the Python version, while leveraging derive to avoid boilerplate code.

### Problem

Currently, the enum variants are exported in uppercase, but the convention is to use lowercase. This needs to be addressed.

```typescript
export type Part =
  | { type: "Text"; text: string }
  | { type: "Function"; id?: string; f: PartFunction }
  | { type: "Value"; value: any }
  | { type: "Image"; image: PartImage };

export type PartDelta =
  | { type: "Text"; text: string }
  | { type: "Function"; id?: string; f: PartDeltaFunction }
  | { type: "Value"; value: any }
  | { type: "Null" };
```

### DIFF
--- a/v2/bindings/nodejs/src/ailoy_core.d.ts
+++ b/v2/bindings/nodejs/src/ailoy_core.d.ts
@@ -7,247 +7,70 @@ export type EmbeddingModel = LocalEmbeddingModel;
 export type VectorStore = FaissVectorStore | ChromaVectorStore;
 /* @ts-ignore */
 export type Tool = BuiltinTool | MCPTool | JsFunctionTool;
-export declare class Agent {
-  constructor(lm: LanguageModel, tools?: Array<Tool>);
-  run(message: Array<JsPart | string> | JsPart | string): AgentRunIterator;
-  addTool(tool: Tool): void;
-  addTools(tools: Array<Tool>): void;
-  removeTool(toolName: string): void;
-  removeTools(toolNames: Array<string>): void;
-}
-export type JsAgent = Agent;
+export type FinishReason =
+  | { type: "Stop" }
+  | { type: "Length" }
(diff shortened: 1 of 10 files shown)


v2/bindings/nodejs/src/ailoy_core.d.ts
```diff
@@ -7,248 +7,71 @@ export type EmbeddingModel = LocalEmbeddingModel;
...
-
-export interface CacheProgress {
-  comment: string;
-  current: number;
-  total: number;
+export type FinishReason =
+  | { type: "Stop" }
+  | { type: "Length" }
+  | { type: "ToolCall" }
+  | { type: "Refusal"; field0: string };
```
khj809, line 14: Can we use field name other than `field0`?

v2/src/value/part.rs
```diff
@@ -606,3 +613,6 @@ mod py {
         }
     }
 }
+
+#[cfg(feature = "nodejs")]
+mod node {}
```
khj809: Is this used in the future?

v2/bindings/nodejs/src/ailoy_core.d.ts
jhlee525: It's because of node auto binding mechanism... I think in case of "enum with tuple argument", automatic binding results to the field name`fieldN`.

We'll need to change rust signature to
```rust
Refusal { refusal: String }
```

cc. @grf53

v2/src/value/part.rs
jhlee525: I think we can remove

khj809 (approved):

Issue #206, opened 2025-10-02
Consolidating public structs


jhlee525:
As of v2, `Model`, `Tool`, `Knowledge`, and `Agent` are currently defined as traits, with multiple implementations for each trait.

From the "user" structs(holding these as a variable), these are typically wrapped as:

```
Arc<Mutex<T>>
```
or
```
Arc<T>
```
However, considering cross-language bindings (Python / Node / WASM), this usage is not ideal: It requires additional boilerplate code. It complicates the public API surface.

### Proposed Change

We propose introducing new unified structs, following the principle of **1 public struct per 1 module**.

Each concept will expose a single pub struct. I hope this policy will reduce the number of structs that need to be bound, and concise API.

- [x] LanguageModel
- [x] EmbeddingModel
- [ ] Tool
- [ ] Knowledge
- [x] Agent

PR #212, opened 2025-10-02
Struct version of tool & knowledge

jhlee525:
Please refer #206 for more details

### DIFF
--- a/v2/src/knowledge/base.rs
+++ b/v2/src/knowledge/base.rs
@@ -5,9 +5,12 @@ use anyhow::Result;
 use serde::{Deserialize, Serialize};
 
 use crate::{
+    knowledge::{CustomKnowledge, VectorStoreKnowledge},
+    model::EmbeddingModel,
     to_value,
     tool::ToolBehavior,
     value::{ToolDesc, Value},
+    vector_store::VectorStore,
 };
 
 type Metadata = serde_json::Map<String, serde_json::Value>;
@@ -20,7 +23,7 @@ pub struct KnowledgeRetrieveResult {
 
 #[maybe_send_sync]
 #[multi_platform_async_trait]
-pub trait Knowledge: std::fmt::Debug {
+pub trait KnowledgeBehavior: std::fmt::Debug {
     fn name(&self) -> String;
 
     async fn retrieve(&self, query: String) -> Result<Vec<KnowledgeRetrieveResult>>;
@@ -28,7 +31,7 @@ pub trait Knowledge: std::fmt::Debug {
 
 #[derive(Clone)]
 pub struct KnowledgeTool {
-    inner: Arc<dyn Knowledge>,
+    inner: Arc<dyn KnowledgeBehavior>,
     desc: ToolDesc,
 }
 
@@ -43,7 +46,7 @@ impl std::fmt::Debug for KnowledgeTool {
 }
 
 impl KnowledgeTool {
-    pub fn from(knowledge: impl Knowledge + 'static) -> Self {
+    pub fn from(knowledge: impl KnowledgeBehavior + 'static) -> Self {
         let default_desc = ToolDesc {
             name: format!("retrieve-{}", knowledge.name()),
             description: Some("Retrieve the relevant context from knowledge base.".into()),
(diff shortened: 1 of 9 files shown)


v2/src/knowledge/base.rs
```diff
@@ -107,45 +110,99 @@ impl Tool for KnowledgeTool {
...
+                            document: "Ailoy enables running LLMs in local environment easily."
+                                .into(),
+                            metadata: None,
+                        },
+                    ];
+                    Ok(documents)
+                }
+                .boxed()
+            }),
+        ));
```
khj809: Since we now have a separate `custom_knowledge.rs`, I think it would be better to move this test to there, just like vectorstore knowledge.

khj809 (commented):
LGTM with minor comment.

v2/src/knowledge/custom_knowledge.rs
```diff
@@ -0,0 +1,55 @@
...
+            name: name.into(),
+            f,
+        }
+    }
+}
+
+#[multi_platform_async_trait]
+impl KnowledgeBehavior for CustomKnowledge {
+    fn name(&self) -> String {
+        "about-ailoy".into()
```
grf53:
```suggestion
        self.name.clone()
```

v2/src/knowledge/base.rs
```diff
@@ -107,45 +110,99 @@ impl Tool for KnowledgeTool {
...
-                    metadata: None,
-                },
-            ];
-            Ok(documents)
+    pub fn with_top_k(self, top_k: usize) -> Self {
+        match self.inner {
+            KnowledgeInner::VectorStore(knowledge) => Self {
+                inner: KnowledgeInner::VectorStore(knowledge.with_top_k(top_k)),
+            },
+            inner => Self { inner },
```
grf53: Why do we create a new `Knowledge` instance with the same `inner` here, instead of returning moved `self` as is?

v2/src/knowledge/base.rs
```diff
@@ -107,45 +110,99 @@ impl Tool for KnowledgeTool {
...
-                    document: "Ailoy supports Python, Javascript and Rust.".into(),
-                    metadata: None,
-                },
-                KnowledgeRetrieveResult {
-                    document: "Ailoy enables running LLMs in local environment easily.".into(),
-                    metadata: None,
-                },
-            ];
-            Ok(documents)
+    pub fn with_top_k(self, top_k: usize) -> Self {
```
grf53: This function appears to work as if `with_top_k()` (returning only the top k results) is working for all Knowledge functions, but it doesn't actually do anything for CustomKnowledge, raising questions about whether this is a proper interface.

#### Alternatives
1. Leave it as is
2. Raise an error instead of doing nothing
3. Add logic equivalent to `with_top_k()` for CustomKnowledge (e.g., returning only the top k results retrieved through the actual operation)

v2/src/knowledge/vector_store_knowledge.rs
```diff
@@ -165,7 +166,9 @@ mod tests {
             let mut agent_guard = agent.lock().await;
             // Example of customizing with_stringify
```
grf53: `with_stringify` Looks outdated.

grf53: And also: https://github.com/brekkylab/ailoy/blob/8a9ac6b6ddfd4d4e0ffa85669b78e17dc7c8abdd/v2/src/knowledge/base.rs#L43

v2/src/knowledge/custom_knowledge.rs
jhlee525: 👍

v2/src/knowledge/base.rs
jhlee525: In this logic, `self` is already destructed while matching `self`, resulting to compile error.

v2/src/knowledge/base.rs
jhlee525: I modified the API similar to `LanguageModel`. The `LanguageModel` takes `InferenceConfig` as a parameter for the `infer` function. Likewise, `top_k` is now treated as a parameter of `Knowledge`. Therefore, the `Knowledge::run` function has been changed to take `top_k`.

v2/src/knowledge/vector_store_knowledge.rs
jhlee525: 👍

v2/src/knowledge/base.rs
khj809: The reason why I made Knowledge's `.retrieve()` not to take `top_k` is that there's no way to control the top_k while agent is running (the current agent code seems using hardcoded 1).
By using `with_top_k` on Knowledge instantiation time, users can at least configure desired `top_k` values in advance.

jhlee525: It makes sense. Currently there is no way to handle this parameters. (also `InferenceConfig`) I have no idea how to make it configurable.

This should be our future works.

jhlee525: @khj809 Can we merge as-is in this PR, and later find a way to handle this problem?

khj809: Yes go ahead.

PR #213, opened 2025-10-07
Modify chat template

grf53:
## Description
While updating the chat template file, we'll update any code that requires additional modifications.

- Add tests for chat template behavior.
- Add `"preserve_order"` feature to minijinja.
- Change `{enable/disable}_reasoning()` to `{enable/disable}_thinking()`.
- Add the `think_effort` setting. (Used by gpt-oss, etc.)


v2/src/model/local/chat_template.rs
```diff
@@ -39,19 +40,26 @@ impl ChatTemplate {
...
+        let mut v = self.do_thinking.lock().unwrap();
         *v = true;
     }
 
-    /// Only affects to hybrid reasoning models
-    pub fn disable_reasoning(&self) {
-        let mut v = self.do_reasoning.lock().unwrap();
+    /// Only affects to hybrid thinking models
+    pub fn disable_thinking(&self) {
+        let mut v = self.do_thinking.lock().unwrap();
```
khj809, line 64: It seems setting `think_effort` also enables `do_thinking`. Then should disabling `do_thinking` unset `think_effort` too?

v2/src/model/local/chat_template.rs
```diff
@@ -96,3 +106,188 @@ impl TryFromCache for ChatTemplate {
...
+
+    use ailoy_macros::multi_platform_test;
+
+    use super::*;
+    use crate::{
+        to_value,
+        value::{Part, Role, ToolDesc, ToolDescBuilder},
+    };
+
+    // const TARGET_INDEX: i32 = -1;
```
khj809: unused?

grf53: This was added by mistake. I was using it to test only certain indices, but I think it should have been removed.

Speaking of which, A loop is currently being used to test multiple scenarios, and I think parameterizing it would allow me to see better results if some of the cases fail. What are your thoughts on this? I'm considering using a crate called `yare`.

khj809: I was also thinking about using test parametrization, and `yare` seems one of the good solutions. Although it seems being unmaintained during 10 months, I couldn't find any comparable alternatives..

v2/src/model/local/chat_template.rs
grf53: It makes a lot of sense.

As the variety of local models increases, the details may change, but for now, we'll consider an empty string as the default for `think_effort`(In `gpt-oss`, you should consider changing `think_effort` to `"medium"` when an empty string is entered.), so we'll add a section that sets it to an empty string. For `set_think_effort`, we'll enable thinking only when the string is non-empty to prevent unintended behavior.

v2/src/model/local/chat_template.rs
```diff
@@ -96,3 +109,186 @@ impl TryFromCache for ChatTemplate {
...
+            if !think_effort.is_empty() {
+                template.set_think_effort(think_effort);
+                // template.enable_thinking();
+            } else {
+                template.disable_thinking();
+            }
+            let prompt = template.apply(msgs, tools, true);
+            debug!("<{}>\n{}", i, prompt.as_ref().unwrap());
+            assert_eq!(prompt.unwrap().as_str(), expected);
+            // }
```
khj809: Could you clean the unused comments?

v2/src/model/local/chat_template.rs
grf53: we have another option: rstest(~~https://rstest.rs/~~ https://docs.rs/rstest/latest/rstest/). (sorry I linked wrong one)
In terms of activity and functional extensibility, rstest seems better, but I considered `yare` prior to that one due to the concise codes.

v2/src/model/local/chat_template.rs
grf53: Maybe I missed to exclude them when I commited... Thanks for pointing this out!

PR #214, opened 2025-10-07
Add wasm binding for values

khj809:
This PR adds WebAssembly (WASM) bindings for the core components in the `value` module, as outlined in #210.
While complete parity between the generated type definitions for WASM and Node.js is not possible due to binding library limitations, this implementation aims to make them as consistent as possible.

## Implementations
- Similar to other language bindings, `Value` can be converted to and from any primitive JavaScript value, and is typed as `any`.
- Most structs and enums use `tsify` to take advantage of `serde-wasm-bindgen`’s type generation support.
- WASM bindings can be generated using: `wasm-pack build --out-dir ./bindings/web/src --out-name ailoy-web --no-pack`. Note that this command creates a .gitignore file that must be manually removed afterward. (There’s an ongoing request to disable this behavior: [wasm-pack#1408](https://github.com/drager/wasm-pack/pull/1408), though it hasn’t been addressed by maintainers.)

## Changes
- The internal representation of Bytes was changed from `Vec<u8>` to `serde_bytes::ByteBuf`, which is natively supported by `tsify` and `serde-wasm-bindgen`.
- Added `"constEnum": false` in the NAPI configuration so that simple string enums are generated as string literals for both WASM and Node.js.
- Updated the `FinishReason` and `Grammar` enums from tuple variants to struct variants to ensure their generated types align across WASM and Node.js targets.
- Removed the obsolete `header.d.ts` in nodejs binding.

### DIFF
--- a/v2/bindings/web/src/ailoy-web.d.ts
+++ b/v2/bindings/web/src/ailoy-web.d.ts
@@ -6,19 +6,6 @@
  * *This API requires the following crate features to be activated: `ReadableStreamType`*
  */
 type ReadableStreamType = "bytes";
-export interface ToolDesc {
-    name: string;
-    description: string | undefined;
-    parameters: Value;
-    returns: Value | undefined;
-}
-
-export type Bytes = Uint8Array;
-
-export type Value = undefined | boolean | number | number | number | string | Record<string, any> | Value[];
-
-export type APISpecification = "ChatCompletion" | "OpenAI" | "Gemini" | "Claude" | "Responses" | "Grok";
-
 export type PartDelta = { type: "text"; text: string } | { type: "function"; id: string | undefined; function: PartDeltaFunction } | { type: "value"; value: Value } | { type: "null" };
 
 export type PartDeltaFunction = { type: "verbatim"; text: string } | { type: "with_string_args"; name: string; arguments: string } | { type: "with_parsed_args"; name: string; arguments: Value };
@@ -75,6 +62,19 @@ export interface InferenceConfig {
 export type Grammar = { type: "plain" } | { type: "json" } | { type: "jsonschema"; schema: string } | { type: "regex"; regex: string } | { type: "cfg"; cfg: string };
 
 export type ThinkEffort = "disable" | "enable" | "low" | "medium" | "high";
+
+export type APISpecification = "ChatCompletion" | "OpenAI" | "Gemini" | "Claude" | "Responses" | "Grok";
+
+export interface ToolDesc {
+    name: string;
+    description?: string;
+    parameters: Value;
+    returns?: Value;
+}
+
+export type Bytes = Uint8Array;
+
+export type Value = undefined | boolean | number | number | number | string | Record<string, any> | Value[];
 
 export class IntoUnderlyingByteSource {
   private constructor();
--- a/v2/src/value/tool_desc.rs
+++ b/v2/src/value/tool_desc.rs
@@ -10,8 +10,10 @@ use crate::value::Value;
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
 pub struct ToolDesc {
     pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
     pub parameters: Value,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub returns: Option<Value>,
 }
 


v2/bindings/nodejs/src/ailoy_core.d.ts
```diff
@@ -38,14 +39,14 @@ export type FinishReason =
...
 
 export type Grammar =
   | { type: "Plain" }
   | { type: "JSON" }
-  | { type: "JSONSchema"; field0: string }
-  | { type: "Regex"; field0: string }
-  | { type: "CFG"; field0: string };
+  | { type: "JSONSchema"; schema: string }
+  | { type: "Regex"; regex: string }
+  | { type: "CFG"; cfg: string };
```
grf53: It looks like you've specified a lower case transformation, but their `type`s still appear to start with uppercase letters. Are there any potential issues with this?

The consistent format among node and web seems good. :+1:

v2/bindings/nodejs/src/ailoy_core.d.ts
```diff
@@ -19,14 +21,13 @@ export declare class LangModelRunIterator {
...
-  Responses = "Responses",
-  Grok = "Grok",
-}
+export type APISpecification =
+  | "ChatCompletion"
+  | "OpenAI"
+  | "Gemini"
+  | "Claude"
+  | "Responses"
+  | "Grok";
```
grf53, line 30: Unlike others, `APISpecification` doesn't appear to be lowercased. Is this intentional? Is it because it's a type of trademark?

v2/src/model/language_model.rs
```diff
@@ -28,31 +31,37 @@ pub enum ThinkEffort {
...
-    Plain(),
-    JSON(),
-    JSONSchema(String),
-    Regex(String),
-    CFG(String),
+    Plain {},
+    JSON {},
+    JSONSchema { schema: String },
+    Regex { regex: String },
+    CFG { cfg: String },
```
grf53, line 49: Is the impact on Python bindings of these changes using struct variants also checked?

v2/bindings/nodejs/src/ailoy_core.d.ts
khj809: I tried to make it as `snake_case`, then `OpenAI` becomes `open_a_i`.
It's possible to add `#[serde(rename = "openai")]` which works for wasm binding, but for nodejs it's impossible to add `#[cfg_attr(feature = "nodejs", napi_derive::napi(value = "openai"))]` with the following error:
```
expected non-macro attribute, found attribute macro `napi_derive::napi`
```

So there's several options:
- Do not change casing (which is the current state)
- Change `OpenAI` to `Openai`
- Change to `lowercase` (then we endure using `chatcompletion`)

Which option do you think the best?

v2/bindings/nodejs/src/ailoy_core.d.ts
khj809: It's the limitation of napi-rs. It seems `string_enum = "xxx"` does not work for enums with associated data.
If we decide not to convert the cases at all, then wasm and node becomes equal. Is that better?

v2/src/model/language_model.rs
khj809: I didn't modified Python binding at all, in order to prevent the conflict with your task.

v2/bindings/web/src/ailoy-web.d.ts
```diff
@@ -0,0 +1,103 @@
...
+    description: string | undefined;
+    parameters: any;
+    returns: any | undefined;
+}
+
+export type Bytes = Uint8Array;
+
+export type Value = undefined | boolean | number | number | number | string | Record<string, any> | Value[];
+
+export type PartDelta = { text: string } | { id: string | undefined; function: PartDeltaFunction } | { value: Value } | [];
```
grf53: This is not tagged with `type` attribute or sth. It may have been intentional, but I'm letting you know in case you missed it.

v2/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2024"
 
 [package.metadata.wasm-pack.profile.release]
-wasm-opt = ["-O", "--enable-bulk-memory"]
+wasm-opt = ["-O", "--all-features"]
```
grf53: Could you explain what is this for?

khj809: At the first time I worked on this PR, the build was failed on running `wasm-opt`.
I couldn't figure out the exact reason, so I added `--all-features` for the time being.
But I confirmed just now that `wasm-opt` is running successfully with previous option, which is quite strange.. So let me revert this change.

v2/bindings/nodejs/src/ailoy_core.d.ts
grf53: I think it's better to keep it that way, as I think there are some reasons to keep the case format, aside from technical reasons.

v2/src/model/language_model.rs
grf53: What I meant was, we just need to make sure that it is clear enough that there are no critical issues from the changes on common elements.
While it may be nice to choose a method that doesn't cause conflicts, if a change to a common element would entail changes for non-target environments, I think that should be included in this PR. I believe it's better to have a conflicting change than to have it simply not work.

khj809: Okay, then I'll look for Python bindings too.

v2/bindings/nodejs/src/ailoy_core.d.ts
khj809: I managed to convert the cases by applying https://github.com/napi-rs/napi-rs/pull/2960.
But I'm not sure whether this is merged, so I used the commit revision (https://github.com/brekkylab/napi-rs/commit/5c8991b91072b5e468b1781de5703dc8449e000a) for now.

v2/src/model/language_model.rs
khj809: I tried to fix the affected parts, but I'm currently experiencing too many errors and most of them are from anyhow results (maybe from #205). Did you also have these errors on your work too?

grf53: Ah, there were some more parts to fix for anyhow. I am sorry to miss to share that. That was only on my working branch. Let me see and try to apply those on this PR branch.

grf53: Added a commit to fix for python. Please try on that commit.

v2/src/value/message.rs
```diff
@@ -22,9 +24,12 @@ pub enum Role {
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
jhlee525: For here and other changes: Do we have to use camel case? I thought the convention is snake case.

khj809: We need to discuss about the concrete convention, but currently I'm using `snake_case` only for enums (string enums and structured enums with discriminant) and using `camelCase` for structs. (one exception is `APISpecification`)
Especially, the Message has `tool_calls` field, which is converted to `toolCalls` on napi by default and remain as-is on wasm-bindgen by default.

jhlee525: The serialization schema can be the input of local AI or API's input, so we need to follow their convention.

1. The Qwen3 uses `tool_calls` in their chat template, also HF functions may does.
2. Also openai's chat completion API uses `tool_calls` (snake case).

I think `APISpecification` is already camal case, since API is already acronym.

jhlee525: Is the serde logic related to `napi-rs`'s field name binding?

khj809: I see. Then we should use `snake_case`, but also need to specify `js_name` on napi to have snake cased name, since they automatically convert field names to camelCase by default (https://napi.rs/docs/concepts/naming-conventions#snake_case-to-camelcase). 
For example:
```
    #[cfg_attr(feature = "nodejs", napi_derive::napi(js_name = "tool_calls"))]
    pub tool_calls: Vec<Part>,
```

khj809: > Is the serde logic related to `napi-rs`'s field name binding?

napi-rs has nothing to do with serde, so it makes nodejs binding too tedious.

grf53 (approved):
LGTM!

PR #216, opened 2025-10-13
docstring for value module

jhlee525:
I added docstrings for only (considered as a) public APIs.

Please let me know if more docstring to be added for `Value` or something else.

### DIFF
--- a/v2/src/value/tool_desc.rs
+++ b/v2/src/value/tool_desc.rs
@@ -2,6 +2,49 @@ use serde::{Deserialize, Serialize};
 
 use crate::value::Value;
 
+/// Describes a **tool** (or function) that a language model can invoke.
+///
+/// `ToolDesc` defines the schema, behavior, and input/output specification of a callable
+/// external function, allowing an LLM to understand how to use it.
+///
+/// The primary role of this struct is to describe to the LLM what a *tool* does,
+/// how it can be invoked, and what input (`parameters`) and output (`returns`) schemas it expects.
+///
+/// The format follows the same **schema conventions** used by Hugging Face’s
+/// `transformers` library, as well as APIs such as *OpenAI* and *Anthropic*.
+/// The `parameters` and `returns` fields are typically defined using **JSON Schema**.
+///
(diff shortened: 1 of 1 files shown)


v2/src/value/part.rs
```diff
@@ -296,6 +383,26 @@ impl Part {
...
 }
 
+/// Represents an incremental update (delta) of a function part.
+///
+/// This type is used during streaming or partial message generation, when function calls are being streamed as text chunks or partial JSON fragments.
+///
+/// # Variants
+/// * `Verbatim(String)` — Raw text content, typically a partial JSON fragment.
+/// * `WithStringArgs { name, args }` — Function name and its serialized arguments as strings.
+/// * `WithParsedArgs { name, args }` — Function name and parsed arguments as a `Value`.
```
khj809: I have changed some field names not to have serde rename in #214 (e.g. `args` -> `arguments`).
Let me fix them after merging #214.

v2/src/value/part.rs
```diff
@@ -77,16 +146,23 @@ pub enum PartImage {
...
         id: Option<String>,
         #[serde(rename = "function")]
         f: PartFunction,
     },
+
+    /// Holds a structured data value, typically a JSON object.
     #[serde(rename = "value")]
     Value { value: Value },
+
+    /// Contains image data or references with optional metadata.
```
khj809: What is the meaning of "references with optional metadata" in here?

jhlee525: "It is image data(image binary or encoded png / jpeg / etc.) or references(URL) with optional metadata."

It may be an ambiguous explaination. Should we add some parenthesis clauses?

jhlee525: I'll change it to

```
/// Contains an image payload or reference used within a message part.
/// The image may be provided as raw binary data or an encoded format (e.g., PNG, JPEG),
/// or as a reference via a URL. Optional metadata can be included alongside the image.
```

v2/src/value/part.rs
jhlee525: I changed it to latest name

grf53 (approved):

PR #220, opened 2025-10-15
Introducing document & document polyfill

jhlee525:
Many LLM models and libraries provide RAG functionality under the concept of **documents**.

For example, Hugging Face’s [apply_chat_template](https://huggingface.co/docs/transformers/internal/tokenization_utils#transformers.PreTrainedTokenizerBase.apply_chat_template) function accepts a documents argument, and Cohere’s [Command-R chat template](https://huggingface.co/CohereLabs/command-a-reasoning-08-2025/blob/main/chat_template.jinja) also takes documents as input.

Typically, `Document` struct has following schema:
```
{
  "title": ...,
  "text": ...
}
```
Goal in this PR is to modify the `LangModel` so that it can accept this documents parameter. Integration with the actual `Knowledge` component will be handled in a subsequent PR.

Currently, models like Qwen3 (and many others) natively support tools but not documents. Apart from Cohere’s Command models, there are few models that handle documents natively. Therefore, this PR also introduces the concept of a **polyfill**, which enables the use of documents for models that do not natively support them. This is an initial implementation, and many additional configuration options will likely be required later.

Implementing this change will eventually require a large change of the `VectorStore` and `Knowledge` systems.

It is to be addressed in a future PR.


jhlee525:
@CodiumAI-Agent /review

v2/src/model/local/local_language_model.rs
```diff
@@ -156,6 +157,8 @@ impl LocalLangModelImpl {
             let mut mode = "content".to_owned();
             let mut finish_reason = FinishReason::Stop{};
 
+            yield MessageOutput{delta: MessageDelta::new().with_role(Role::Assistant), finish_reason: None};
+
```
khj809, line 161: Is this a necessary yield?

v2/src/model/local/chat_template.rs
```diff
@@ -38,50 +37,38 @@ impl ChatTemplate {
...
+            ThinkEffort::High => "high",
+            _ => "",
+        };
+        let ctx = context!(
+            messages => messages,
+            tools => if !tools.is_empty() { Some(tools) } else { None::<_> },
+            documents => if !documents.is_empty() { Some(documents) } else { None::<_> },
+            add_generation_prompt=>add_generation_prompt,
+            enable_thinking=>enable_thinking,
+            reasoning_effort=>if !reasoning_effort.is_empty() { Some(reasoning_effort) } else { None::<_> },
```
khj809:
```suggestion
            add_generation_prompt => add_generation_prompt,
            enable_thinking => enable_thinking,
            reasoning_effort => if !reasoning_effort.is_empty() { Some(reasoning_effort) } else { None::<_> },
```

v2/src/knowledge/custom_knowledge.rs
```diff
@@ -61,17 +62,20 @@ mod tests {
         let knowledge = Knowledge::new_custom(CustomKnowledge::new(Arc::new(|_, _| {
             async {
                 let documents = vec![
-                    KnowledgeRetrieveResult {
-                        document: "Ailoy is an awesome AI agent framework.".into(),
-                        metadata: None,
+                    Document {
+                        id: String::from("1"),
```
khj809: Just out of curiosity, why used this in favor of `"1".into()`?

v2/src/model/local/local_language_model.rs
jhlee525: Yes. Without this, it results to error in `MessageDelta::finish()` since the role is not provided.

khj809: Ah, I also faced the error in #214 so added `let delta = MessageDelta::new().with_role(Role::Assistant);` in the loop.
Then now we can remove `.with_role(Role::Assistant)` from there.

v2/src/knowledge/custom_knowledge.rs
jhlee525: There’s no particular reason - it’s just a habit I picked up from Rust’s “Hello, world!” example, which uses `String::from`.

I know that there’s a small performance difference among various string creation methods. In some cases, using `to_owned` instead of others can offer a slight performance improvement. If I had to pick the best option, `to_owned` would probably be the most efficient choice in most cases.

https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/2

khj809 (approved):

v2/src/knowledge/mod.rs
```diff
@@ -1,3 +1,84 @@
...
+//!   Internally, it's unified abstraction that wraps either a [`VectorStoreKnowledge`] or a [`CustomKnowledge`] implementation.
+//! - [`KnowledgeBehavior`]: Trait defining how a knowledge source retrieves documents.
+//! - [`KnowledgeTool`]: Exposes a retriever as an LLM-callable tool.
+//! - [`KnowledgeConfig`]: Retrieval configuration (e.g., `top_k` results).
+//!
+//! # Example
+//!
+//! ```rust
+//! use crate::knowledge::{Knowledge, KnowledgeConfig};
+//! use crate::vector_store::SimpleVectorStore;
```
khj809:
```suggestion
//! use crate::vector_store::FaissStore;
```

PR #222, opened 2025-10-15
WASM binding: `LangModel` and `EmbeddingModel`

khj809:
This PR implements WASM binding for `LangModel` and `EmbeddingModel` as stated in #221.
- Added a basic project setup on `bindings/web` with test cases for lang models and embedding model.
- In order to correctly load JS shim's static files, we need to copy the built `dist` directory of JS shim to `binding/web/src/shim_js`. Also, it should be included in the bundled package of `ailoy-web`. 
- There's a name conflict between Rust's `EmbeddingModel` and JS shim's `EmbeddingModel` (same for `LangModel`, but it wasn't detected by chance). So I put prefix `TVM` in JS shim's class names.
- Fixed some build errors on Knowledge and Tool.

### DIFF
--- a/v2/bindings/web/package.json
+++ b/v2/bindings/web/package.json
@@ -17,6 +17,8 @@
         "vitest": "^3.2.4"
     },
     "scripts": {
+        "build:shim": "mkdir -p ./src/shim_js/dist && cp ../../shim_js/dist/* ./src/shim_js/dist/",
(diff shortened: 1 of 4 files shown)


jhlee525 (approved):
👍

grf53 (approved):

PR #223, opened 2025-10-17
LangModel & Tool for Python

grf53:
## Description
Add bindings for various structs and methods for using `LangModel` and `Tool` in Python.

### `LangModel`
`LangModel` and its execution-related `Part`, `PartDelta`, `Role`, `Message`, `MessageDelta`, InferenceConfig, etc. were exposed as Python APIs, and methods for using them were added. `__repr__()` was appropriately implemented to ensure consistent results when outputting via `print()`.

### `Tool`
First, I exposed `Tool` and added the class method `new_py_function()` for using Python functions as tools.
`Tool` is executed by `Agent` by default, but user also can call it so that they can test it directly in Python.
- `__call__()`: The `Tool` instance is itself a callable object. In this case, it returns a Future, so you need to await it.
-`call()/call_sync()`: You can also try executing it through the two methods of the `Tool` instance. `call()` is equivalent to `__call__()`, and `call_sync()` is for synchronous execution.

### Examples
- `v2/bindings/python/tests/test_langmodel.py`
- `v2/bindings/python/tests/test_tools.py`


jhlee525 (approved):
👍

v2/src/model/language_model.rs
```diff
@@ -314,28 +314,64 @@ mod py {
...
+    impl InferenceConfig {
+        #[new]
+        // #[pyo3(signature = (think_effort=None, temperature=None, top_p=None, max_tokens=None, grammar=None))]
+        #[pyo3(signature = (think_effort=None, temperature=None, top_p=None, max_tokens=None))]
+        fn __new__(
+            think_effort: Option<ThinkEffort>,
+            temperature: Option<f64>,
+            top_p: Option<f64>,
+            max_tokens: Option<i32>,
+            // grammar: Option<Grammar>,
```
khj809: Why grammar is commented?

v2/bindings/python/tests/test_tools.py
```diff
@@ -0,0 +1,109 @@
...
+                    "description": "temperature unit",
+                    "enum": ["Celsius", "Fahrenheit"],
+                },
+            },
+            "required": ["location", "unit"],
+        },
+    )
+
+    # with async function
+    # tool = ai.PythonAsyncFunctionTool(func=tool_temperature)
```
khj809: unused?

v2/src/model/language_model.rs
grf53: Grammar is not currently supported, so I thought we don't need to include it in the User API.

v2/bindings/python/tests/test_tools.py
grf53: I often leave comments in my old code to quickly check for changes compared to the previous version, and this seems to be the case. While this will depend on the implementation of Python patches, I suspect the existing `PythonAsyncFunctionTool` will likely be deprecated.

Let me remove that.

v2/src/tool/mod.rs
```diff
@@ -31,6 +30,8 @@ pub enum ToolInner {
 }
 
 #[derive(Debug, Clone)]
+#[cfg_attr(feature = "python", pyo3_stub_gen_derive::gen_stub_pyclass)]
+#[cfg_attr(feature = "python", pyo3::pyclass(subclass))]
```
khj809, line 34: Are you planning to use the class `Tool` as the base class of another classes?
Unless, I think `subclass` attribute is not necessary.

grf53: I added it out of necessity. I already have a subclass from `Tool` class. 
Take a look at this:
https://github.com/brekkylab/ailoy/pull/223/files/7c55fb6770a82ca4c1fcd8f6380ec5dbf4fd0325#diff-acc9f83b749268b15a901d0260302a50a7a59dc19eb4b2d1947a5dfde0fb9e3fR244

khj809: Ah I see. I was forgetting about this.

PR #224, opened 2025-10-19
WASM binding for Agent, etc.

khj809:
This PR adds the remaining WASM binding such as Agent, VectorStore, Knowledge, Tool, MCPClient, etc.

## Tool
- Added `Tool` binding with `newFunction()` constructor which gets async js function. In order to store this function properly, I changed `ToolFunc` signature from `dyn Fn(Value) -> Value` to `dyn Fn(Value) -> Pin<Box<dyn Future<Output = Result<Value, anyhow::Error>>>>`. This may affect another language bindings, so we need to make sure if this change looks appropriate.
- For MCP tools, added `MCPClient` binding with `streamableHttp` constructor and `tools` attribute.
- Added test cases for function tool and MCP tool.

## VectorStore
- Like another variant classes, I changed `VectorStore` from trait to struct which holds `VectorStoreInner` enum for either Faiss or Chroma. The previous trait is renamed as `VectorStoreBehavior`.
- Added `VectorStore` binding with `newFaiss` and `newChroma` constructors and methods.
- By default, the type of `Metadata` is generated as [javascript Map](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Map), which makes it quite cumbersome to handle rather than plain js object. So I used `#[wasm_bindgen(typescript_custom_section)]` to manually generate its type as `Record<string, any>`.
- Added test cases for Faiss and Chroma vectorstore.

## Knowledge
- Added `Document` and `DocumentPolyfill` bindings.
- Added `Knowledge` binding with `newVectorStore` constructor and methods. 
- Note that the constructor's argument types are references, to prevent transferring ownership of vectorstore and embedding model objects.
- Added test case for VectorStoreKnowledge.

## Agent
- Added `Agent` binding with a constructor and methods.
- Similar to Knowledge, the constructor takes the reference of LangModel. However, it's impossible to deliver the vector of references in wasm-bindgen, so the ownership of `tools` will be transferred. I added the comments warning about this.
- In order to reuse `stream_to_async_iterable` for Agent's `run()`, I modified the signature of `run()` to return `BoxStream` similar to LangModel.
- Added test case for Local and OpenAI agents.

## Model
- Renamed some constructors: `create_local` -> `newLocal`, `create_stream_api` -> `newStreamAPI`.


jhlee525 (approved):
👍

khj809:
@grf53 I've fixed the Python binding for VectorStore in the latest commit (de9b4fc). PTAL.

PR #226, opened 2025-10-19
Python Agent

grf53:
## Description
Add bindings for using `Agent` in Python based on bindings for `LangModel` and `Tool`(#223)

### Notable changes
- `create_terminal_tool()`  now returns instance of `Tool` instead of `FunctionTool`.
  There is no interface to create `Tool` with `FunctionTool` instance now. But we can just get `Tool` directly, then use it.
- `Agent::clear_messages()`, `Agent::add_tools()`, `Agent::add_tool()`, `Agent::remove_tools()`, and `Agent::remove_tool()` are no longer async functions and do not return results to indicate errors.
  `Arc` is not used for `Agent`, as it is regarded that its operations do not need to be async nor return errors.
- `Agent::run()` and `Agent::run_sync()` take `Option<InferenceConfig>` and pass it to `LangModel::infer()`(`default()` if `None`). This could be extended to a unified Agent Run Config in the future.
- `LanguageModelRunAsyncIterator`, `LanguageModelRunSyncIterator` are renamed into `LangModelRunAsyncIterator`, `LangModelRunSyncIterator`.
  These are auxiliary types that just represent iterator to get the streamed results from `LangModel`, so it's unlikely that the name will be used directly, and it's not a big deal, but we're changing it for consistency with LangModel.

### DIFF
--- a/v2/src/agent/agent.rs
+++ b/v2/src/agent/agent.rs
@@ -310,7 +310,12 @@ mod tests {
         use futures::StreamExt;
 
         use super::*;
-        use crate::{model::LangModel, to_value, value::ToolDesc};
+        use crate::{
+            model::LangModel,
+            to_value,
+            tool::ToolFunc,
+            value::{ToolDesc, Value},
+        };
 
         let model = LangModel::try_new_local("Qwen/Qwen3-0.6B").await.unwrap();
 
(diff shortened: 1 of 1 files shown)


v2/src/ffi/py/value.rs
```diff
@@ -148,7 +147,7 @@ impl Message {
         }
     }
 
-    fn __repr__(&self) -> String {
+    pub fn __repr__(&self) -> String {
```
khj809, line 151: Is it necessary to put `pub` in `__repr__`?
If so, we need to put it for every `__repr__` for consistency.

grf53: Some `__repr__()` methods are marked as `pub` because they need to be called from another `__repr__()` methods in a different module.

I am also glad to put `pub` for every `__repr__()` methods.
Let me do that.

grf53: Done!

PR #227, opened 2025-10-20
Agent run with messages instead of a single message

### LINKED ISSUE #207: Change Agent::run input - to `Vec<Message>`
Our most important API, `Agent::run`, currently has the following function signature:

```rust
pub fn run<'a>(&'a mut self, contents: Vec<Part>) -> impl Stream<Item = Result<AgentResponse, String>>
```

This API takes new user input as its argument. However, this design introduces several problems:
- Since the Agent internally holds the conversation history, it cannot remain stateless.
- We are forced to keep adding APIs that allow the Agent to query or modify its conversation history.
- Because run is asynchronous, the message history must be made Send and Sync, which requires additional synchronization (e.g., mutexes).

### Proposed Change

We propose changing the API as follows:

```rust
pub fn run<'a>(&'a mut self, messages: Vec<Message>) -> impl Stream<Item = Result<AgentResponse, String>>
```

With this design:
- The user explicitly manages the conversation history and can reassemble messages as needed.
- For each step, the user provides the exact set of messages to be used.

This simplifies the Agent’s responsibilities and results in a cleaner, more flexible API.

grf53:
## Description
This PR fixes `Agent::run()` to get multiple messages instead of the contents of a single message, which resolves #207.

Previously, it was assumed that the contents of the user message were always coming in and that the system message was not given by the user, but now it has been improved to process the incoming messages in an appropriate manner by evaluating whether the system message was given(to override system message) or whether the user message was received last(to compose query for knowledge).

### DIFF
--- a/v2/src/agent/agent.rs
+++ b/v2/src/agent/agent.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use anyhow::Context;
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


v2/src/agent/agent.rs
```diff
@@ -119,49 +113,51 @@ impl Agent {
...
-                    }
-                };
-                Some(retrieved)
+            let system_message_content = if let Some(message) = first_message && message.role == Role::System {
+                messages.remove(0).contents.iter().filter(
+                    |&c| matches!(c, Part::Text{..})
+                ).map(|c| c.as_text().unwrap()).collect::<Vec<_>>().join("\n")
             } else {
-                None
+                "You are a helpful assistant.".to_owned()
```
grf53: The default system message is hardcoded like this, and I'm wondering if this is the proper way to do it.

jhlee525: Can’t we just use `Option`?
Is there a reason it needs to be filled with a default value?

khj809: This part has been removed as of the latest main, so pull the latest main and forget about this.

khj809 (approved):

PR #229, opened 2025-10-21
Add nodejs binding for Agent, etc.

khj809:
This PR implements Nodejs bindings for Agent, VectorStore, Knowledge, Tool and relevant objects.

## Embedding (breaking change)
- Since Nodejs does not support binding `f32` (same for `Vec<f32>`) natively, I updated it as a transparent struct: `pub struct Embedding(pub Vec<f32>);`. Then I fixed all the affected bindings and it's now interpreted natively as `list[float]` in Python, `Float32Array` in Nodejs and WASM.
- To minimize the impact on the core logic, I implemented `From<Vec<f32>>` and `Into<Vec<f32>>` for `Embedding`, so we need to update the existing `embedding` variables to `embedding.into()` (see `faiss.rs` and `chroma.rs` for detail).

## Models
- Added EmbeddingModel binding with constructors and methods.
- In LangModel, I updated the default value for `InferenceConfig` to have default document polyfill. Without specifying the document polyfill config, the models can consume documents as expected. (if this is not an expected behavior, let me revert it @jhlee525)

## VectorStore
- Added VectorStore binding with constructors and methods.
- By the same reason with Embedding, the `distance` field in `VectorStoreRetrieveResult` has been updated to `f64`.
- Updated the test cases according to the latest usages.

## Tool
- Added Tool and MCPClient binding with constructors and methods.
- To prepare for extending builtin tools, I added `BuiltinToolKind` enum, which currently has `Terminal` only. Then I added the constructor for builtin tools as `new_builtin(kind: BuiltinToolKind)`, and also added binding constructors using this. (Note that the existing `terminal` staticmethod in Python has been removed in favor of this update @grf53)
- Now WASM binding also has constructor for builtin tools, though it may result in error for `Terminal`.

## Knowledge
- Added Knowledge binding with constructors and methods.

## Agent
- Added Agent binding with constructors and methods. The iterator for `AgentResponse` is implemented similar to LangModel.

## Bugfixes
- There's two errors in Gemini:
  - When returning tool message (`functionResponse`), it assumes `msg.id` to be exist but it was not collected on unmarshaling `functionCall` part. So I filled `id` with function name.
  - The marshaled `functionCall` body should not include `id` field, but it was trying to include it. So I removed that part.
- <del>The patch I made on napi-rs (https://github.com/napi-rs/napi-rs/pull/2960) was incorrect, so I fixed it and used the updated commit revision.</del> Now the PR is merged, so using the commit revision of official repo until the next version is released.

## Improvements
- In wasm-bindgen, the order of generated type definitions is not deterministic, which results in meaningless changes for each commit. To mitigate this, I added `scripts/revise-typedef.js` which sorts the type definitions in alphabetical order.

### DIFF
--- a/v2/src/value/embedding.rs
+++ b/v2/src/value/embedding.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 use crate::utils::Normalize;
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
-pub struct Embedding(pub Vec<f32>);
+pub struct Embedding(Vec<f32>);
 
 impl From<Vec<f32>> for Embedding {
     fn from(value: Vec<f32>) -> Self {
--- a/v2/src/vector_store/local/faiss.rs
+++ b/v2/src/vector_store/local/faiss.rs
@@ -35,7 +35,7 @@ impl FaissStore {
 #[multi_platform_async_trait]
 impl VectorStoreBehavior for FaissStore {
     async fn add_vector(&mut self, input: VectorStoreAddInput) -> anyhow::Result<String> {
-        let ids: Vec<String> = self.index.add_vectors(&[input.embedding.0]).unwrap();
+        let ids: Vec<String> = self.index.add_vectors(&[input.embedding.into()]).unwrap();
         let id = ids.iter().next().unwrap().clone();
         self.doc_store.insert(
             id.clone(),
@@ -68,7 +68,7 @@ impl VectorStoreBehavior for FaissStore {
             .add_vectors(
                 embeddings
                     .into_iter()
-                    .map(|emb| emb.0)
+                    .map(|emb| emb.into())
                     .collect::<Vec<_>>()
                     .as_slice(),
             )
@@ -125,7 +125,7 @@ impl VectorStoreBehavior for FaissStore {
         query_embedding: Embedding,
         top_k: usize,
     ) -> anyhow::Result<Vec<VectorStoreRetrieveResult>> {
-        let index_results = self.index.search(&[query_embedding.0], top_k).unwrap();
+        let index_results = self.index.search(&[query_embedding.into()], top_k).unwrap();
         let index_result = index_results.into_iter().next().unwrap();
 
         Ok(index_result
@@ -156,7 +156,7 @@ impl VectorStoreBehavior for FaissStore {
             .search(
                 query_embeddings
                     .into_iter()
-                    .map(|query| query.0)
+                    .map(|query| query.into())
                     .collect::<Vec<_>>()
                     .as_slice(),
                 top_k,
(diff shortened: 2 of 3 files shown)


v2/bindings/nodejs/src/index.ts
```diff
@@ -1,66 +1,42 @@
...
+  PartFunction,
+  PartImage,
+  PartImageColorspace,
+  Role,
+  ThinkEffort,
+  ToolDesc,
+  VectorStoreAddInput,
+  VectorStoreGetResult,
+  VectorStoreRetrieveResult,
+} from "./ailoy_core";
```
jhlee525: Do we necessarily have to explicitly export every name?

I think it offers little practical benefit, but if a new class or interface is introduced, it should be added here for completeness. (error-prone)

v2/src/model/polyfill.rs
```diff
@@ -4,17 +4,43 @@ use serde::{Deserialize, Serialize};
 use crate::value::{Document, Message, Part, Role};
 
 /// Provides a polyfill for LLMs that do not natively support the Document feature.
-#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
jhlee525: Current `DocumentPolyfill` might not working outside of `qwen`.

I think better default behavior for polyfill is "empty" just ignoring documents.

(Someone explicitly add polyfill to use this feature)

khj809: Reverted it, and added test case using document polyfill explicitly.

v2/bindings/nodejs/src/index.ts
khj809: It was intentional, I wanted more fine-grained control for targets to be exported. But decided to apply your feedback.

v2/src/model/local/local_embedding_model.rs
```diff
@@ -132,12 +132,12 @@ mod tests {
...
 
-    fn dot(a: &[f32], b: &[f32]) -> f32 {
+    fn dot(a: &Embedding, b: &Embedding) -> f32 {
         if a.len() != b.len() {
             panic!("Cannot dot two vectors of different lengths");
         }
 
-        a.iter().zip(b.iter()).map(|(x, y)| x * y).sum()
+        a.0.iter().zip(b.0.iter()).map(|(x, y)| x * y).sum()
     }
```
grf53: Since `Embedding` is now a value struct, I think we can implement the `std::ops::Mul` trait to make `a * b` replace `dot(a, b)`. `dot()` is not mandatory while using ailoy, but can be very useful in many situation.

For Python bindings, we could define `__matmul__(a, b)` and use `a@b`. For consistency, `__mul__(a, b)`(`a * b`) is also possible, but Python users may be more familiar with the `@` operator.

khj809: Thanks for the good suggestion. Let me add it.

khj809: @grf53 About `__matmul__`, the embedding is converted to `list[float]`, so it's impossible to apply your idea. If we convert it as `np.ndarray`, we can use already implemented `__matmul__` out of the box. But this seems out of scope of this PR, so let's consider it as a future work.

v2/src/value/embedding.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+impl std::ops::Mul for &Embedding {
+    type Output = f32;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        self.clone() * rhs.clone()
+    }
+}
```
grf53: I think it was probably just a matter of implementation order, but it seems inefficient to use `.clone()` to create an object with ownership for the reference(`&Embedding`) implementation.

While there's no significant difference in code and behavior, implementing the actual implementation using a reference and then using the reference implementation for the ownership object seems like it would reduce unnecessary cloning.

like:
```suggestion
impl std::ops::Mul for &Embedding {
    type Output = f32;

    fn mul(self, rhs: Self) -> Self::Output {
        if self.len() != rhs.len() {
            panic!("Cannot dot-product two embeddings of different lengths");
        }

        self.0.iter().zip(rhs.0.iter()).map(|(x, y)| x * y).sum()
    }
}

impl std::ops::Mul for Embedding {
    type Output = f32;

    fn mul(self, rhs: Self) -> Self::Output {
        &self * &rhs
    }
}
```

jhlee525 (approved):

grf53 (approved):

v2/src/vector_store/local/faiss.rs
```diff
@@ -63,7 +63,16 @@ impl VectorStoreBehavior for FaissStore {
                 )
             })
             .unzip();
-        let ids: Vec<String> = self.index.add_vectors(&embeddings).unwrap();
+        let ids: Vec<String> = self
+            .index
+            .add_vectors(
+                embeddings
+                    .into_iter()
+                    .map(|emb| emb.0)
```
grf53: Just a comment.
When converting from Embedding to Vec<f32>, is it recommended to use `.0`? If not, it seems more consistent to use `.into()` for both conversions. In that case, you could also use `Into::into` when applying `map()`, as shown here.

```suggestion
                    .map(Into::into)
```

khj809: Your suggestion makes sense. For the consistency, I removed all `.0` usages and made the inner Vec private.

Issue #234, opened 2025-10-23
Representing simple enum variants by its name in APIs in Python


grf53:
When using simple enums like `ailoy.Role`, it is a little cumbersome to write codes correctly in Python.
`ailoy.Role.User` without parentheses is correct usage while `"user"` simply can represent that in javascript or other LLM API specs.
For simple enums, we can represent them as strings of variant names by implementing appropriate `Into` or `From`, like the case of the color space.

PR #239, opened 2025-10-24
Revise Python binding

khj809:
- Added `PyStringEnum` which converts a simple enum into string literals, as described in #234. Applied to `Role`, `ThinkEffort`, `PartImageColorspace`, `BuiltinToolKind`, and `APISpecification`.
- Renamed LangModel and EmbeddingModel constructors.
- Fixed API model bugs:
  - OpenAI: Responses API requires `"type": "output_text"` for assistant role message, but we were using `input_text` only.
  - Claude: Cannot handle haiku 4.5 model name in current logic.
- Revised pytest codes and verified all cases passed for every LangModel providers.
  - TODO: need to restore multimodal test cases.

### DIFF
--- a/v2/src/ffi/py/value.rs
+++ b/v2/src/ffi/py/value.rs
@@ -156,67 +156,8 @@ impl Message {
         )
     }
 
-    // #[getter]
-    // fn role(&self) -> Role {
-    //     self.role.clone()
-    // }
-
-    // #[setter]
-    // fn set_role(
-    //     &mut self,
-    //     #[gen_stub(override_type(
-    //         type_repr = "Role | typing.Literal[\"system\",\"user\",\"assistant\",\"tool\"]"
-    //     ))]
-    //     role: Bound<'_, PyAny>,
-    // ) -> PyResult<()> {
-    //     Python::attach(|py| {
-    //         if let Ok(role) = role.downcast::<PyString>() {
-    //             self.role = Some(
-    //                 Role::from_str(&role.to_string())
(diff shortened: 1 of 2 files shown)


v2/src/ffi/py/value.rs
```diff
@@ -151,7 +137,7 @@ impl Message {
     pub fn __repr__(&self) -> String {
         format!(
             "Message(role={}, id={:?}, thinking={:?}, contents=[{}], tool_calls=[{}], signature={:?})",
```
grf53:
```suggestion
            "Message(role=\"{}\", id={:?}, thinking={:?}, contents=[{}], tool_calls=[{}], signature={:?})",
```

I just saw `Message` is printed as `Message(role=system, ...)`, but it's better to be `Message(role="system", ...)`.
So suggesting change above. (same for L265 too!)

khj809: For MessageDelta, this may also wrap `None` with double quotes, which is not a desired behavior.
I'll try to convert `Option<Role>` to `Option<String>` and print as `{:?}`.

khj809: I tried to resolve it by introducing `PyRepr` trait and implementing it for every types having `std::fmt::Display`.
See 24bd5db for the details.

grf53: Great to see that.
The `__repr__()` definitions of Rust structs are quite boilerplate, but it seems like we could make a structure with that to improve them.

grf53 (approved):

PR #244, opened 2025-10-24
Enabling to configure temperature and top_p for local models

### LINKED ISSUE #240: Enabling to configure `temperature` and `top_p` for local models

grf53:
## Ticket
resolves #240

## Notes
During the automatic evaluation of some tests, there was an issue where the success or failure of the test was determined probabilistically due to the performance and randomness of LLM. This issue can be resolved by removing randomness by setting `temperature` and `top_p` to 0 through `InferenceConfig`.

### DIFF
--- a/v2/bindings/python/tests/test_agent_async.py
+++ b/v2/bindings/python/tests/test_agent_async.py
@@ -122,9 +122,7 @@ async def test_simple_qna(agent: ai.Agent):
         )
         async for resp in agent.run(
             messages,
-            config=ai.InferenceConfig(
-                temperature=0.0, top_p=0.0, think_effort="disable"
-            ),
+            config=ai.InferenceConfig(temperature=0.0, think_effort="disable"),
         ):
             if resp.finish_reason is not None:
                 result = resp.accumulated


v2/bindings/python/tests/test_agent_async.py
```diff
@@ -72,7 +72,9 @@ async def test_simple_qna(agent: ai.Agent):
         )
         async for resp in agent.run(
             messages,
-            config=ai.InferenceConfig(think_effort=ai.ThinkEffort.Disable),
+            config=ai.InferenceConfig(
+                temperature=0.0, top_p=0.0, think_effort=ai.ThinkEffort.Disable
+            ),
```
khj809: As I mentioned earlier, Claude does not allow providing `temperature` and `top_p` at the same time.
We may need to add an appropriate branching for this.

grf53: I checked that setting only the temperature as zero makes the result text determined,  so I just changed the only `temperature` be zero in config(`top_p` would be the default value).

Issue #241, opened 2025-10-24
Restore Image Part features


khj809:
1. We need to restore handling image url inputs by adding `Url` variant in `PartImage`.
2. For `Binary` image, it's not really practical to ask users fill the image's width, height, and colorspace. Instead we need to provide a helper function to accept binary or base64 data and parse it as `PartImage` (since it's not a class, we cannot create a constructor for that).

PR #246, opened 2025-10-24
Restore image input features

khj809:
This PR restores image input handling logics on binding level, as described in #241.

## Core
- Added `PartImage::Url` variant to handle image url input.
- Added `image_binary_from_bytes()`, `image_binary_from_base64()`, `image_url()` constructors under `Part`.
- Added `base64()` which returns base64 encoded string with PNG format. This reduces the duplicated logics in API models.
- Moved the implementation of `PartImage` -> `image::DynamicImage` conversion as `TryInto`.

## Python
- Added `image_from_bytes()`, `image_from_base64()`, and `image_from_url()` in `Part` class.
- Revised multimodal test cases.

## Nodejs
- Added `imageFromBytes()`, `imageFromBase64()`, `imageFromUrl()` functions.
- Changed `LangModel.newStreamAPI` as async function, to align usages with WASM binding.
- Restored image input tests.

## WASM
- Added `imageFromBytes()`, `imageFromBase64()`, `imageFromUrl()` functions.
- Fixed `DocumentPolyfill` to be camelCased.
- Aligned agent tests with nodejs, except terminal tool and MCP stdio tools.

### DIFF
--- a/v2/src/ffi/py/value.rs
+++ b/v2/src/ffi/py/value.rs
@@ -10,22 +10,6 @@ use crate::{
     },
 };
 
-// #[getter]
-// fn text(&self) -> Option<String> {
-//     match &self {
-//         Part::Text { .. } => Some(self.to_string()),
-//         _ => None,
-//     }
-// }
-
-// #[getter]
-// fn function(&self) -> Option<String> {
-//     match &self {
-//         Part::Function { .. } => Some(self.to_string()),
-//         _ => None,
-//     }
-// }
-
 #[gen_stub_pymethods]
 #[pymethods]
 impl PartDelta {


v2/src/ffi/py/value.rs
```diff
@@ -10,74 +10,21 @@ use crate::{
...
-        };
-        format!("Part.{}", s)
-    }
+// #[getter]
+// fn text(&self) -> Option<String> {
+//     match &self {
+//         Part::Text { .. } => Some(self.to_string()),
+//         _ => None,
+//     }
+// }
```
grf53: This was probably left there by mistake.

v2/src/ffi/py/value.rs
```diff
@@ -10,74 +10,21 @@ use crate::{
     },
 };
 
-#[gen_stub_pymethods]
-#[pymethods]
-impl Part {
```
grf53, line 15: Besides `Part`, would it be more desirable to move other `impl` statements under `v2/src/ffi/py/value.rs` under `mod py` in the file where each struct is defined?

khj809: I thought our policy for ffi has been changed to consolidating binding impls in a single file. Unless, we need to discuss and make it clear about this. @jhlee525

jhlee525: I thought it would be better for the Python binding of value to be placed within the same module(`crate::value::part`) for consistency.
Is there any particular advantage to placing the binding under `ffi::value` instead?

khj809: @grf53 Migrating another impls in `src/ffi/py` to each struct files seems beyond the scope of this PR. Let me do this in the future PR.

grf53: I could help that too!

PR #248, opened 2025-10-27
Add shortcut to get the text of the first contents to `Message` and `MessageDelta`

### LINKED ISSUE #233: Providing alias of `.contents[0].text` of Message object
A expression like `msg.contents[0].text` for a Message `msg` is repeatedly used, but it's a bit complicated and verbose.
We are good to have a short alias for this expression.

grf53:
## Ticket
resolves #233

## Description
This PR adds
- method `Message::to_text()`
- method `MessageDelta::to_text()`
- property `Message.text`
- property `MessageDelta.text`

I also considered nodejs, but since `Message` and `MessageDelta` are defined as interfaces, I couldn't find a solution beyond a simple js helper function that gets the value at a specific location in a js object. In that case, I questioned whether it was even necessary to provide it at the library level, so I didn't add it.


v2/src/value/message.rs
```diff
@@ -120,6 +120,21 @@ impl Message {
...
+                Part::Text { text } => Ok(text.clone()),
+                Part::Value { value } => Ok(serde_json::to_string(&value)
+                    .map_err(|e| anyhow!("Failed to convert part value to text: {}", e))?),
+                _ => {
+                    bail!("The part is not convertible to text.")
+                }
+            },
+            _ => bail!("Cannot convert multiple parts into a text."),
+        }
+    }
```
khj809, line 137: How about just implementing `std::fmt::Display` for `Message`?

grf53: This function is intended to make it easy to retrieve the text value only when a Message has single text contents(raises an error otherwise). I think its purpose is different from the `std::fmt::Display` trait, which is intended for representing general Messages.

Contrary to my initial thoughts, Rust's strictness led me to include quite an amount of validations, and it made the function seems to have more meaning than it actually has.
Since the purpose is to improve the convenience of using the user API, I think it would be sufficient to simply provide it through just a method in the python patch.
How do you think?

khj809: Do you mean moving the `to_text` logics to Python patch level? Then we should make the equivalent logics in another bindings, and this seems not good in terms of maintainability.

grf53: It is correct that you meant to move to the python patch level, but rather than moving `to_text()` completely, I meant I would move it to a naive implementation that provides an simple alias like the code below.

```python
class ...:
    @property
    def text(self) -> str:
        return self.contents[0].text
```

This method doesn't care if contents consists of multiple parts, and raises Python's `IndexError` or `AttributeError` when `contents` is an empty list or `contents[0]` doesn't contain a `text` attribute.

Users will still need to handle the errors generated in Rust(currently `PyValueError` without distinguishment), but rather, this error may be less intuitive than the Python level errors.

I feel like I've taken an overly complex approach for a simple convenience feature, so if I were to do it again, I'd probably approach it at a simpler Python level.

Since I chose the Rust-based approach first, I don't have a preference for either approach and I'm open to suggestions.

khj809: I understood. Raising errors more language friendly makes sense, then I'm okay with applying your suggestion.
But this would be applied in Python only unless we change `Message` and `MessageDelta` as class in JS.

grf53: I think if we keep `Message` and `MessageDelta` as interfaces (which I think is worth to keep), it will still be difficult to utilize the `to_text()` method, and it would be simpler to provide a separate JS function like `message_to_text()`. This would be better from a consistency standpoint, as it would provide a simple binding-level convenience for other languages ​​as well.

grf53: Oh I just realized that the binding level patches are also not that easy. I need to re-define all the functions that returns the overwritten classes _recursively_...

grf53: It was totally bad idea to redefine the patch classes for some tiny common features.

This is useful only when instances of a patched class are created only through user code, or when the patched feature is used only in some limited ways.

The `Tool.new_py_function()` already being used is a perfect fit for this case, meanwhile `MessageDelta.text` is the complete opposite.

grf53: Currently, I became quite skeptical about adding this feature.
The most general and clean way to do this seems to be to add a utility function like `message_to_text()` for each bindings environment. This seems more like a feature that users can define and use themselves when needed, rather than something that should be provided by a library.

The current `to_text()` and `text` property could be reflected as is, but considering the heterogeneity with other environments, this doesn't seem like a very valuable feature.

**I think it might be better to close this PR and related issues.**

@jhlee525 @khj809 I beg your opinions.

khj809: Okay, then I think the changes in this PR should be excluded in #249, and rebase it to main.

jhlee525: It seems the decision on whether to proceed with this PR is up to you.
If you prefer to explore a better solution for this issue later, you’re welcome to do so.

PR #249, opened 2025-10-27
Easier way to represent 'single text contents' `Message`

### LINKED ISSUE #235: Representing 'single text part `contents`' as a string
Message content is defined as a list of parts because it can have multiple types, but this often makes it too verbose to express even a simple case of just one text.

Messages containing only single text content are repeatedly used, so a simpler way to write them can be useful. We can use a single string instead of a list, as the case of many LLM APIs or transformers.

grf53:
## Ticket
resolves #235 

## Description
This PR provides a way to input the `contents` part as a single string instead of multiple `Part`s when using `Message` in python and node bindings, making it simpler to represent `Message`s with 'single text contents', which is the case in most cases.


v2/bindings/nodejs/src/_patches.ts
```diff
@@ -0,0 +1,51 @@
...
+        ? message.contents
+        : [{ type: "text", text: message.contents } as Part],
+    }));
+    return super.run(new_messages, config);
+  }
+}
+
+export class LangModel extends _LangModel {
+  run(
+    messages: Array<SimpleMessage | Message>,
```
grf53: I am not sure this approach is appropriate, needs comment.

grf53: And also, `LangModel` has private constructor in the web environment, so it cannot be extended directly.

khj809: I think we can do this in Rust level with custom typing (also for Python). Let me try this.

khj809:
@grf53 As you requested, I tried to apply the same solution for `contents` in `Message`, instead of subclassing on patch.
But this is available only on Python since it's class and we can override the constructor, and impossible on JS without manipulating the original `Message` itself.

PR #253, opened 2025-10-28
introducing agentconfig

### LINKED ISSUE #217: Configurable parameter for `Agent`
Currently, the `LanguageModel` allows runtime inference configuration using the following structure:

```rust
pub struct InferenceConfig {
    pub think_effort: ThinkEffort,
    pub temperature: Option<f64>,
    pub top_p: Option<f64>,
    pub max_tokens: Option<i32>,
    pub grammar: Grammar,
}
```

`Knowledge` currently uses only one runtime parameter `top_k`.
However, if additional runtime parameters are introduced in the future, this can be extended into a more general configuration struct such as:

```rust
pub struct KnowledgeConfig {
    pub top_k: Option<usize>,
    ...
}
```

Each of these configuration values is passed as a parameter to trait methods such as `infer` or `retrieve`.
For example, the `LangModelInference` trait defines its infer method as follows:

```rust
pub trait LangModelInference {
    /// Runs the language model with the given tools and messages, returning a stream of `MessageOutput`s.
    fn infer<'a>(
        &'a mut self,
        msgs: Vec<Message>,
        tools: Vec<ToolDesc>,
        config: InferenceConfig,
    ) -> BoxStream<'a, anyhow::Result<MessageOutput>>;
}
```

When each module (e.g., `LanguageModel`, `Knowledge`) defines and uses its own configuration struct, there is no issue with local functionality.
However, at the `Agent` level, there is currently no way to control all these runtime configurations dynamically during execution.

To solve this, we can introduce a unified configuration structure named `AgentConfig`, which consolidates the configurations from each module.
The agent’s run function would then receive this as an argument, enabling centralized control over runtime behavior.

For example:
```rust
pub struct Agent
....
fn run(&mut self, ..., config: AgentConfig) {
...
}
```

jhlee525:
Resolves #217 

Also some bug in js unit tests are also fixed


v2/src/agent/agent.rs
```diff
@@ -188,18 +204,27 @@ impl Agent {
...
+        config: Option<AgentConfig>,
     ) -> BoxStream<'a, anyhow::Result<MessageDeltaOutput>> {
         let knowledge = self.knowledge.clone();
         let tools = self.tools.clone();
         let strm = async_stream::try_stream! {
-            let docs = Self::get_docs(&messages, &knowledge).await?;
+            let docs = Self::get_docs(
+                &messages,
+                &knowledge,
+                config.clone().unwrap_or_default().knowledge.unwrap_or_default()
```
khj809: How about creating a temporary variable `let config = config.unwrap_or_default();` and reuse it in here and below?

v2/src/agent/agent.rs
```diff
@@ -231,16 +256,25 @@ impl Agent {
...
+        config: Option<AgentConfig>,
     ) -> BoxStream<'a, anyhow::Result<MessageOutput>> {
         let knowledge = self.knowledge.clone();
         let tools = self.tools.clone();
         let strm = async_stream::try_stream! {
-            let docs = Self::get_docs(&messages, &knowledge).await?;
+            let docs = Self::get_docs(
+                &messages,
+                &knowledge,
+                config.clone().unwrap_or_default().knowledge.unwrap_or_default()
```
khj809: ditto

v2/bindings/nodejs/src/ailoy_core.d.ts
```diff
@@ -1,5 +1,34 @@
...
+ * Deltas can be accumulated into a full message using the provided accumulation utilities.
+ * This allows real-time streaming while preserving the ability to reconstruct the final structured result.
+ *
+ * See `MessageDelta`.
+ *
+ * # Components
+ * - **Language Model**: Generates natural language and structured outputs.
+ *   It interprets the conversation context and predicts the assistant’s next action.
+ * - **Tool**: Represents external functions or APIs that the model can dynamically invoke.
+ *   The `Agent`` detects tool calls and automatically executes them during the reasoning loop.
```
khj809: It seems the backtick is accidentally doubled after Agent. Since this is originated from docstring in Rust, need to fix there and regenerate typedefs.

v2/src/agent/agent.rs
jhlee525: 👍

v2/src/agent/agent.rs
jhlee525: 👍

v2/bindings/nodejs/src/ailoy_core.d.ts
jhlee525: applied

PR #250, opened 2025-10-27
Split agent run

jhlee525:
Split the LangModel and Agent interfaces as follows:
- The existing streaming functions are renamed to `infer_delta` and `run_delta`.
- New non-streaming counterparts infer and run are introduced(`infer`, `run`), which return complete results in once (`Future`).
- The `AgentResponse` struct is removed.
- Common iterator bindings for Python and Node are moved into `value::Message` (`MessageDeltaOutputIterator`, `MessageOutputIterator`, `MessageDeltaOutputSyncIterator`, `MessageOutputSyncIterator`), since both `LangModel` and `Agent` share them.

Component | Function | Return Type | Description
-- | -- | -- | --
`LangModel` | `infer_delta` | `Stream<MessageDeltaOutput>` | Streams delta outputs (token-by-token or chunked).
`LangModel` | `infer` | `Future<MessageOutput>` | Returns a single completed message result.
`Agent` | `run_delta` | `Stream<MessageDeltaOutput>` | Streams intermediate deltas during reasoning and tool calls.
`Agent` | `run` | `Stream<MessageOutput>` | Streams completed message-level results.

Please let me know if there are any opinions for naming or behavior

PR #256, opened 2025-10-28
Enable parallel tool call for local lang models

grf53:
## Description
Due to differences in tool call handling introduced by LocalLangModel, which was written in Rust, parallel tool calls were not properly supported.
This PR addresses this issue, allowing multiple tool calls to be executed within a single `infer()` call.

## Changes
- `LocalLangModel::infer_delta()` will yield `MessageDeltaOutput` along with `FinishReason::ToolCall{}` when a Tool Call finished to be generated(the 'end of tool call' token is generated).
  : Otherwise, multiple tool call objects generated in a single `infer_delta()` will be concatenated as a single string, which will fail to parse.
- ~~`LangModel::infer()` now returns `Stream<MessageOutput>` along with other functions (refer to #250)~~ `LangModel::infer()` now can return multiple tool_calls(parallel tool call)
  : improved internal logics to distinguish multiple verbatim tool call texts as multiple tool calls
- Minor changes
  - Improve python bindings(`__repr__()`, getters)
  - Improve python tests

### DIFF
--- a/v2/bindings/python/tests/test_agent_async.py
+++ b/v2/bindings/python/tests/test_agent_async.py
@@ -165,18 +165,18 @@ async def test_builtin_tool(agent: ai.Agent):
             for content in resp.delta.contents:
                 if isinstance(content, ai.PartDelta.Text):
                     print(content.text, end="")
-                elif content.part_type == "value":
+                elif isinstance(content, ai.PartDelta.Value):
                     print(content.value)
                 else:
                     raise ValueError(
                         f"Content has invalid part_type: {content.part_type}"
                     )
             for tool_call in resp.delta.tool_calls:
-                if tool_call.part_type == "function":
+                if isinstance(tool_call, ai.PartDelta.Function):
                     print(tool_call.function.text, end="")
                 else:
                     raise ValueError(
-                        f"Tool call has invalid part_type: {content.part_type}"
+                        f"Tool call has invalid part_type: {tool_call.part_type}"
                     )
     print()
 
@@ -224,18 +224,18 @@ async def test_python_async_function_tool(agent: ai.Agent):
             for content in resp.delta.contents:
                 if isinstance(content, ai.PartDelta.Text):
                     print(content.text, end="")
-                elif content.part_type == "value":
+                elif isinstance(content, ai.PartDelta.Value):
                     print(content.value)
                 else:
                     raise ValueError(
                         f"Content has invalid part_type: {content.part_type}"
                     )
             for tool_call in resp.delta.tool_calls:
-                if tool_call.part_type == "function":
+                if isinstance(tool_call, ai.PartDelta.Function):
                     print(tool_call.function.text, end="")
                 else:
                     raise ValueError(
-                        f"Tool call has invalid part_type: {content.part_type}"
+                        f"Tool call has invalid part_type: {tool_call.part_type}"
                     )
     print()
 
@@ -287,20 +287,20 @@ async def test_parallel_tool_call(agent: ai.Agent):
         config=ai.InferenceConfig(think_effort="disable"),
     ):
         for content in resp.message.contents:
-            if content.part_type == "text":
+            if isinstance(content, ai.PartDelta.Text):
                 print(f"{content.text=}")
-            elif content.part_type == "value":
+            elif isinstance(content, ai.PartDelta.Value):
                 print(f"{content.value=}")
             else:
                 raise ValueError(f"Content has invalid part_type: {content.part_type}")
         for tool_call in resp.message.tool_calls:
-            if tool_call.part_type == "function":
+            if isinstance(tool_call, ai.PartDelta.Function):
                 print(
                     f"function_call={tool_call.function.name}(**{tool_call.function.arguments})"
                 )
             else:
                 raise ValueError(
-                    f"Tool call has invalid part_type: {content.part_type}"
+                    f"Tool call has invalid part_type: {tool_call.part_type}"
                 )
 
     agent.remove_tools([t.get_description().name for t in tools])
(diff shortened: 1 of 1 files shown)


v2/bindings/python/tests/test_agent_async.py
```diff
@@ -164,6 +165,19 @@ async def test_builtin_tool(agent: ai.Agent):
...
+                    raise ValueError(
+                        f"Content has invalid part_type: {content.part_type}"
+                    )
+            for tool_call in resp.delta.tool_calls:
+                if tool_call.part_type == "function":
+                    print(tool_call.function.text, end="")
+                else:
+                    raise ValueError(
+                        f"Tool call has invalid part_type: {content.part_type}"
+                    )
```
khj809:
```suggestion
                elif isinstance(content, ai.PartDelta.Value):
                    print(content.value)
                else:
                    raise ValueError(
                        f"Content has invalid part_type: {content.part_type}"
                    )
            for tool_call in resp.delta.tool_calls:
                if isinstance(tool_call, ai.PartDelta.Function):
                    print(tool_call.function.text, end="")
                else:
                    raise ValueError(
                        f"Tool call has invalid part_type: {tool_call.part_type}"
                    )
```

v2/src/agent/agent.rs
```diff
@@ -238,21 +238,25 @@ impl Agent {
...
+                    yield assistant_out;
+                    messages.push(assistant_msg.clone());
+
+                    if let Some(tool_calls) = assistant_msg.tool_calls && !tool_calls.is_empty() {
+                        for delta in Self::handle_tool_calls(&tools, tool_calls).await? {
+                            let message_output = MessageOutput { message: delta.finish()?, finish_reason: FinishReason::Stop{} };
+                            yield message_output.clone();
+                            messages.push(message_output.message);
+                            need_infer = true;
+                        }
```
khj809: It seems that `handle_tool_calls` iterates the provided `tool_calls`, so if it's empty the inner for loop is not iterated ever. Hence, I think there's no need to check for `!tool_calls.is_empty()`.
```suggestion
                    if let Some(tool_calls) = assistant_msg.tool_calls {
                        for delta in Self::handle_tool_calls(&tools, tool_calls).await? {
                            let message_output = MessageOutput { message: delta.finish()?, finish_reason: FinishReason::Stop{} };
                            yield message_output.clone();
                            messages.push(message_output.message);
                            need_infer = true;
                        }
```

v2/bindings/python/tests/test_agent_async.py
```diff
@@ -11,7 +11,7 @@
 
 @pytest.fixture(scope="module")
 def qwen3():
-    return ai.LangModel.new_local_sync("Qwen/Qwen3-0.6B", progress_callback=print)
+    return ai.LangModel.new_local_sync("Qwen/Qwen3-4B", progress_callback=print)
```
khj809, line 14: Should we use 4B model for testing?

v2/bindings/python/tests/test_agent_async.py
```diff
@@ -210,6 +224,19 @@ async def tool_temperature(
...
+                    raise ValueError(
+                        f"Content has invalid part_type: {content.part_type}"
+                    )
+            for tool_call in resp.delta.tool_calls:
+                if tool_call.part_type == "function":
+                    print(tool_call.function.text, end="")
+                else:
+                    raise ValueError(
+                        f"Tool call has invalid part_type: {content.part_type}"
+                    )
```
khj809:
```suggestion
                elif isinstance(content, ai.PartDelta.Value):
                    print(content.value)
                else:
                    raise ValueError(
                        f"Content has invalid part_type: {content.part_type}"
                    )
            for tool_call in resp.delta.tool_calls:
                if isinstance(tool_call, ai.PartDelta.Function):
                    print(tool_call.function.text, end="")
                else:
                    raise ValueError(
                        f"Tool call has invalid part_type: {tool_call.part_type}"
                    )
```

v2/bindings/python/tests/test_agent_async.py
```diff
@@ -243,6 +329,19 @@ async def test_mcp_tools(agent: ai.Agent):
...
```
khj809:
```suggestion
                elif isinstance(content, ai.PartDelta.Value):
                    print(content.value)
                else:
                    raise ValueError(
                        f"Content has invalid part_type: {content.part_type}"
                    )
            for tool_call in resp.delta.tool_calls:
                if isinstance(tool_call, ai.PartDelta.Function):
                    print(tool_call.function.text, end="")
                else:
                    raise ValueError(
                        f"Tool call has invalid part_type: {tool_call.part_type}"
                    )
```

v2/bindings/python/tests/test_agent_async.py
```diff
@@ -220,14 +247,73 @@ async def tool_temperature(
...
+                raise ValueError(f"Content has invalid part_type: {content.part_type}")
+        for tool_call in resp.message.tool_calls:
+            if tool_call.part_type == "function":
+                print(
+                    f"function_call={tool_call.function.name}(**{tool_call.function.arguments})"
+                )
+            else:
+                raise ValueError(
+                    f"Tool call has invalid part_type: {content.part_type}"
+                )
```
khj809, line 304:
```suggestion
        for content in resp.message.contents:
            if isinstance(content.part_type, ai.Part.Text):
                print(f"{content.text=}")
            elif isinstance(content.part_type, ai.Part.Value):
                print(f"{content.value=}")
            else:
                raise ValueError(f"Content has invalid part_type: {content.part_type}")
        for tool_call in resp.message.tool_calls:
            if isinstance(tool_call.part_type, ai.Part.Function):
                print(
                    f"function_call={tool_call.function.name}(**{tool_call.function.arguments})"
                )
            else:
                raise ValueError(
                    f"Tool call has invalid part_type: {tool_call.part_type}"
                )
```

v2/bindings/python/tests/test_agent_async.py
grf53: Ah, this is my suggestion that I missed to mention.
When using 0.6B, we observed that some other tests also suffer performance issues: often resulting in abnormal behavior. This was somewhat controllable through temperature.
However, when using 0.6B or 1.7B, parallel tool call tests almost always used only one tool, with the rest responding with appropriate values ​​or an unknown response. Therefore, it's difficult to test for normal operation when a parallel tool call is issued, and I suggest using the 4B model.

khj809: Ok I see. Then I think we should somehow notify at Tools documentation that parallel tool calls might not work in light models.

jhlee525 (approved):

PR #260, opened 2025-10-30
Fix bug on calling async python function tools

grf53:
## Description
Fixes a bug that causes a freeze when an exception is raised within the for loop that processes this message while running an async python function tool on an agent with multiple tools registered.

### DIFF
--- a/src/tool/mod.rs
+++ b/src/tool/mod.rs
@@ -217,15 +217,9 @@ mod py {
             });
 
             // create Rust Future to run tool
-            let inner = self.inner.clone();
+            let tool = self.clone();
             let future = async move {
-                let result = match inner {
-                    ToolInner::Function(tool) => tool.run(input).await,
-                    ToolInner::MCP(tool) => tool.run(input).await,
-                    ToolInner::Knowledge(tool) => tool.run(input).await,
-                }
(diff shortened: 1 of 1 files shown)


jhlee525 (approved):

khj809 (approved):

src/agent/agent.rs
```diff
@@ -348,15 +356,18 @@ mod py {
...
                 if tx.send(item).is_err() {
                     break; // Exit if consumer vanished
                 }
             }
-        });
+        };
+        match pyo3_async_runtimes::tokio::get_current_locals(py) {
+            Ok(locals) => rt.spawn(pyo3_async_runtimes::tokio::scope(locals, future)),
+            Err(_) => rt.spawn(future),
+        };
```
khj809, line 370: Should this also applied to LangModel's spawn functions?

grf53: This change is intended to pass Python event loop information inside async move to prevent deadlock situations with Python function execution, so I don't think it's necessary to reflect it in LangModel, which is not used concurrently with Python execution due to its role.
It might be necessary if such a situation exists, but I don't think it is.

khj809 (approved):

PR #262, opened 2025-10-31
add device selection functionality

jhlee525:
During the process, I realized that the cache structure needed some refactoring.
As a result, the TryFromCache trait has undergone a slight modification.
Previously, the context field inside CacheContents was of type Any,
but now it has been changed to a mutable reference to the cache context.
The new cache signature is as follows:

```
pub trait TryFromCache: Sized + MaybeSend {
    fn claim_files<'a>(
        cache: Cache,
        key: impl AsRef<str>,
        ctx: &'a mut std::collections::HashMap<String, crate::value::Value>,
    ) -> BoxFuture<'a, anyhow::Result<CacheClaim>>;

    fn try_from_contents<'a>(
        contents: CacheContents,
        ctx: &'a std::collections::HashMap<String, crate::value::Value>,
    ) -> BoxFuture<'a, anyhow::Result<Self>>;
}

```

### DIFF
--- a/src/model/language_model.rs
+++ b/src/model/language_model.rs
@@ -314,7 +314,7 @@ mod py {
     impl LangModel {
         #[classmethod]
         #[gen_stub(override_return_type(type_repr = "typing.Awaitable[LangModel]"))]
-        #[pyo3(name = "new_local", signature = (model_name, device_id=None, progress_callback = None))]
+        #[pyo3(name = "new_local", signature = (model_name, device_id = None, progress_callback = None))]
         fn new_local_py<'a>(
             _cls: &Bound<'a, PyType>,
             py: Python<'a>,


src/model/language_model.rs
```diff
@@ -309,17 +314,23 @@ mod py {
     impl LangModel {
         #[classmethod]
         #[gen_stub(override_return_type(type_repr = "typing.Awaitable[LangModel]"))]
-        #[pyo3(name = "new_local", signature = (model_name, progress_callback = None))]
+        #[pyo3(name = "new_local", signature = (model_name, device_id=None, progress_callback = None))]
```
khj809:
```suggestion
        #[pyo3(name = "new_local", signature = (model_name, device_id = None, progress_callback = None))]
```

src/model/language_model.rs
```diff
@@ -520,13 +538,16 @@ mod node {
         #[napi(js_name = "newLocal")]
         pub async fn new_local_js(
             model_name: String,
+            device_id: Option<i32>,
```
khj809, line 541: It's not urgent, but for js bindings it would be better to merge `device_id` and `progress_callback` into a single config object.
Since the order of arguments are fixed (unlike Python), even though users want to give `progress_callback` only, they should insert `null` for `device_id`, which makes the api looks quite inflexible.

khj809 (approved):

grf53 (approved):

PR #265, opened 2025-10-31
doc update

jhlee525:


docs/docs/resources/available-models.mdx
```diff
@@ -0,0 +1,75 @@
...
+        height="20"
+      />
+      Qwen3
+    - `Qwen/Qwen3-0.6B`
+    - `Qwen/Qwen3-1.7B`
+    - `Qwen/Qwen3-4B`
+    - `Qwen/Qwen3-8B`
+    - `Qwen/Qwen3-14B`
+    - `Qwen/Qwen3-32B`
+    - `Qwen/Qwen3-30B-A3B` (MoE)
```
grf53:
```suggestion
      - `Qwen/Qwen3-0.6B`
      - `Qwen/Qwen3-1.7B`
      - `Qwen/Qwen3-4B`
      - `Qwen/Qwen3-8B`
      - `Qwen/Qwen3-14B`
      - `Qwen/Qwen3-32B`
      - `Qwen/Qwen3-30B-A3B` (MoE)
```

docs/docs/resources/command-line-interfaces.mdx
grf53, line 1: I think features that are not yet supported should be marked as such.

jhlee525: Ok We'll update it. If it is decided to exclude, then we should mark it as not implemented

PR #268, opened 2025-11-03
MessageDelta methods for JS environments

grf53:
## Description
Add helper functions that call `MessageDelta::accumulate()` and `MessageDelta::finish()`, which are no longer available when `MessageDelta` is bound as an interface in Node.js and web environments.


jhlee525 (approved):
👍 Can we do same thing on python? (add `accumulate` and `finish`)

khj809 (approved):

PR #270, opened 2025-11-03
Local model finish reason

jhlee525:
1. Set `finish_reason` to `ToolCall` when a tool call is triggered
2. Adjust function parameter signature in `new_py_function` (same as .pyi)

### DIFF
--- a/bindings/python/tests/test_agent_async.py
+++ b/bindings/python/tests/test_agent_async.py
@@ -255,63 +255,63 @@ async def test_python_async_function_tool(agent: ai.Agent):
     agent.remove_tool(tool.get_description().name)
 
 
-@pytest.mark.parametrize(
-    "agent", ["qwen3", "openai", "gemini", "claude", "grok"], indirect=True
-)
-async def test_parallel_tool_call(agent: ai.Agent):
-    async def tool_temperature(
-        location: str, unit: Literal["Celsius", "Fahrenheit"] = "Celsius"
-    ):
-        """
-        Get temperature of the provided location
-        Args:
-            location: The city name
-            unit: The unit of temperature
-        Returns:
-            int: The temperature
-        """
-        await asyncio.sleep(1.0)
-        return 35 if unit == "Celsius" else 95
+# @pytest.mark.parametrize(
+#     "agent", ["qwen3", "openai", "gemini", "claude", "grok"], indirect=True
+# )
+# async def test_parallel_tool_call(agent: ai.Agent):
+#     async def tool_temperature(
+#         location: str, unit: Literal["Celsius", "Fahrenheit"] = "Celsius"
+#     ):
+#         """
+#         Get temperature of the provided location
+#         Args:
+#             location: The city name
+#             unit: The unit of temperature
+#         Returns:
+#             int: The temperature
+#         """
+#         await asyncio.sleep(1.0)
+#         return 35 if unit == "Celsius" else 95
 
-    async def tool_wind_speed(location: str):
-        """
-        Get the current wind speed in km/h at a given location
-        Args:
-            location: The city name
-        Returns:
-            float: The current wind speed at the given location in km/h, as a float.
-        """
-        await asyncio.sleep(1.0)
-        return 23.5
+#     async def tool_wind_speed(location: str):
+#         """
+#         Get the current wind speed in km/h at a given location
+#         Args:
+#             location: The city name
+#         Returns:
+#             float: The current wind speed at the given location in km/h, as a float.
+#         """
+#         await asyncio.sleep(1.0)
+#         return 23.5
 
-    tools = [
-        ai.Tool.new_py_function(tool_temperature),
-        ai.Tool.new_py_function(tool_wind_speed),
-    ]
-    agent.add_tools(tools)
+#     tools = [
+#         ai.Tool.new_py_function(tool_temperature),
+#         ai.Tool.new_py_function(tool_wind_speed),
+#     ]
+#     agent.add_tools(tools)
 
-    async for resp in agent.run(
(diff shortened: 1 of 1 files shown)


src/model/local/local_language_model.rs
```diff
@@ -203,14 +204,14 @@ impl LocalLangModelImpl {
...
-                    yield MessageDeltaOutput{delta, finish_reason: Some(FinishReason::Stop{})};
+                    yield MessageDeltaOutput{delta, finish_reason: Some(finish_reason)};
                     break;
                 } else if s == "<tool_call>" {
                     mode = "tool_call".to_owned();
                     continue;
                 } else if s == "</tool_call>" {
+                    finish_reason = FinishReason::ToolCall{};
                     mode = "content".to_owned();
-                    yield MessageDeltaOutput{delta, finish_reason: Some(FinishReason::ToolCall{})};
```
grf53, line 213: We need to yield here or the parallel tool calls are not processed properly in `Agent::run_delta()`.

It would be good to keep the `finish_reason` to be yielded at the end of the generation , cuz there would not be cases that a text comes after tool calls.

bindings/python/ailoy/_patches.py
```diff
@@ -240,7 +240,9 @@ def get_json_schema(func: Callable) -> dict:
 
 
 def new_py_function(
-    cls: Tool, func: Callable, tool_desc: Optional[ToolDesc] = None
+    cls: Tool,
+    tool_desc: Optional[ToolDesc],
+    func: Callable,
```
grf53: `tool_desc` intentionally has the default value `None` here, that marks that the `tool_desc` would be generated from the function docstring.

If the users just want to use the `tool_desc` they defined manually, the parameter would be not `None` and used instead of generated one.

Is there another reason that I didn't catch for this?

jhlee525: I agree with that.

However, since in the .pyi file the order appears as `tool_desc`, `func`, users see that order in autocomplete, even though the actual order is `func`, `tool_desc`.

If we can change order in pyi, it'll be fine

khj809: There's a fake function for the monkeypatched classmethod. We can change the order of parameters in there.

src/model/local/local_language_model.rs
jhlee525: I’m not entirely sure what the best strategy is for handling parallel tool calls in `LangModel` or `Agent`.

Here’s what I’m currently thinking:
- If a `<tool_call>` tag immediately after a `</tool_call>` tag: treat it as a parallel tool call and continue processing the next call.
- any other token appears after a `</tool_call>` tag: stop processing and just return the tool call result.

Does that make sense?

bindings/python/ailoy/_patches.py
jhlee525: Ok then let me change the hollow function in rust, instead of API.

src/tool/mod.rs
```diff
@@ -164,8 +164,8 @@ mod py {
 
         #[allow(unused_variables)]
         #[classmethod]
-        #[pyo3(name = "new_py_function", signature = (desc, func))]
-        pub fn new_py_function(_cls: &Bound<'_, PyType>, desc: ToolDesc, func: Py<PyAny>) -> Self {
+        #[pyo3(name = "new_py_function", signature = (func, desc))]
+        pub fn new_py_function(_cls: &Bound<'_, PyType>, func: Py<PyAny>, desc: ToolDesc) -> Self {
```
khj809: `desc` should be `Option<ToolDesc>` to be aligned with monkeypatched method's `Optional[ToolDesc]`.

jhlee525: Thank you!

khj809 (approved):
I'm okay if the issue from @grf53 is addressed.

src/model/local/local_language_model.rs
grf53: The `<tool_call>` or `</tool_call>` token can only be checked within `LocalLangModel::infer_delta()`.
In `Agent::run_delta()` or `LocalLangModel::infer()`, which uses `LocalLangModel::infer_delta()`, it checks for the presence of `finish_reason` before processing the tool call (either executing the tool or converting to `PartDeltaFunction::WithParsedArgs`). If not, it continues accumulating the delta.

If `LocalLangModel::infer_delta()` encounters `</tool_call>` and does not yield a delta containing `finish_reason`, the verbatim tool call text will continue to accumulate, resulting in values ​​such as `\n{...}\n\n{...}`. Calling `finish()` on such a delta will cause problems parsing values ​​that are not valid JSON strings.

If you provide a delta with a clear `finish_reason` when a `</tool_call>` token comes out in `LocalLangModel::infer_delta()`, the verbatim tool call will be processed once when the text is `\n{...}\n` and once when the text is `\n{...}`, so there would be no problem.

If we can convert strings like `\n{...}\n\n{...}` into two JSON objects properly(can be multiple not just two), it would be okay. However, this is not really easy because the delimiter is unclear (the `\n\n` in between is a result of a combination of Qwen3's output format and our processing, and may differ depending on the model or processing logic).

While it's possible to read the content directly and use an algorithm to distinguish between them, I don't think it's necessary to remove the clear delimiter and then re-separate it with an algorithm.

jhlee525: Ok then we need to use another approach:

1. When `</tool_call>`  -> just stream delta (without finish reason)
2. When EOS occured -> end streaming delta (finish reason is determined by whether tool call occured or not)

We need to handle this problem since currently it always stop with "stop" finish reason (regardless of tool call or not).

jhlee525: After discussion, we decided to disable parallel tool calling for now, later find a solution for handle it.

PR #277, opened 2025-11-05
Revise web example with wasm tutorial docs

khj809:
Revised the web-assistant-ui example code with latest APIs, and updated WASM tutorial docs according to this.

## Additional updates
- Improved cache download logics by adding retry on failure (with `sleep` util for both native and wasm).
- Added `Ellipsis` util trait to truncate long string with ellipsis (used on truncating bytes on Python)
- Due to the existing custom `Serialize` and `Deserialize` for `Bytes`, it's not represented as Uint8Array as expected in wasm. Since serde-wasm-bindgen natively supports `serde_bytes::ByteBuf`, I removed our custom `Serialize` and `Deserialize` and let binding libraries handle them.
- `PartImage` with `binary` type now holds the original image bytes provided from users, instead of pixel matrix data. While I worked on web example, it's often required to get the original bytes back to make base64 data url, but there was no way to do it easily.
- The structs using tsify converts `Value` to Javascript Map instead of plain object, which is not an expected behavior. I found that even we implemented `From<Value>` to return object correctly, the ser/de via tsify does not respect this. So I added `hashmap_as_object` in structs using tsify and including `Value` type fields.
- Added `addTools` and `removeTools` in wasm agent. (but the ownership of tools argument is transferred to agent, as mentioned in #224)
- Removed `vite-plugin-top-level-await` plugin, which introduces undefined errors.

### DIFF
--- a/examples/web-assistant-ui/src/components/assistant-ui/thread.tsx
+++ b/examples/web-assistant-ui/src/components/assistant-ui/thread.tsx
@@ -245,7 +245,7 @@ const AssistantMessage: FC = () => {
         </div>
 
         <div className="text-foreground col-span-2 col-start-2 row-start-1 ml-4 leading-7 break-words">
-          <MessagePrimitive.Content
+          <MessagePrimitive.Parts
             components={{
               Text: MarkdownText,
               tools: { Fallback: ToolFallback },


src/utils/ellipse.rs
```diff
@@ -0,0 +1,60 @@
+use std::borrow::Cow;
+use unicode_segmentation::UnicodeSegmentation;
+
+pub trait Ellipse {
```
grf53: 'Ellipse' is a geometric shape, while 'Ellipsis' is the `...`.
As I know, `Ellipsis` would be correct for this one.

docs/docs/tutorial/webassembly-supports.mdx
grf53, line 57: My thought after seeing this part is that it would be good to provide a way to check it right away without separate installation using something like 'jsdelivr'.
If we use 'jsdelivr', we could utilize github release based URL(`https://cdn.jsdelivr.net/gh/user/repo@version/file`).

src/utils/ellipse.rs
khj809: Wow my english sucks.. Thanks

docs/docs/tutorial/webassembly-supports.mdx
khj809: Yes using CDN-based assets is also a viable solution, and it's not limited to using `isWebGPUSupported` only.
Maybe we can add a description about CDN in "Installation" section, which does not exist for now.

grf53 (approved):

khj809:
@jhlee525 @grf53 I found that there was no description and implementation for reasoning, so I added them.

PR #287, opened 2025-11-07
fix: bug in web js value conversion

grf53:
## Description
Similar issue with #273 but in wasm binding.
The web js conversion code was originally in a form that seemed to take this issue into account, but in reality it was a problematic code and this PR fixes the issue.


jhlee525 (approved):
👍

PR #300, opened 2025-11-10
Introduce bump-my-version

khj809:
This PR introduces [bump-my-version](https://callowayproject.github.io/bump-my-version/) for handling systematic version bumps.
- For git tags, we strictly follows semantic versioning with `v` prefix.
- The version of Rust crates, NPM packages strictly follows semantic versioning handled by bump-my-version.
- The version of Python package is dynamically converted from version written in `Cargo.toml` into PEP440-compatible version by maturin.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -34,7 +34,8 @@ pre_commit_hooks = []
 post_commit_hooks = []
 
 [tool.bumpversion.parts.pre_l]
-values = ["dev", "rc", "final"]
+values = ["", "rc", "final"]
+first_value = ""
 optional_value = "final"
 
 [[tool.bumpversion.files]]


khj809:
> Do we plan to always go through these steps for releases going forward? If we don't have `pre_l`, `pre_n` in the configurations, I think that would be: `0.2.0 --(patch)-> 0.2.1`.

The prerelease part is optional, so we don't have to consider the prerelease bumps every time.
If we decided not to create the prerelease version, it's enough to just run `bump-my-version bump patch`.

> If the intermediate steps don't leave any traces globally, then maybe it would be okay. (I think it would be enough to just push out the content that needs global reflection.) I'm concerned that if this tool were to automatically push git tags or something, these visible in the repo would look overly complex, or giving a pointless hassle on bumping versions.

I didn't mentioned, but the config has `tag = false` so git tags are not generated from this command. We assume that git tags are always generated by creating releases from Github.

grf53:
I checked that this configuration makes the `"final"` as the default of `pre_l` part. So we could have `0.2.1` by running `bump patch` on `0.2.0`.
```
[tool.bumpversion.parts.pre_l]
values = ["final", "dev", "rc"]
optional_value = "final"
```

jhlee525 (approved):

PR #301, opened 2025-11-10
Fix dependabot alerts

khj809:
- Fixed dependabot alert (https://github.com/brekkylab/ailoy/security/dependabot/54)
- Added dependabot config to ignore examples directories.


grf53 (approved):
:+1:

PR #309, opened 2025-11-13
File versioning in cache manifest

khj809:
This PR implements the file versioning in `_manifest.json` schema, which enables using appropriate cache files corresponding to the current ailoy version.
- Each key in the inner map represents the minimum available ailoy version. The largest minimum version less than or equal to current ailoy version is selected. In the below example, if the ailoy version is `0.2.1`, the file with `0.2.0` is selected.
```
{
    "rt.dylib": {
        "0.3.0": {
            "size": 1282914,
            "sha1": "040ba4836746c33166b60341cade6724fde73dcb"
        },
        "0.2.0": {
            "size": 1169576,
            "sha1": "1c842d06300e4d3d4880b9a36279597a7d541e97"
        },
        "*": {
            "size": 867856,
            "sha1": "a41c36b11632bb70eac7bd839839a479d8ea3ad1"
        }
    }
}
```
- After downloading the manifest, it's written in the cache directory along with the cache entries. If failed to download the manifest (due to internet connection or any other issues), the one in local filesystem is used instead.

## Notes
- This feature is required for #308 since the model lib files should be updated according to the latest tvm version. But anyway the new manifest structure may not work for Ailoy 0.2.0, so we need to eventually yank 0.2.0 later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "879b6c89592deb404ba4dc0ae6b58ffd1795c78991cbb5b8bc441c48a070440d"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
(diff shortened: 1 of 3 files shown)


grf53 (approved):
For me, it looks okay.

PR #310, opened 2025-11-14
[JS] Consolidate LangModel/EmbeddingModel config

khj809:
This PR merges the parameters `deviceId` and `progressCallback` into a single config object in both Nodejs and WASM.
Also updated the docs affected by this change.


docs/docs/tutorial/rag-using-documents.mdx
grf53: Let's remove this line(and the same codes around the documents), then. This maybe was added for consistency with JS codes.

jhlee525 (approved):

Issue #282, opened 2025-11-05
Add static methods `*Config.from_dict()` for Python


grf53:
**Is your feature request related to a problem? Please describe.**
In Python, add the `.from_dict(config: dict)` static method to Config files such as InferenceConfig.

PR #312, opened 2025-11-17
Get pre-defined `DocumentPolyfill` from its name

### LINKED ISSUE #304: A better way to get `DocumentPolyfill`
Currently, there is only one document polyfill, Qwen3, and it is got explicitly with `ai.getQwen3Polyfill()`.
This way is not very scalable, so it is better to change it to get ones based on strings, such as `ai.Tool.new_builtin("terminal")` when getting builtin tools.
Maybe it could be `ai.DocumentPolyfill.get("Qwen/Qwen3-0.6B")` or `ai.DocumentPolyfill.get("Qwen3")`

grf53:
## Ticket
resolves #304 

## Description
This PR adds `DocumentPolyfill.get(kind)` for Python and `getDocumentPolyfill(kind)` for JS to get the pre-defined `DocumentPolyfill`. (Currently `kind` is only `"Qwen3"`.)
This allows users to get `DocumentPolyfill`s by name, and also forms the basis for resolving #282.


src/model/polyfill.rs
```diff
@@ -1,8 +1,19 @@
...
 
 use crate::value::{Document, Message, Part, Role};
 
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Display, EnumString)]
+#[cfg_attr(feature = "python", derive(ailoy_macros::PyStringEnum))]
+#[cfg_attr(feature = "nodejs", napi_derive::napi(string_enum))]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(from_wasm_abi, into_wasm_abi))]
+pub enum DocumentPolyfillKind {
+    Qwen3,
```
khj809, line 14: Honestly, I don't think the current existing polyfill is for Qwen3 models only, so I'm quite hesitant on using the name "Qwen3" (it was my old thought on the function name `getQwen3Polyfill` too).
But I don't have any alternatives to suggest, so let's use this for now.

grf53: That seems reasonable. We could do this for now, and then later, if we find a better name, we could provide compatibility through aliases or something like that.

PR #313, opened 2025-11-17
Add static methods `*Config.from_dict()` for Python

### LINKED ISSUE #282: Add static methods `*Config.from_dict()` for Python

grf53:
## Ticket
resolves #282 

## Description
This PR adds static methods `.from_dict()` for `AgentConfig`, `InferenceConfig`, and `KnowledgeConfig` to generate those from Python dict.
Example usages are in `tests/test_configs.py`.

## Notes
By #312, it is possible to get `DocumentPolyfill` by its name, but for JS (Node and Web) it is not possible to include it as a string(`DocumentPolyfillKind`) like `"Qwen3"` in `InferenceConfig`. Still need to include it as like:
```
{
  documentPolyfill: getDocumentPolyfill("Qwen3")
}
```
instead of
```
{
  documentPolyfill: "Qwen3"
}
```

To support the latter one, we might patch the `DocumentPolyfill` to be `DocumentPolyfill | DocumentPolyfillKind`.
But it is not simple, and also not in the range of this PR.


jhlee525 (approved):

src/model/language_model.rs
```diff
@@ -509,8 +515,57 @@ mod py {
...
+                .get_item("max_tokens")?
+                .and_then(|max_tokens| python_to_value(&max_tokens).ok())
+                .and_then(|max_tokens| max_tokens.as_integer())
+                .map(|max_tokens| max_tokens as i32);
+            // let grammar = config
+            //     .get_item("grammar")?
+            //     .and_then(|grammar| python_to_value(&grammar).ok())
+            //     .and_then(|grammar| {
+            //         serde_json::from_str::<Grammar>(grammar.as_str().unwrap_or("{}")).ok()
+            //     });
```
khj809, line 560: It would be better to add a comment why this has been commented (or todo).

PR #319, opened 2025-11-24
refactor: split faiss cxx bridge as ailoy-faiss-sys crate

khj809:
When rebuilding Ailoy with different profiles or feature sets, the entire C++ shim is rebuilt each time. This causes significant development overhead—both in build time and in the size of generated artifacts.
This PR reduces that overhead by splitting our C++ shims into separate crates, starting with FAISS (which is simpler to isolate than TVM).

I introduced the `ailoy-faiss-sys` crate, responsible for the FAISS C++ bridge on native targets. Instead of pulling FAISS via CMake `FetchContent`, the project now uses a dedicated git submodule.

On the Web side, the FAISS shim has also been cleaned up: the redundant wrapper around the FAISS index has been removed, and we now use `FaissIndexInner` as defined directly by Emscripten.

I initially experimented with creating an `ailoy-faiss-web-sys` crate as well, but this added substantial complexity due to `#[wasm_bindgen]` with `raw_module` path pointing to the path inside subcrate, which should be eventually bundled in the same relative path. Therefore, I retain the current Web implementation.

## Additional Fixes
- Updated Linux Dockerfile to install reference-lapack instead of openblas, in order to maintain `build.rs` consistantly across different platforms.
- Updated Windows Dockerfile to set `LD_LIBRARY_PATH` so workflows don't need to set it manually, and slightly reduced build time by downloading installers in parallel.
- `cxx` started to generate mangled symbol names from 1.0.189 (see https://github.com/dtolnay/cxx/pull/1665), but our locked `cxx` version (1.0.186) didn't follow this and resulted in undefined symbol errors.
  So we need to align the locked `cxx` version and `cxxbridge-cmd` version used in our workflow to be identical (also for local dev environments).
  Eventually, I'm planning to totally remove `cxxbridge-cmd` and use `cxx_build` similar to `ailoy-faiss-sys`, so we don't need to consider about version disparity between `cxx` and `cxxbridge-cmd` at all.

### DIFF
--- a/.github/workflows/build_node_linux.yaml
+++ b/.github/workflows/build_node_linux.yaml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install cxxbridge (required to build shim)
         run: |
-          cargo install cxxbridge-cmd
+          cargo install cxxbridge-cmd@1.0.189
 
       - name: Install dependencies
         working-directory: ${{ github.workspace }}/bindings/nodejs
--- a/.github/workflows/build_node_macos.yaml
+++ b/.github/workflows/build_node_macos.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install cxxbridge (required to build shim)
         run: |
-          cargo install cxxbridge-cmd
+          cargo install cxxbridge-cmd@1.0.189
 
       - name: Install dependencies
         working-directory: ${{ github.workspace }}/bindings/nodejs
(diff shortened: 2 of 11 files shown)


khj809:
@jhlee525 @grf53 Please make sure to update cxxbridge-cmd to the latest (1.0.189) in your environment when testing this PR.

grf53 (approved):
Successfully confirmed that FAISS works well with this!

PR #328, opened 2025-11-28
feat(tool): add new builtin tools `web_search_duckduckgo` and `web_fetch`

khj809:
This PR introduces new builtin tools for searching webpages via DuckDuckGo (`web_search_duckduckgo`) and fetching url contents (`web_fetch`).
The implementation mostly follows the existing MCP server https://github.com/nickclyde/duckduckgo-mcp-server.

## Notes
- `web_search_duckduckgo` is not available on web environments due to the CORS policy, so it returns error on such case similar to `terminal` tool.

## Additional Fixes
- Without `rlib` in crate-type, the CLI modules cannot import `ailoy` as extern crate, so added it. (@jhlee525 Please review if this does not make any issue with our set-crate-type scripts)
- Fixed `PyStringEnum` macro not handling explicit strum and serde renames.
- The versions in nodejs's `ailoy_core.js` were not updated by bump-my-version, so added configuration for this file.

### DIFF
--- a/bindings/nodejs/tests/agent.test.ts
+++ b/bindings/nodejs/tests/agent.test.ts
@@ -153,9 +153,7 @@ for (const cfg of modelConfigs) {
     test.sequential(
       "Tool Calling: Builtin Tool (web search)",
       async () => {
-        const tool = ailoy.Tool.newBuiltin("web_search_duckduckgo", {
-          baseUrl: "http://www.naver.com",
-        });
+        const tool = ailoy.Tool.newBuiltin("web_search_duckduckgo");
         agent.addTool(tool);
 
         for await (const resp of agent.runDelta(


docs/docs/resources/builtin-tools.mdx
```diff
@@ -30,3 +34,42 @@ This tool is not available on web environments.
...
 | `stderr`    | string |  stderr of the executed command.   |
 | `exit_code` | number | Exit code of the executed command. |
+
+## `web_search_duckduckgo`
+
+Search webpages using DuckDuckGo and return formatted results.
+
+:::info
+
+This tool is not available on web environments.
```
jhlee525:
```suggestion
This tool may not work in web environments due to [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) restrictions. To avoid this issue, configure your requests to pass through a proxy server that forwards the call to the original endpoint.
```

khj809: This means that our builtin tools should be configurable.
I implemented this by adding optional `config` args in `newBuiltin()` methods and each tool creators use this.
Accordingly, I added actual proxy server in the global setup of web tests, and verified it works correctly.
Also updated the docs to refer this proxy server implementation.

jhlee525 (approved):

PR #332, opened 2025-12-08
feat(web): add `proxy_url` config to `web_fetch` tool

khj809:
Similar to `web_search_duckduckgo` tool, it's almost impossible to use `web_fetch` tool from web browser environment due to the CORS restriction on most webpages. So I added `proxy_url` parameter to let users configure this tool to forward requests to their own proxy server.

## Additional fixes
- Modified web search tools to rotate User-Agent header, in order to minimize the probability of getting blocked by DuckDuckGo bot detection.
- Introduced `fancy_regex` to unify the regex usages.

### DIFF
--- a/docs/docs/resources/builtin-tools.mdx
+++ b/docs/docs/resources/builtin-tools.mdx
@@ -88,10 +88,10 @@ implementation in our test codes.
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


grf53 (approved):
LGTM!

PR #333, opened 2025-12-09
feat(message): Generate random tool call ID if not exist

khj809:
Currently, both local models and Gemini models do not provide a tool call ID. As a result, tool call messages are sent with an empty `id` field. This can make it difficult to match tool calls and their results at the application layer. To address this, we've decided to always include a tool call ID by generating a random string if one is not provided by the model.
- **Local models** will generate the ID in the format `call-{random hex string}`.
- **Gemini models** will use the format `{tool name}/call-{random hex string}`, ensuring the tool name is preserved during message marshaling and guaranteeing the uniqueness of the ID.

## Additional fixes
- Fixed a bug in the `image_binary()` method of `Part`.
- Improved the base64 serialization of PNG images using the `png` crate.
- Fixed Gemini to handle URL images correctly when the URL is a base64 data URI.
- Unified logging usages and removed warnings from utils mods.

### DIFF
--- a/src/model/local/local_language_model.rs
+++ b/src/model/local/local_language_model.rs
@@ -644,7 +644,7 @@ mod tests {
         let mut strm = model.infer_delta(msgs, Vec::new(), Vec::new(), config);
         while let Some(out) = strm.next().await {
             let out = out.unwrap();
-            crate::debug!(format!("{:?}", out));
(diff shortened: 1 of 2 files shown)


grf53 (approved):
LGTM

PR #335, opened 2025-12-11
feat(cache): Expose model management APIs 

khj809:
This PR aims to expose model management APIs such as `download()` and `remove()` on binding level, with a huge refactoring on cache internal logics.

## Major updates
- Introduced `validate_checksum` parameter to control checksum validation, in order to bypass SHA1 calculation on unnecessary situations.
- Extracted the preparation of cache entries in `try_create()` into a separate `prepare_files()` function. This is reused on both `try_new()`, `try_new_stream()` (prepare files + instantiation) and `download()` (prepare files only)
- Added `download()` and `remove()` to both `LocalLanguageModel` and `LocalEmbeddingModel`, together with corresponding static methods in bindings.
- Added `clear_tools()` which clears all the tools from agent (this was necessary while building webui demo).

## Chore
- Added `boxed!` macro to unify `BoxFuture` and `BoxStream` usages for wasm32 and non-wasm32 environments.
- Updated huggingface tokenizers to 0.22.2.

### DIFF
--- a/src/model/local/local_embedding_model.rs
+++ b/src/model/local/local_embedding_model.rs
@@ -111,14 +111,7 @@ impl LocalEmbeddingModel {
         model: impl Into<String>,
         config: Option<LocalEmbeddingModelConfig>,
     ) -> anyhow::Result<Self> {
-        let config = config.unwrap_or_default();
-        let cache = Cache::new();
-        let mut ctx = HashMap::new();
-        if let Some(device_id) = config.device_id {
-            ctx.insert("device_id".to_owned(), Value::integer(device_id.into()));
-        };
-        let mut strm =
-            Box::pin(cache.try_create::<Self>(model, Some(ctx), config.validate_checksum));
+        let mut strm = Self::try_new_stream(model, config);
         while let Some(v) = strm.next().await {
             if let Some(result) = v?.result {
                 return Ok(result);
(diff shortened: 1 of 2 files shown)


src/model/local/model_common.rs
```diff
@@ -1,11 +0,0 @@
...
-
-use crate::value::Value;
-
-pub(crate) fn get_cache_context(device_id: Option<i32>) -> HashMap<String, Value> {
-    let mut ctx = HashMap::new();
-    if let Some(device_id) = device_id {
-        ctx.insert("device_id".to_owned(), Value::integer(device_id.into()));
-    }
-    ctx
-}
```
grf53, line 11: `get_cache_context()` looks replaced.
Could you please explain how the code that previously used get_cache_context() has been replaced and what improvements have been made through this?

khj809: The role of this function is creating a hashmap containing `device_id`, and it was used in both LocalLangModel, LocalEmbeddingModel and their binding constructors.
This is now handled by the constructor (`try_new_stream`) of `LocalLangModel` and `LocalEmbeddingModel` explicitly, and the constructors of bindings depend on this directly. So it's not necessary to keep this function anymore.

khj809: BTW, I reduced the duplicated logics further in the last commit. PTAL.

PR #336, opened 2025-12-11
chore(mcp): update rmcp to latest

khj809:
Updates rmcp to the latest version: [0.11.0](https://github.com/modelcontextprotocol/rust-sdk/releases/tag/rmcp-v0.11.0)
Still it doesn't work on wasm32 with streamable http transport.


grf53 (approved):
Version bump doesn't seem to have problem, so it looks good to me to be merged.
But the motivation of this looks not shared. What made you do this?

PR #337, opened 2025-12-12
fix(openai): adjust minimum reasoning effort value

khj809:
According to [openai's documentation](https://platform.openai.com/docs/api-reference/chat/create#chat_create-reasoning_effort):
- The minimum allowed value of `reasoning_effort` for `gpt-5.1`, `gpt-5.2` is `none`.
- All models before `gpt-5.1` does not support `none`, so it should be set to `minimal`.

## Discussion
All models after `gpt-5.1-codex-max` supports the new reasoning effort `xhigh`, but we don't have the corresponding value in our enum. Should we support this? @jhlee525 @grf53

### DIFF
--- a/src/model/api/gemini.rs
+++ b/src/model/api/gemini.rs
@@ -139,40 +139,50 @@ impl Marshal<ToolDesc> for GeminiMarshal {
 
 impl Marshal<RequestConfig> for GeminiMarshal {
     fn marshal(&mut self, config: &RequestConfig) -> Value {
-        let (include_thoughts, thinking_budget) = if let Some(model) = &config.model
-            && matches!(
-                model.as_str(),
-                "gemini-2.5-pro" | "gemini-2.5-flash" | "gemini-2.5-flash-lite"
-            ) {
-            match &config.think_effort {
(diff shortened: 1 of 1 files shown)


src/model/api/openai.rs
```diff
@@ -132,10 +132,10 @@ impl Marshal<RequestConfig> for OpenAIMarshal {
         let (reasoning_effort, reasoning_summary) = if is_reasoning_model {
             match &config.think_effort {
                 ThinkEffort::Disable => {
-                    if model.starts_with("gpt-5") {
-                        (Some("minimal"), None)
+                    if model.starts_with("gpt-5.2") || model.starts_with("gpt-5.1") {
```
grf53: I just checked that `gpt-5-pro` only supports `high` and `gpt-5.2-pro` only supports `medium`, `high`, and `xhigh` and the requests on those models with the unsupported effort option fail.
So these should be handled separately.
And we should maintain the `low` as well for the `o` models.

If you don't mind, let me apply those fixes on this.

khj809: Thanks! I'd be really appreciated if you help that.

PR #355, opened 2026-01-12
fix(workflow): Refactor build workflows

khj809:
This PR refactors build workflows as below:
- Separated publish job from build job for nodejs workflows, and applied retry on intermittent publish failure.
- Added super workflow for Python (`build_python.yaml`) and Node.js (`build_nodejs.yaml`) to trigger builds for all platforms at once.
- Added super super workflow `build_all.yaml` to trigger builds for all bindings for all platforms at once. This can be manually triggered or triggered via publishing new releases.


khj809:
@jhlee525 @grf53 These workflows cannot be tested before merged, so let me merge this first.

PR #356, opened 2026-01-14
fix(workflow): Fix broken workflows

khj809:
This PR fixes broken workflows from #355.
Verified in https://github.com/brekkylab/ailoy/actions/runs/20984161820 with expected publish failures.


grf53 (approved):
Thank you for keeping eyes on things while tasks are flooding!

PR #363, opened 2026-04-06
Add helper constructors for LangModelProvider

khj809:
In order to create a new `LangModelProvider`, users should be aware of each provider's API endpoint, as well as importing `url::Url` and `LangModelAPISchema` too, which results in poor developer experience.
This PR adds some helper constructors of `LangModelProvider` for each well-known API providers, so users don't need to import any other. Users can still provide arbitrary API endpoints via `LangModelProvider::chat_completion()`.

### DIFF
--- a/src/agent/rt/lang_model/mod.rs
+++ b/src/agent/rt/lang_model/mod.rs
@@ -15,13 +15,13 @@ pub struct LangModelRuntime {
     provider: LangModelProvider,
 }
 
-struct LangModelRequest {
-    pub model: String,
-    pub messages: Vec<Message>,
-    pub tools: Vec<ToolDesc>,
-    pub url: Url,
-    pub api_key: Option<String>,
-    pub infer_config: LangModelInferConfig,
+struct LangModelRequest<'a> {
+    pub model: &'a str,
+    pub messages: &'a [Message],
+    pub tools: &'a [ToolDesc],
+    pub url: &'a Url,
+    pub api_key: &'a Option<String>,
+    pub infer_config: &'a LangModelInferConfig,
 }
 
 impl LangModelRuntime {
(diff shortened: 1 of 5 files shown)


src/agent/rt/lang_model/mod.rs
```diff
@@ -15,13 +15,13 @@ pub struct LangModelRuntime {
     provider: LangModelProvider,
 }
 
-struct LangModelRequest<'a> {
-    pub model: &'a str,
-    pub messages: &'a [Message],
-    pub tools: &'a [ToolDesc],
-    pub url: &'a Url,
-    pub api_key: &'a Option<String>,
-    pub infer_config: &'a LangModelInferConfig,
```
jhlee525, line 24: 1. The purpose of `LangModelRequest` struct is to perform marshaling and produce a JSON object that can be sent as a request.
2. The current `Marshal` trait does not require ownership of the target structures.
3. Therefore, `LangModelRequest` itself does not need to own any of its members.
4. Based on this, functions such as `LangModel::infer` are also designed to take references rather than requiring ownership that would necessitate cloning.

If this design changes, the flow would become inefficient:

1. `LangModel` receives data by reference,
2. then unconditionally copies it to take ownership,
3. and finally passes a reference of that copied data to marshal.

Although this may seem like a small difference, repeatedly performing memory allocations and re-referencing without clear benefits can make 1. variable tracking more difficult and 2. potentially degrade performance.

jhlee525 (commented):
I agree that the helper function is needed, but have opinion of changing the signature of `LangModelRequest`.

Possible solution
1. Make `Marshal` taking ownership of target, consumes it
2. Keep it referencing

src/agent/rt/lang_model/mod.rs
khj809: I understood your concern, so reverted `LangModelRequest` as before.

jhlee525 (approved):
👍

PR #364, opened 2026-04-06
Add web_search builtin tool

### LINKED ISSUE #366: Implementing web search tool
Implemented in brekkylab/ailoy#364. Now needs to be integrated into the main agent for validation.

khj809:
This PR adds `web_search` builtin tool, which searches the query keywords across multiple search engines including below providers:
- Bing
- Brave
- DuckDuckGo
- Google
- LibreX
- Mojeek
- Startpage
- Yahoo
- Yandex

The results are aggregated and ranked by relevance score (number of engines that returned the same url).

For Bing, it requires TLS fingerprint impersonation to bypass bot detection. That is handled by [wreq](https://github.com/0x676e67/wreq), but it depends on BoringSSL written in C++. If this makes the whole package deployment too complicated, we may need to consider removing Bing provider entirely.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -560,3 +560,6 @@ test_*
 
 # omc
 .omc
+
+# git worktrees
+.worktrees/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,23 @@ dependencies = [
  "url",
  "urlencoding",
  "uuid",
+ "wreq",
+ "wreq-util",
+]
+
+[[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
 ]
 
 [[package]]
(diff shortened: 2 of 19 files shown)


nuri-yoo (approved):
Build and all tests pass (unit + network integration).

LGTM

nit: Bing creates a new `wreq::Client` on every `search()` call (`bing.rs:90-94`), which repeats the BoringSSL TLS handshake each time. Storing it as a struct field would avoid that cost. Also, since `wreq`/BoringSSL pulls in a heavy C++ build chain (`boring-sys2`, `bindgen`, `cmake`, `clang-sys`) for this one engine, it might be worth putting Bing behind a feature flag.

Ref : https://docs.rs/wreq/5.3.0/wreq/struct.Client.html

jhlee525 (approved):
LGTM 👍 

P.S. Personally, I’m not sure whether a consensus-based approach across multiple web search engines provides enough benefit relative to its heaviness.

In the end, it’s essentially similar to the principle of an ensemble, but ensembles tend to bias toward the worst-performing engine’s results.

For example: If we have search engine A(performance score 100), B(performance score 100), C(performance score 50)

In a consensus-based setup, the final result is derived by aggregating outputs from all three engines.

The issue is that even though A and B are highly reliable, the incorrect or low-quality outputs from C are still incorporated into the aggregation process. As a result, C’s errors can dilute or distort the final outcome, effectively pulling the overall result away from the optimal answers provided by A and B.

khj809:
@jhlee525 I understand your concern. Instead of a simple vote counting, I applied Reciprocal Rank Fusion (https://www.elastic.co/docs/reference/elasticsearch/rest-apis/reciprocal-rank-fusion) to compensate for higher ranked results.

khj809:
@nuri-yoo 

> nit: Bing creates a new `wreq::Client` on every `search()` call (`bing.rs:90-94`), which repeats the BoringSSL TLS handshake each time. Storing it as a struct field would avoid that cost.

Creating `wreq::Client` itself does not do networking, so there's no actual handshaking happened. 
In terms of reusing connections, it would be meaningful if we send quite many requests via the same client. But considering the tool usage scenario (`web_search` is not usually invoked massively in a short time), the benefit of reusing the client seems negligible.

> Also, since wreq/BoringSSL pulls in a heavy C++ build chain (boring-sys2, bindgen, cmake, clang-sys) for this one engine, it might be worth putting Bing behind a feature flag.

I agree that we can make the Bing search feature as feature-gated, but it's not clear when it comes to Python / JS packaging. Let's discuss about how to deal with Bing search in the future.

PR #365, opened 2026-04-07
introducing virtual shell & AgentState

### LINKED ISSUE #367: Introducing agent shell
Many modern LLM-based agents (e.g., code agents like Claude Code, OpenAI Codex CLI, Open Interpreter, and Auto-GPT) rely on a shell-based environment. In practice, tool interactions are most effective when operating within a local terminal-like interface, where commands, file system access, and process execution are naturally integrated.

However, in server-based environments, providing a native shell interface is non-trivial. Inspired by [just-bash](https://github.com/vercel-labs/just-bash), I propose a virtual shell that appears to the LLM as a standard bash-like tool, while internally being fully emulated and sandboxed.

## AgentState

The scope of state management is no longer limited to message history; the agent shell is now also part of it.

Therefore, we introduce a new struct called AgentState, which combines both components. This struct is responsible for maintaining the current state of each agent session.

AgentState is defined as follows:

```rust
struct AgentState {
  history: Vec<Message>,
  shell: Option<AgentShell>
}
```

The shell will be integrated with tools such as a Bash-like interface, enabling the agent to execute commands within the virtual shell environment.

jhlee525:
inspired by https://github.com/vercel-labs/just-bash


src/state/state.rs
```diff
@@ -0,0 +1,21 @@
...
+            history: Vec::new(),
+            shell: None,
+        }
+    }
+
+    pub fn shell(mut self) -> Self {
+        self.shell = Some(Shell::new());
+        self
+    }
+}
```
khj809: Would be better to move this file under `src/agent`, and rename `src/state` as `src/shell`.

jhlee525: 👍

khj809 (approved):
LGTM

It would be nice to see if the bash commands generated by agents actually works on either in-memory or filesystem.

grf53 (approved):
Looks like a nice beginning of implementation of virtual shell that can be used by Ailoy agents.

We also could refer to https://github.com/vercel-labs/bash-tool to connect this to the agents(of ai-sdk). This is because the part that allows the agent to use the bash environment created with just-bash actually appears to be located here.

nuri-yoo (approved):
LGTM

selloriwoo (approved):
LGTM

PR #372, opened 2026-04-08
feat: Add python_repl builtin tool

### LINKED ISSUE #371: New builtin tool: Python REPL
- Run Python scripts generated by LLM statelessly
- Use lightweight Python interpreter considering environments where Python/Docker is not installed
- Support external packages (e.g. numpy, matplotlib)
- The same venv should be reused during the same agent session

khj809:
## Summary

Implements a `python_repl` builtin tool that allows LLMs to execute Python scripts they generate. (#371)

- Works even if Python is not installed on the user's system (automatically provisions standalone Python via uv)
- Dynamically installs external libraries like `numpy`, `matplotlib` via the `pip_install` parameter
- Reuses venv within the same Agent Session for cumulative package installation
- Supports persistent venv paths that can be reused across sessions

---

## Background

LLM agents need a code execution environment to perform tasks such as data analysis, numerical computation, and chart generation. Pydantic Monty (sandboxed execution) was ruled out due to lack of external library support; [uv](https://github.com/astral-sh/uv) was adopted as the Python environment manager instead.

uv handles all of the following as a single static binary:
- Provisions the Python runtime by downloading [python-build-standalone](https://github.com/astral-sh/python-build-standalone) builds (same source as GitHub Actions `setup-python`)
- Creates venvs and installs pip packages (10–100× faster than pip)
- Cross-platform (Linux / macOS / Windows)

---

#### `uv.rs` — uv binary management

```
resolve_uv_path() search order:
  1. $UV environment variable (CI/test override)
  2. uv on system PATH
  3. <XDG cache dir>/ailoy/bin/uv[.exe]  (managed location)
```

Platform-specific managed paths:
| OS | Path |
|---|---|
| Linux | `~/.cache/ailoy/bin/uv` |
| macOS | `~/Library/Caches/ailoy/bin/uv` |
| Windows | `%LOCALAPPDATA%\ailoy\bin\uv.exe` |

#### `env.rs` — `PythonEnv`

| Feature | Behavior |
|---|---|
| `new_temp()` | Creates a temporary venv; automatically deleted on Drop |
| `new(venv_path, is_temp=false)` | Persistent venv — reusable across sessions |
| `install_packages(&[String])` | In-memory cache prevents duplicate installs for the same specifier; supports version specifiers (`numpy>=1.24`, `pandas==2.1.0`) |
| `run_code(code, timeout_secs)` | Executes via a temporary `.py` file; reads stdout/stderr concurrently and asynchronously; handles timeouts |

**Key design points:**

- **Pipe deadlock prevention**: Two `tokio::spawn` tasks drain stdout/stderr concurrently
- **Output truncation**: Output exceeding `MAX_OUTPUT_CHARS = 8_000` is truncated with a `[output truncated at N chars]` marker — prevents LLM context pollution from large outputs such as full numpy array dumps
- **`timed_out: bool`**: On timeout, returns `Ok(ExecResult { timed_out: true, .. })` instead of `Err`, so the LLM can clearly distinguish between runtime errors and timeouts
- **pip install failure handling**: Failures installing pre-configured packages (`packages` config) → AgentRuntime creation fails (caught early); failures installing LLM-requested packages → returns an error result with `phase: "pip_install"` (agent execution continues)

```rust
pub struct ExecResult {
    pub stdout: String,
    pub stderr: String,
    pub exit_code: i32,
    pub timed_out: bool,
}
```

#### `mod.rs` — `build_python_repl_tool()`

Tool schema exposed to the LLM:

```json
{
  "name": "python_repl",
  "parameters": {
    "code": "string (required)",
    "pip_install": "string[] — supports version specifiers"
  }
}
```

Returned result schema:

```json
{
  "stdout": "string",
  "stderr": "string",
  "exit_code": "integer",
  "timed_out": "boolean",
  "phase": "pip_install | execution | validation  (only on failure)"
}
```

---

### Modified: `src/agent/spec.rs`

```rust
pub enum BuiltinToolProvider {
    WebSearch {},
    PythonRepl {                          // new
        python_version: Option<String>,   // None → latest stable
        venv_path: Option<String>,        // None → temp (auto-cleanup)
        packages: Vec<String>,            // pre-installed packages
    },
}
```

---

### Modified: `src/agent/rt/tool/mod.rs`

Changed `ToolSet::with_builtin()` to `async fn` and added the `PythonRepl` match arm.

---

### Modified: `src/agent/rt/agent.rs`

Changed `AgentRuntime::new()` to `async fn` to propagate the async from `with_builtin()`.

#### Breaking change

```rust
// before
let agent = AgentRuntime::new(spec, provider, tool_set);

// after
let agent = AgentRuntime::new(spec, provider, tool_set).await?;
```

---

### Modified: `Cargo.toml`

Added dependencies:

| Crate | Purpose |
|---|---|
| `tempfile` | Temporary script files / temporary venv directories |
| `dirs` | XDG cache directory lookup |
| `which` | Locate uv binary on PATH |
| `shellexpand` | Expand `~` in venv_path |

---

### Unit tests (no uv required, always run)

| Test | Verifies |
|---|---|
| `test_managed_uv_path_is_under_cache_dir` | Path is under XDG cache directory |
| `test_managed_uv_path_structure` | Exact `<cache>/ailoy/bin/uv[.exe]` structure |
| `test_resolve_uv_path_env_override` | `$UV` environment variable takes priority |
| `test_uv_binary_name_platform` | Windows `.exe` / others `uv` |
| `test_truncate_output_short_string_unchanged` | Strings below limit are unchanged |
| `test_truncate_output_exact_limit_unchanged` | Boundary value handled correctly |
| `test_truncate_output_over_limit_is_shortened` | Truncates with marker when exceeded |
| `test_truncate_output_respects_utf8_boundary` | No panic at multibyte character boundaries |

### Integration tests (`#[ignore = "requires uv"]`)

| Test | Verifies |
|---|---|
| `test_create_temp_env_venv_dir_exists` | venv and Python binary creation |
| `test_temp_env_is_cleaned_up_on_drop` | Temporary directory deleted on Drop |
| `test_persistent_env_not_cleaned_up_on_drop` | Persistent venv survives Drop |
| `test_install_package_success` | pip install works correctly |
| `test_install_same_package_twice_uses_cache` | Same specifier uses cache, avoids subprocess |
| `test_install_version_specifier_change_reinstalls` | Version change triggers uv re-invocation |
| `test_install_nonexistent_package_returns_failed` | Non-existent package → `InstallResult::Failed` |
| `test_run_hello_world` | Basic execution and stdout capture |
| `test_run_code_captures_stderr` | stderr capture |
| `test_run_code_runtime_error_nonzero_exit` | Exception → non-zero exit code |
| `test_run_code_each_execution_is_stateless` | Variable isolation between executions |
| `test_run_code_timeout_returns_timed_out_flag` | Timeout → `timed_out: true`, returns `Ok` |
| `test_run_code_success_has_timed_out_false` | Successful execution → `timed_out: false` |
| `test_run_code_large_output_is_truncated` | Large stdout is truncated |
| `test_run_code_large_stderr_is_truncated` | Large stderr is truncated |
| `test_tool_name_is_python_repl` | Tool descriptor name |
| `test_tool_schema_requires_code` | `code` is a required parameter |
| `test_tool_schema_pip_install_is_array_of_strings` | `pip_install` schema type |
| `test_missing_code_param_returns_validation_error` | Missing `code` → `phase: validation` |
| `test_run_print_returns_stdout` | Full tool call pipeline |
| `test_pip_install_failure_returns_phase_pip_install` | pip failure → `phase: pip_install` |

### E2E integration test (`#[ignore = "requires uv + network + OPENAI_API_KEY"]`)

`test_python_repl_numpy_matplotlib_chart` (`src/agent/rt/agent.rs`):

1. Asks `gpt-4o-mini` to generate sine wave data with numpy and save a chart to a file using matplotlib
2. Verifies that the `python_repl` tool was called
3. Verifies `exit_code == 0` for all tool calls
4. Verifies the chart file exists
5. Validates PNG magic bytes (`\x89PNG\r\n\x1a\n`)
6. Verifies a final Assistant text response is present

### DIFF
--- a/src/agent/rt/tool/python_repl/env.rs
+++ b/src/agent/rt/tool/python_repl/env.rs
@@ -15,12 +15,37 @@ pub enum InstallResult {
     Failed { stderr: String },
 }
 
+/// Maximum number of UTF-8 characters retained from stdout / stderr.
+/// Output beyond this limit is replaced with a truncation notice so that
+/// huge prints (e.g. a raw numpy array) don't flood the LLM context window.
+pub const MAX_OUTPUT_CHARS: usize = 8_000;
+
+/// Truncate `s` to at most [`MAX_OUTPUT_CHARS`] characters, preserving valid
+/// UTF-8 boundaries, and append a notice when truncation occurs.
+pub fn truncate_output(s: String) -> String {
+    if s.len() <= MAX_OUTPUT_CHARS {
+        return s;
+    }
+    // Walk backward from MAX_OUTPUT_CHARS to find a valid char boundary.
+    let mut end = MAX_OUTPUT_CHARS;
+    while end > 0 && !s.is_char_boundary(end) {
+        end -= 1;
+    }
+    format!(
+        "{}\n[output truncated at {} chars]",
+        &s[..end],
+        MAX_OUTPUT_CHARS
+    )
+}
+
 /// Result of executing a Python script.
 #[derive(Debug)]
 pub struct ExecResult {
     pub stdout: String,
     pub stderr: String,
     pub exit_code: i32,
+    /// `true` when the script was killed because it exceeded the timeout.
+    pub timed_out: bool,
 }
 
 /// A managed Python virtual environment backed by `uv`.
@@ -158,31 +183,83 @@ impl PythonEnv {
     /// the exit code.
     ///
     /// The code is written to a temporary file and passed to the venv's Python
-    /// interpreter.  The temp file is deleted after the child exits.
+    /// interpreter.  stdout and stderr are read concurrently on separate tasks
+    /// to avoid pipe-buffer deadlocks.  Output is truncated at
+    /// [`MAX_OUTPUT_CHARS`] chars to protect the LLM context window.
+    ///
+    /// On timeout the child process is killed and [`ExecResult::timed_out`] is
+    /// set to `true`; the function still returns `Ok`.
     pub async fn run_code(&self, code: &str, timeout_secs: u64) -> Result<ExecResult> {
         // Write code to a temp file.
         let script = tempfile::NamedTempFile::with_suffix(".py")
             .context("failed to create temp script file")?;
         std::fs::write(script.path(), code).context("failed to write script to temp file")?;
 
-        let child = Command::new(self.python_path())
+        let mut child = Command::new(self.python_path())
             .arg(script.path())
             .stdout(std::process::Stdio::piped())
             .stderr(std::process::Stdio::piped())
             .spawn()
             .context("failed to spawn Python interpreter")?;
 
-        let deadline = tokio::time::Duration::from_secs(timeout_secs);
-        let output = tokio::time::timeout(deadline, child.wait_with_output())
-            .await
-            .context("Python script timed out")?
-            .context("failed to wait for Python process")?;
+        // Drain stdout and stderr on separate tasks to prevent pipe-buffer
+        // deadlock when a process produces large output on both streams.
+        let stdout_pipe = child.stdout.take().expect("stdout was piped");
+        let stderr_pipe = child.stderr.take().expect("stderr was piped");
 
-        Ok(ExecResult {
-            stdout: String::from_utf8_lossy(&output.stdout).into_owned(),
-            stderr: String::from_utf8_lossy(&output.stderr).into_owned(),
-            exit_code: output.status.code().unwrap_or(-1),
-        })
+        let out_task = tokio::spawn(async move {
+            use tokio::io::AsyncReadExt as _;
+            let mut buf = Vec::new();
+            tokio::io::BufReader::new(stdout_pipe)
+                .read_to_end(&mut buf)
+                .await?;
+            Ok::<_, std::io::Error>(buf)
+        });
+        let err_task = tokio::spawn(async move {
+            use tokio::io::AsyncReadExt as _;
+            let mut buf = Vec::new();
+            tokio::io::BufReader::new(stderr_pipe)
+                .read_to_end(&mut buf)
+                .await?;
+            Ok::<_, std::io::Error>(buf)
+        });
+
+        let deadline = tokio::time::Duration::from_secs(timeout_secs);
+        match tokio::time::timeout(deadline, child.wait()).await {
+            Ok(Ok(status)) => {
+                let stdout_bytes =
+                    out_task.await.unwrap_or(Ok(vec![])).unwrap_or_default();
+                let stderr_bytes =
+                    err_task.await.unwrap_or(Ok(vec![])).unwrap_or_default();
+                Ok(ExecResult {
+                    stdout: truncate_output(
+                        String::from_utf8_lossy(&stdout_bytes).into_owned(),
+                    ),
+                    stderr: truncate_output(
+                        String::from_utf8_lossy(&stderr_bytes).into_owned(),
+                    ),
+                    exit_code: status.code().unwrap_or(-1),
+                    timed_out: false,
+                })
+            }
+            Ok(Err(e)) => {
+                out_task.abort();
+                err_task.abort();
+                Err(anyhow::anyhow!("failed to wait for Python process: {}", e))
+            }
+            Err(_) => {
+                // Timeout — kill the child and abort I/O tasks.
+                let _ = child.kill().await;
+                out_task.abort();
+                err_task.abort();
+                Ok(ExecResult {
+                    stdout: String::new(),
+                    stderr: format!("script timed out after {}s", timeout_secs),
+                    exit_code: -1,
+                    timed_out: true,
+                })
+            }
+        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


nuri-yoo (approved):
LGTM

src/agent/rt/tool/python_repl/env.rs
```diff
@@ -0,0 +1,569 @@
...
+            .install_packages(&["requests".to_string()])
+            .await
+            .unwrap();
+        assert!(matches!(result, InstallResult::Success));
+    }
+
+    #[tokio::test]
+    #[ignore = "requires uv + network"]
+    async fn test_install_same_package_twice_uses_cache() {
+        let uv = uv_or_skip();
```
grf53: A quick question: does the `"requires uv + network"` or `uv_or_skip()` have some function to decide to perform or skip this test?

src/agent/rt/tool/python_repl/uv.rs
```diff
@@ -0,0 +1,325 @@
...
+/// been downloaded yet), downloads the latest `uv` release from GitHub.
+pub async fn ensure_uv() -> Result<PathBuf> {
+    let path = resolve_uv_path()?;
+    if path.exists() {
+        return Ok(path);
+    }
+    log::info!("uv not found — downloading to {}", path.display());
+    download_uv(&path).await?;
+    Ok(path)
+}
```
grf53, line 56: It seems like an approach that does not require the Python interpreter itself, but rather requires UV as a 'Python provider' and intends to use Python based on that. This was quite fresh to me.

Since Python interpreters vary widely in version and supply channels, this approach seems quite make sense.

We should utilize this to use docling for document conversions.

src/agent/rt/tool/python_repl/env.rs
khj809: No, there's no mechanism to check condition and decide to skip the test on `cargo test`.
So I think we should just early-return if `uv` is not found (with optional warning log indicating that `uv` does not exist). Let me fix them.

khj809: There's no mechanism to check condition at runtime and decide to skip the test on cargo test.
So I think we should just early-return if uv is not found (with optional warning log indicating that uv does not exist). Let me fix them.

khj809: @grf53 @nuri-yoo 
I've utilized [test-with](https://github.com/yanganto/test-with) to skip tests if `uv` executable is not found. But using the library as-is introduced unnecessary dependencies, so I tried to fix it on https://github.com/yanganto/test-with/pull/137. Until that is merged to upstream, we're using our forked version of [test-with](https://github.com/brekkylab/test-with).

grf53: I looked into the `test-with`, and I found that there are the features like `env`, `runtime_env`, `https`, etc.
I think we don't have a reason not to use those for `requires network` and `requires OPENAI_API_KEY`.

src/agent/rt/tool/python_repl/env.rs
```diff
@@ -0,0 +1,557 @@
...
+            bail!("uv venv failed for {:?}", self.venv_path);
+        }
+        Ok(())
+    }
+
+    /// Install pip packages into this venv.
+    ///
+    /// Specifiers that are already present in the in-memory cache are skipped;
+    /// for everything else `uv pip install` is called.
+    pub async fn install_packages(&self, packages: &[String]) -> Result<InstallResult> {
```
grf53: I doubt we need `Result<...>` in the return type.
`InstallResult` already implies failures, and as I see, every code calling this function calls `unwrap()` at the end. So it looks redundant.

Maybe it is because of the `?` in L168, which could be replaced with `InstallResult::Failure`. [Match expression](https://doc.rust-lang.org/reference/expressions/match-expr.html) can be used for that.

grf53: Or, we could use `Result<String>` or sth else, instead of `InstallResult`.
I am not so sure that we really need the `InstallResult` enum, if we just can replace it.

src/agent/rt/tool/python_repl/env.rs
```diff
@@ -0,0 +1,557 @@
...
+    #[test_with::executable(uv)]
+    #[tokio::test]
+    async fn test_install_empty_list_returns_already_installed() {
+        let uv = resolve_uv_path().unwrap();
+        let env = PythonEnv::new_temp(uv, None).await.unwrap();
+        let result = env.install_packages(&[]).await.unwrap();
+        assert!(matches!(result, InstallResult::AlreadyInstalled));
+    }
+
+    // ── code execution ───────────────────────────────────────────────────────
```
grf53, line 390: I think it would be good to have an example where you execute something using a package after installing it.

For instance, after installing requests, test sending a GET request to a URL like `https://jsonplaceholder.typicode.com/todos/1`.
To avoid network dependencies in the test, install PyYAML, load and dump a simple example yaml string, and then compare whether they are identical.

src/agent/rt/tool/python_repl/env.rs
khj809: @grf53 I also agree that there's no reason to return `Result` for this function. I've updated it to return `InstallResult` directly.

src/agent/rt/tool/python_repl/env.rs
khj809: @grf53 I applied `#[test_with::env(OPENAI_API_KEY)]` to those tests.

src/agent/rt/tool/python_repl/env.rs
khj809: Actually `test_python_repl_numpy_matplotlib_chart`(in `src/agent/rt/agent.rs`) already covers such example scenario by installing numpy, matplotlib and using them to generate a chart image.
However, for the self-containedness of `python_repl` module, it also makes sense to locate it under `python_repl`. I'll add it.

src/agent/rt/tool/python_repl/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+use env::{InstallResult, PythonEnv};
+use futures::future::BoxFuture;
+use uv::ensure_uv;
+
+use crate::{
+    agent::rt::tool::{ToolFunc, ToolRuntime},
+    datatype::Value,
+    message::ToolDescBuilder,
+};
+
```
grf53, line 15: I tried to make an ascii diagram to show the structure of this tool.
(cuz I had a little difficulty understanding it quickly.)

Adding this is totally **_optional_**. Just accept if you want.

```suggestion

/// ```text
/// +----------------------+                   call with (`pip_install`, `code`)
/// | PythonReplConfig     |                                 |
/// | - python_version     |                                 v
/// | - venv_path          |   +-----------------------------+--------------------+
/// | - packages           |-->| "python_repl" ToolRuntime                        |
/// +----------------------+   |                                                  |
///                            | +-----------+     +----------------------------+ |
///                            | | Acquire   | --> | Create PythonEnv           | |
///                            | | `uv`      |     | - prepare venv             | |
///                            | +-----------+     | - install initial packages | |
///                            |                   +----------------------------+ |
///                            +-----------------------------+--------------------+
///                                                          |
///                                                          v
///                                       Install additional packages `pip_install`
///                                                          |
///                                                          v
///                                                   Run python `code`
/// ```

```

src/agent/rt/tool/python_repl/env.rs
grf53: Thanks! After this, I think this would be good to be merged.

grf53 (approved):
Thank you for actively accepting my suggestions!

PR #374, opened 2026-04-09
fix: make `FileDyn` and `Directory` Sendable

khj809:
Due to the lack of `Send` in `FileDyn` and `Directory`, it makes the whole `AgentRuntime` un-Sendable.
This PR adds `Send` to `FileDyn` and `Directory`, and it's safe since the members of structs implementing `FileDyn` and `Directory` are all owned data type.


nuri-yoo (approved):
LGTM

PR #375, opened 2026-04-09
feat: Streaming Tool Results via `TurnEvent` and `ToolResultDelta`

### LINKED ISSUE #368: Add subagent connection logic

### LINKED ISSUE #373: Support yielding intermediate results from tool execution
Currently the tool execution is only handled in a synchronous mannar with a single tool result.
However, some long-running tools (especially subagents) may need to yield intermediate results, so users can recognize what's happening during the tool execution. This means we need to introduce a new concept of something like `ToolResultDelta` which is yielded iteratively until the tool execution is completed, and they should be aggregated into an atomic tool result before delivered to LangModel.

khj809:
## Motivation

Tool execution was fully synchronous — `stream_turn()` blocked on each `tool.run().await` before yielding a complete `MessageOutput`. For long-running tools like subagents, callers saw nothing until the entire execution finished.

This PR introduces incremental tool results that can be yielded during execution for UI progress, while still aggregating into an atomic `Message` before feeding back to the LLM.

### New types

**`TurnEvent`** (`src/agent/rt/turn_event.rs`) — replaces the flat `MessageOutput` yield from `stream_turn()`:
```rust
pub enum TurnEvent {
    AssistantMessage(MessageOutput),
    ToolCall { id: Option<String>, name: String, arguments: Value },
    ToolResult(Message),
    ToolDelta(ToolResultDelta),
}
```

**`ToolResultDelta`** + **`StreamingToolOutput`** (`src/message/tool_result.rs`) — intermediate progress snapshots and explicit streaming tool output:
```rust
pub struct ToolResultDelta { pub tool_call_id: Option<String>, pub tool_name: String, pub content: Part }

pub enum StreamingToolOutput {
    Delta(ToolResultDelta),  // UI progress only, not added to history
    Result(Value),           // Definitive result; must be the last item
}
```

`ToolResultDelta.content` is a `Part` (complete snapshot), not `PartDelta` — deltas are **not** accumulated into the final result. Only the explicit `StreamingToolOutput::Result` becomes the tool result in history.

### Three-tier tool function system

`ToolFuncKind` now has three meaningfully distinct variants:

| Variant | Type | Behavior |
|---|---|---|
| `Sync` | `ToolSyncFunc = Fn(Value) -> Value` | Direct return, no future |
| `Async` | `ToolAsyncFunc = Fn(Value) -> BoxFuture<Value>` | Single async result |
| `Streaming` | `ToolStreamingFunc = Fn(Value) -> BoxFuture<Stream<StreamingToolOutput>>` | Yields deltas + explicit result |

New constructors: `ToolRuntime::new_sync()`, `ToolRuntime::new_async()`, `ToolRuntime::new_streaming()`.

### Agent loop (`src/agent/rt/agent.rs`)

- **‼️ Breaking Change**: `stream_turn()` return type changed to `Stream<Item = Result<TurnEvent>>`
- Tool-calling turns yield only `TurnEvent::ToolCall` events (no redundant `AssistantMessage`)
- Final turns yield `TurnEvent::AssistantMessage`
- Streaming path: `StreamingToolOutput::Delta` → `TurnEvent::ToolDelta` (with `tool_call_id` injected from the agent's context); `StreamingToolOutput::Result` → builds atomic `Message` → pushed to history → `TurnEvent::ToolResult`
- Error recovery: streaming tool that never yields `Result` produces a synthetic error tool result so history stays consistent

### Parallel tool execution

When the LLM returns multiple tool calls in a single response, all tools are now executed concurrently via `tokio::spawn` + `mpsc` channel:

```
LLM response with [tool_call_A, tool_call_B]
    │
    ├── tokio::spawn → tool_A.run() ──────── 50ms ──→ sends ToolResult(A) to channel
    └── tokio::spawn → tool_B.run_stream() ─ 300ms → sends ToolDelta(B)... ToolResult(B) to channel
                                                       │
                                          main loop drains channel, yields events
                                          then pushes all ToolResult messages to history
                                          in completion order (A before B)
```

Key properties:
- `ToolRuntime` is `Clone + Send` (Arc-based), so cloning for each spawn is cheap
- `run_stream()` returns a `'static` + `Send` stream, polled independently per task
- History is updated **after all tools complete**, in completion order (not call order)
- Same subagent called twice: handled naturally via Mutex contention (sequential)
- Tool that never yields `StreamingToolOutput::Result`: synthetic error result is inserted to keep history consistent

### In-memory subagent now uses streaming (`src/agent/rt/tool/subagent/mod.rs`)

Intermediate `AssistantMessage` events from the inner agent become `StreamingToolOutput::Delta` items; the final assistant text becomes `StreamingToolOutput::Result`. 
The A2A subagent remains `Async` (single result), but will be changed to `Streaming`.

## Testing

- All 75 tests pass
- `test_streaming_subagent_emits_tool_deltas` — verifies `TurnEvent::ToolDelta` items are yielded during subagent execution
- `test_turn_event_sequence` — verifies the full sequence `ToolCall → ToolResult → AssistantMessage` for a simple tool call flow
- `test_parallel_tool_calls` — uses `claude-haiku-4-5` to trigger two parallel tool calls; verifies fast tool (50 ms) completes before slow tool (300 ms), and that history records them in completion order

### DIFF
--- a/src/agent/rt/agent.rs
+++ b/src/agent/rt/agent.rs
@@ -2,11 +2,12 @@ use std::pin::Pin;
 
 use futures::{Stream, StreamExt as _};
 
+use super::turn_event::TurnEvent;
 use crate::{
     agent::{
         AgentProvider, AgentSpec, LangModelRuntime, ToolKind, ToolProvider, ToolRuntime, ToolSet,
     },
-    message::{FinishReason, Message, MessageOutput, Part, Role},
+    message::{Message, Part, Role, StreamingToolOutput},
     shell::Shell,
 };
 
@@ -91,6 +92,10 @@ impl AgentRuntime {
         }
     }
 
+    /// Run a full turn and return the final assistant message.
+    ///
+    /// Drains [`stream_turn`](Self::stream_turn) and returns the message from
+    /// the last [`TurnEvent::AssistantMessage`] with `is_final: true`.
     pub fn run(
         &mut self,
         query: Message,
@@ -98,11 +103,12 @@ impl AgentRuntime {
         Box::pin(async move {
             let mut last = None;
             let mut strm = self.stream_turn(query);
-            while let Some(output) = strm.next().await {
-                last = Some(output?);
+            while let Some(event) = strm.next().await {
+                if let TurnEvent::AssistantMessage(output) = event? {
+                    last = Some(output.message);
+                }
             }
-            last.map(|o: MessageOutput| o.message)
-                .ok_or_else(|| anyhow::anyhow!("No assistant response"))
+            last.ok_or_else(|| anyhow::anyhow!("No assistant response"))
         })
     }
 
@@ -114,10 +120,19 @@ impl AgentRuntime {
         self.state.history = history;
     }
 
+    /// Stream all events for a single agent turn.
+    ///
+    /// Yields a sequence of [`TurnEvent`] items:
+    /// - [`TurnEvent::AssistantMessage`] for each LLM response
+    /// - [`TurnEvent::ToolCall`] for each tool call extracted from the response
+    /// - [`TurnEvent::ToolDelta`] for intermediate progress from streaming tools
+    /// - [`TurnEvent::ToolResult`] for each finalized tool result
+    ///
+    /// The loop continues until the LLM produces a response with no tool calls.
     pub fn stream_turn(
         &mut self,
         query: Message,
-    ) -> Pin<Box<impl Stream<Item = anyhow::Result<MessageOutput>> + '_>> {
+    ) -> Pin<Box<impl Stream<Item = anyhow::Result<TurnEvent>> + '_>> {
         Box::pin(async_stream::stream! {
             self.state.history.push(query);
             let tool_descs: Vec<_> = self.tools.iter().map(|t| t.desc().clone()).collect();
@@ -132,13 +147,25 @@ impl AgentRuntime {
                 };
                 let assistant_msg = output.message.clone();
                 self.state.history.push(assistant_msg.clone());
-                yield Ok(output);
 
-                let tool_calls = assistant_msg.tool_calls.unwrap_or_default();
+                let tool_calls = assistant_msg.tool_calls.clone().unwrap_or_default();
+
                 if tool_calls.is_empty() {
+                    // Final assistant response — yield the message and end the turn
+                    yield Ok(TurnEvent::AssistantMessage(output));
                     break;
                 }
+
+                // Tool-calling turn — yield one ToolCall per call, skip AssistantMessage
                 for tool_call in tool_calls {
+                    if let Some((id, name, arguments)) = tool_call.as_function() {
+                        yield Ok(TurnEvent::ToolCall {
+                            id: id.map(|s| s.to_string()),
+                            name: name.to_string(),
+                            arguments: arguments.clone(),
+                        });
+                    }
+
                     let tool = match self
                         .tools
                         .iter()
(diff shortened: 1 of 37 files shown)


src/agent/rt/agent.rs
```diff
@@ -150,21 +179,106 @@ impl AgentRuntime {
...
+
+                for (tool, tool_call) in tool_tasks {
+                    let tx = tx.clone();
+                    let tool_name = tool.desc().name.clone();
+                    let tool_call_id = tool_call
+                        .as_function()
+                        .and_then(|(id, _, _)| id)
+                        .map(|s| s.to_string());
+
+                    tokio::spawn(async move {
```
nuri-yoo, line 207: ### Tool panic during parallel execution permanently breaks the agent

When a tool function panics inside `tokio::spawn`, the panic is silently caught by tokio and the `tx` sender is dropped without sending any result. The drain loop finishes normally but with incomplete results, leaving history in an inconsistent state.

**Impact chain:**

1. Tool panics inside `tokio::spawn` → `tx` dropped without `send()`
2. Drain loop collects only the non-panicking tool's result
3. History becomes: `[assistant(tool_use_A, tool_use_B), tool_result_A]` — missing B
4. Next loop iteration calls LLM with invalid history → API returns 400
5. Agent is **permanently unusable** — every subsequent call (even unrelated ones) fails with the same 400 because the corrupted history is never cleaned up

**Reproduction:**

Register two parallel tools, one of which panics:

```rust
let bad_fn: Arc<crate::ToolAsyncFunc> = Arc::new(|_args| {
    Box::pin(async move {
        panic!("simulated tool crash");
    })
});
```

Test 1 — history inconsistency after panic:
```
assertion `left == right` failed:
  History is inconsistent: 2 tool_use call(s) but 1 tool result(s).
  The panicked tool's result was silently lost.
    left: 2
   right: 1
```

Test 2 — subsequent innocent query (`"What is 1 + 1?"`) also fails, confirming the agent is permanently broken.

<details>
<summary>Full test code</summary>

```rust
#[test_with::env(OPENAI_API_KEY)]
#[tokio::test]
async fn test_tool_panic_causes_inconsistent_history() {
    dotenvy::dotenv().ok();

    let api_key =
        std::env::var("OPENAI_API_KEY").expect("OPENAI_API_KEY must be set in .env");

    let good_desc = ToolDescBuilder::new("get_weather")
        .description("Get the current weather for a city")
        .parameters(to_value!({
            "type": "object",
            "properties": { "city": { "type": "string" } },
            "required": ["city"]
        }))
        .build();
    let good_fn: Arc<crate::ToolAsyncFunc> = Arc::new(|_args| {
        Box::pin(async move {
            tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
            crate::to_value!("sunny, 25 degrees")
        })
    });

    let bad_desc = ToolDescBuilder::new("get_traffic")
        .description("Get the current traffic conditions for a city")
        .parameters(to_value!({
            "type": "object",
            "properties": { "city": { "type": "string" } },
            "required": ["city"]
        }))
        .build();
    let bad_fn: Arc<crate::ToolAsyncFunc> = Arc::new(|_args| {
        Box::pin(async move {
            panic!("simulated tool crash");
        })
    });

    let mut tool_set = ToolSet::new();
    tool_set.insert(
        "get_weather".to_string(),
        ToolRuntime::new_async(good_desc, good_fn),
    );
    tool_set.insert(
        "get_traffic".to_string(),
        ToolRuntime::new_async(bad_desc, bad_fn),
    );

    let mut agent = AgentRuntime::new(
        AgentSpec::new("gpt-4o-mini")
            .with_tools(["get_weather", "get_traffic"])
            .with_instruction(
                "When asked about a city, ALWAYS call both get_weather AND \
                 get_traffic tools in a single response. Never call just one."
                    .to_string(),
            ),
        AgentProvider {
            lm: LangModelProvider::openai(api_key),
            tools: vec![],
        },
        tool_set,
    )
    .await
    .unwrap();

    let query = Message::new(Role::User).with_contents([Part::text(
        "Tell me about Seoul. Use get_weather for weather and get_traffic for traffic.",
    )]);

    {
        let mut strm = agent.stream_turn(query);
        while let Some(result) = strm.next().await {
            let _ = result;
        }
    }

    let history = agent.get_history();

    let tool_use_count: usize = history
        .iter()
        .filter(|m| m.role == Role::Assistant)
        .filter_map(|m| m.tool_calls.as_ref())
        .map(|tc| tc.len())
        .sum();

    let tool_result_count = history.iter().filter(|m| m.role == Role::Tool).count();

    if tool_use_count < 2 {
        eprintln!(
            "LLM only produced {} tool call(s); skipping consistency check",
            tool_use_count
        );
        return;
    }

    assert_eq!(
        tool_use_count, tool_result_count,
        "History is inconsistent: {} tool_use call(s) but {} tool result(s). \
         The panicked tool's result was silently lost.",
        tool_use_count, tool_result_count,
    );
}
```
</details>

**Suggested fix:**

The codebase already has the right pattern for streaming tools that never yield `Result` (synthetic error insertion at line 235-248). The same approach can be applied to `tokio::spawn` by either:

- Wrapping the spawn body with `std::panic::AssertUnwindSafe` + `catch_unwind` and sending a synthetic error result on panic
- Or tracking `JoinHandle`s and checking for `JoinError` after all tasks complete

This would keep history consistent and let the LLM see the error and recover on its own.

khj809: @nuri-yoo Thanks for catching the edge case. I applied your suggested solution and verified your test code passes.
For suppressing the traceback logs on panic, I added a simple macro `suppress_panics` on `src/macros.rs`.

nuri-yoo: LGTM

nuri-yoo (approved):

ljhh-0611 (approved):
Sorry for the late review :(
Looks Good To Me Too!

PR #379, opened 2026-04-14
fix: regressions in web search engines

khj809:
- bing: added `nfpr=1` to avoid query reformulation
- google: updated UAs and added `NSTNWV` suffix to let Google return the server-rendered HTML used by Google Go android app

PR #380, opened 2026-04-17
fix: improve Google block detection in web search

nuri-yoo:
## Summary

Tighten the bot-detection checks in the Google search engine so that the rate-limit / CAPTCHA pages Google actually serves today are detected cleanly, and transient HTTP errors are no longer collapsed into `SearchError::Blocked`.

## Context

While reviewing the UA update in [`52ac176`](https://github.com/brekkylab/ailoy/pull/379/commits/52ac17637e946cf844a56050a2098f9fbac16923) (PR #379) I probed the current block flow from a residential IP. Findings:

- The real rate-limit response is `HTTP 429` with a redirect to `https://www.google.com/sorry/index?continue=...` (terminal host is `www.google.com`, not `sorry.google.com`).
- The body contains `Our systems have detected unusual traffic`, `/sorry/index`, `id="captcha"`, and `CaptchaRedirect` — no occurrence of the literal `sorry.google.com` hostname.
- The existing `html_text.contains("sorry.google.com")` check is therefore effectively dead code against the modern flow; detection is currently carried by `/sorry/` (body substring) and the `status == 429` check.
- `if !response.status().is_success() { return Err(Blocked) }` misclassifies transient 5xx / unrelated 4xx as blocks, masking flaky-network failures.

## Changes

- Replace body substring `sorry.google.com` with a final-URL path check: `response.url().path().starts_with("/sorry/")` is the precise redirect signal and cannot false-positive on result snippets.
- Extract `is_captcha_page()` covering the soft-block case (status 200 with CAPTCHA body). Markers: the visible English copy plus two DOM-level anchors (`id="captcha"`, `CaptchaRedirect`) that survive minor wording changes.
- Separate transient HTTP failures from blocks: non-success statuses that are not 429 and not `/sorry/` now surface as `SearchError::Http` via `error_for_status()`, letting callers distinguish a blocked engine from a flaky one.
- Use `reqwest::StatusCode::TOO_MANY_REQUESTS` instead of the bare `429` literal.
- Add `log::warn!` at each block path with a distinct reason string (HTTP 429 / redirect URL / body marker) for observability.
- Drop the always-true `error_for_status()?` that was chained after the `is_success()` guard.

## Test plan

- [x] `cargo test -p ailoy --lib agent::rt::tool::web_search::engines::google` — 12/12 passing, including 6 new unit tests for `is_captcha_page` (4 positive markers, 2 negative cases: normal results that mention "sorry", and the legacy bare `sorry.google.com` string).
- [x] `cargo test ... test_search_returns_results -- --ignored --nocapture` against a blocked IP: confirmed the new path returns `SearchError::Blocked` (observed via the test's graceful skip branch).
- [x] End-to-end verification with the new UAs + `NSTNWV` flow on a cold IP: Block response was detected correctly by the new logic.

## Dependency

Builds on the UA / `NSTNWV` flow introduced in #379 (merged).

### DIFF
--- a/src/agent/rt/tool/web_search/engines/google.rs
+++ b/src/agent/rt/tool/web_search/engines/google.rs
@@ -25,6 +25,21 @@ static MOBILE_UAS: &[&str] = &[
     "Mozilla/5.0 (Linux; Android 8.0; Pixel 2 Build/OPD3.170816.012) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.9357.1059 Mobile Safari/537.36",
 ];
 
+/// Detects Google's CAPTCHA / "unusual traffic" challenge page by body markers.
+///
+/// Used for the soft-block case where the HTTP status is 200 but the body is the
+/// challenge page. Markers cover the visible English copy plus two DOM-level
+/// anchors that survive minor wording changes.
+fn is_captcha_page(html: &str) -> bool {
+    const BLOCK_MARKERS: &[&str] = &[
+        "Our systems have detected unusual traffic",
+        "/sorry/index",
+        "id=\"captcha\"",
+        "CaptchaRedirect",
+    ];
+    BLOCK_MARKERS.iter().any(|m| html.contains(m))
+}
+
 /// Pick a random base UA and append the `NSTNWV` suffix.
 ///
 /// `NSTNWV` is the token that identifies the "Google Go" native Android app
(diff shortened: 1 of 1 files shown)


src/agent/rt/tool/web_search/engines/google.rs
```diff
@@ -326,6 +349,53 @@ mod tests {
...
+        assert!(!is_captcha_page(body));
+    }
+
+    #[test]
+    fn test_is_captcha_page_rejects_legacy_sorry_google_com_string_alone() {
+        // The literal hostname "sorry.google.com" no longer appears in modern block
+        // responses; a body merely mentioning it should not be flagged by itself.
+        let body = r#"<html><body>See sorry.google.com for history.</body></html>"#;
+        assert!(!is_captcha_page(body));
+    }
```
khj809: I think this test is not necessary if `sorry.google.com` is entirely deprecated.

khj809 (approved):
Thanks for the updates!

src/agent/rt/tool/web_search/engines/google.rs
nuri-yoo: Good point — removed in a39276c.

PR #383, opened 2026-04-19
refactor: move max_tokens from LangModelInferConfig into LangModelPro…

jhlee525:
Remove the separate `LangModelInferConfig` struct and fold `max_tokens` directly into `LangModelProvider::API`. This eliminates an extra parameter from `LangModelRuntime::run()` and makes inference config part of the provider declaration rather than a per-call argument.


grf53 (approved):
LGTM

PR #384, opened 2026-04-20
Refactor: Core library restructure and API simplification

jhlee525:
# Summary

* Module layout: Split monolithic src/model/, src/tool/base.rs, and src/agent/base.rs into focused modules — lang_model/, lang_model_impl/, tool/, tool_impl/, agent/ — with *_impl crates holding concrete provider logic separate from public interfaces
* Naming cleanup: Removed *Runtime suffix from types; renamed stream_turn → run; simplified SubAgent enum to tuple-style variant
* AgentSpec as source of truth: Moved subagent declarations from AgentProvider/ToolProvider into AgentSpec, so agent topology is fully described in one place
* ToolSet improvements: Added merge, iter, and IntoIterator support; added default provider; removed the rt feature flag
try_with_tools API: Made async; now accepts IntoIterator and &AgentProvider
* Scope reduction: Removed bindings (Node.js, Python, Web), docs site, examples, cache, CLI, FFI, knowledge base, and vector store — slimming the repo to the core Rust library

# Test plan
* cargo build passes on the refactoring branch
* Doctests in src/agent/spec.rs and src/tool/ pass
* Verify AgentSpec builder pattern (model → instruction → tools → subagents) works end-to-end
* Confirm removed modules (cache, knowledge, vector_store) have no lingering references in lib.rs

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


jhlee525:
diagram-export-2026 -4 -20 -오후-2_44_54

src/agent/provider.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Register all OpenAI models under the `openai/*` namespace via the Responses API.
+    ///
+    /// Use model names like `"openai/gpt-4o"` or `"openai/o3-mini"` in [`AgentSpec`].
+    pub fn model_openai(&mut self, api_key: impl Into<String>) -> &mut Self {
+        self.models.insert(
+            "openai/*".to_string(),
+            LangModelProvider::API {
+                schema: LangModelAPISchema::ChatCompletion,
+                url: Url::parse("https://api.openai.com/v1/chat/completions").unwrap(),
```
khj809: The comment says about "Responses API", but the endpoint and schema is for chat completion API.

```suggestion
            LangModelProvider::API {
                schema: LangModelAPISchema::OpenAI,
                url: Url::parse("https://api.openai.com/v1/responses").unwrap(),
```

src/lang_model_impl/api/anthropic.rs
```diff
@@ -410,21 +396,22 @@ mod tests {
         let api_key =
             std::env::var("ANTHROPIC_API_KEY").expect("ANTHROPIC_API_KEY must be set in .env");
 
-        let lm = LangModelRuntime::new(
+        let lm = LangModel::new(
```
khj809: If we decided to use `model` instead of `lm`, let's replace every `lm`s at once.
```suggestion
        let model = LangModel::new(
```

src/lib.rs
```diff
@@ -3,18 +3,10 @@ extern crate self as ailoy;
 
 pub mod agent;
 pub mod datatype;
+pub mod lang_model;
+mod lang_model_impl;
```
khj809: In order to align with `tool_impl`:
```suggestion
pub(crate) mod lang_model_impl;
```

src/tool/toolset.rs
```diff
@@ -0,0 +1,179 @@
...
+                            python_version: python_version.clone(),
+                            venv_path: venv_path.clone(),
+                            packages: packages.clone(),
+                        })
+                        .await?;
+                        this.tools.insert(
+                            "python_repl".into(),
+                            (tool_runtime.get_desc().clone(), tool_runtime.get_func()),
+                        );
+                    }
```
khj809: I'd like to suggest moving this under `tool_impl/builtins/mod.rs`, and define `make_builtin_tool(provider)` to be used here, in order to clearly separate the responsibility.

src/tool/toolset.rs
```diff
@@ -0,0 +1,179 @@
...
+        // Initialise sub-agents using the toolset built above.
+        for sub_spec in &spec.subagents {
+            let card = sub_spec.card.clone().ok_or_else(|| {
+                anyhow::anyhow!(
+                    "SubAgent spec must have a card (name + description) to be \
+                     registered as a tool"
+                )
+            })?;
+            let tool_name = card.name.clone();
+            let sub_agent = Agent::try_with_tools(sub_spec.clone(), provider, &this).await?;
```
khj809, line 66: I'm quite worried about using the parent agent's provider and ToolSet for creating subagents. For example, users may want to register subagents which uses different API keys or different tool functions, but there's no way to control that.

src/agent/rt.rs
```diff
@@ -0,0 +1,783 @@
...
+
+        // Initialize history with system instruction if present
+        let history = spec
+            .instruction
+            .as_ref()
+            .map(|inst| vec![Message::new(Role::System).with_contents([Part::text(inst)])])
+            .unwrap_or_default();
+
+        Ok(Self {
+            lm: LangModel::new(spec.model.clone(), lm_provider),
```
khj809: Every agent tests are failing due to the prefixed model names (e.g. `openai/gpt-4o-mini`).
Need to provide the exact model id part:
```suggestion
        let model_id = spec
            .model
            .split_once('/')
            .map(|(_, id)| id.to_string())
            .unwrap_or_else(|| spec.model.clone());
            
        Ok(Self {
            lm: LangModel::new(model_id, lm_provider),
```

src/agent/rt.rs
```diff
@@ -0,0 +1,783 @@
...
+
+                    match last {
+                        Some(mut item) => {
+                            item.depth = None;
+                            item.message.role = Role::Tool;
+                            self.state.history.push(item.message.clone());
+                            yield item;
+                        }
+                        None => Err(anyhow::anyhow!("tool '{}' produced no output", tool_name))?,
+                    }
```
khj809: Current tool execution cannot handle panics occurred inside tool function, which makes `test_tool_panic_causes_inconsistent_history` test to be failed.
If you're okay, I'll commit the fixed codes soon.

jhlee525: All right, thank you for check

src/tool/toolset.rs
jhlee525: I think subagent execution is part of the whole agent execution - at least for in-process subagents.

### Different API keys 

I understand why that this might come up, but I think subagents are the part of the whole agent, not independent agent.

It's like calling one function from another within the same program: you wouldn't consider them separate programs just because they're different functions.

### Different tool functions

The provider doesn't define tools; it only defines how they're executed (the operational behavior).

This means, changing provider should not change how the tool behaves.

So if you need different tool behavior per subagent, it would call for a new abstraction like `ToolSpec`, not a per-subagent provider override.

---

I understand the concern, but I can't think of a clean solution that provides that level of flexibility without breaking the separation between provider-level configuration (e.g. API keys) and the agent's logical definition.

I think we have to collect more concrete requirements and think about how to handle this problem.

khj809 (approved):

src/lang_model_impl/api/gemini.rs
nuri-yoo, line 69: Gemini's API doesn't accept HTTP image URLs (only `inline_data` / Files API), so rejecting this input is correct — the issue is that `panic!` aborts the current thread instead of surfacing as a recoverable error. Callers can't  catch it with `Result` handling.        
                                                                                                                        
Minimum fix that keeps this PR's scope: replace the `panic!` with a returned error `Value`, e.g. `return to_value!({ "error": "Gemini does not support image URL inputs; provide a base64 data URI instead" });`

jhlee525: Thank you for comments!

nuri-yoo (approved):

PR #387, opened 2026-04-21
feat: Introduce `Sandbox` to run tools inside sandboxed environment

khj809:
## Summary

Introduces a `Sandbox` abstraction backed by [microsandbox](https://github.com/microsandbox/microsandbox) (libkrun-based MicroVM) and wires it through the tool execution pipeline, so builtin tools can run either on the host or inside an isolated VM with no change to tool code.

### `src/sandbox/` — new `Sandbox` type (`feature = "sandbox"`)

- Thin wrapper around `microsandbox::Sandbox`, keeping the public API decoupled from the underlying library.
- Each `Sandbox` instance corresponds to a uniquely named (`ailoy-<uuid>`) MicroVM instance. VM state (files written before `stop`) survives a `stop`/`start` cycle, so state persists across tool calls within the same instance.
- Configurable via `SandboxConfig`: OCI image, volume mounts (`VolumeMount`), port forwards, memory/CPU limits.
- Key methods: `start` / `stop`, `shell` / `shell_with_timeout`, `write_file` / `read_file` / `read_file_bytes`.

### `src/tool/` — `ToolContext` + sandbox-aware `ToolFunc`

- Added `ToolContext { sandbox: Option<Arc<Sandbox>> }` — the agent constructs one per turn and forwards it to every tool call.
- `ToolFunc` now accepts `ToolContext` (carrying the optional sandbox) alongside the tool arguments.
- Tools that don't need the sandbox simply ignore it; tools that do branch on `sandbox.is_some()`.

### `src/agent/` — sandbox lifecycle management

- `AgentProvider` gains an optional `SandboxConfig`; when the `sandbox` feature is enabled, `Agent::try_with_tools` creates a `Sandbox`, stops it immediately after init (VM only runs during tool calls), and attaches it to `AgentState`.
- The agent passes `state.sandbox.clone()` into `ToolContext` for every tool invocation.

### `src/tool_impl/builtins/` — sandbox-capable builtin tools

**`bash`** (new)
- Runs arbitrary shell commands via `sh -c`.
- On the host: `tokio::process::Command`. In a sandbox: `Sandbox::shell`.

**`python_repl`** (revised)
- Replaced the previous `PythonEnv`/`SourceTool`/`uv.rs` stack with a single `PythonScriptRunner`.
- One-time setup (`SETUP_CMD`) bootstraps python3 + uv + a venv under `$XDG_CACHE_HOME/ailoy/` — works on bare images (no curl/wget; uses Python `urllib`/`tarfile`).
- Supports `pip_install` before execution. In sandbox mode, venv state persists across calls within the same instance.

**`convert_pdf_to_md`** (removed)
- Removed as a builtin tool. The intended behaviour is instead covered by a new test case (`test_convert_pdf_to_md_skill`) that validates the same functionalities via the Agent Skills mechanism:
  1. Before the agent runs, the SKILL.md and a test PDF are written into the sandbox at known paths.
  2. The agent's system instruction lists available skills by name, description, and SKILL.md path only — no full content upfront (progressive disclosure).
  3. The agent activates the skill by reading the SKILL.md via bash tool, then installs the required dependency (Docling) and performs the PDF-to-Markdown conversion.

### `src/shell/` — removed

The in-memory shell simulator (`src/shell/`) is removed; real execution is now handled by `bash` (host) or `Sandbox::shell` (VM).

## Tests

- [x] `cargo test` — all non-`#[ignore]` tests pass
- [x] `cargo test --features sandbox` — sandbox unit tests pass (requires microsandbox runtime)
- [x] `#[ignore]` integration tests (`test_convert_pdf_to_md_skill`, matplotlib plot test) pass manually
- [x] `cargo check --features sandbox` — no warnings


src/agent/rt.rs
```diff
@@ -177,10 +163,33 @@ impl Agent {
...
+            use crate::sandbox::{Sandbox, SandboxConfig};
+            let config = provider
+                .sandbox_config
+                .clone()
+                .unwrap_or_else(SandboxConfig::default);
+            let sandbox = Arc::new(
+                Sandbox::new(config)
+                    .await
+                    .expect("Failed to initialize sandbox"),
+            );
```
ljhh-0611, line 182: In my opinion,
If a Provider has a sandboxconfig, Agents with the Provider will have a sandbox with the same names, not the fresh_sandbox_name in each agent creation.
Every names will be different in case of absence of sandboxconfig in provider(e.g. default provider?)

I thought it with agent-k, agent-k has each provider for a user, probably the provider set the sandbox configuration- so names of sandbox in agents will be duplicated

Am I right..?

ljhh-0611: So.. how can Config remove name, and Sandbox use fresh_sandbox_name function for sandbox name in new function?

khj809: Thanks for pointing out the issue!
As you mentioned, since AgentProvider includes SandboxConfig instance which already has a fixed `name`, it will keep creating sandboxes with the same name unintentionally.
The solution would be providing an interface to override AgentState during the agent creation (e.g. `.with_state(mut self, state: AgentState)`, so users can either provide their own initialized AgentState (with Sandbox having configured `name`) or use the default one.
AFAIK, agent-k already does sth like that to restore session-wise data(e.g. message history), so it makes sense to store `SandboxConfig` along with those data.

jhlee525 (approved):

jhlee525:
Good! Thank you

ljhh-0611 (approved):
👍

PR #389, opened 2026-04-23
introducing tool factory

jhlee525:
## Summary

- Introduces `ToolFactory` as the unit stored in `ToolSet`, replacing the old `(ToolDesc, Arc<ToolFunc>)` pair. `ToolSet::insert` now takes `(key, ToolFactory)` and `ToolFactory::simple` / `ToolFactory::sandbox_aware` cover the common construction patterns.
- Adds `AgentSpec::sandbox` field (a named key) and changes `AgentProvider::sandbox_config` → `AgentProvider::sandboxes: BTreeMap<String, SandboxConfig>`, so multiple named sandboxes can be registered and looked up by the spec at runtime.
- Fixes doctest in `Agent::try_with_tools` that used the old three-argument `insert` signature and referenced an unimported `ToolContext`.


jhlee525:
This PR includes breaking changes, so it doesn’t need to be merged urgently.

I see it as an example where 1. `AgentSpec` determines whether a sandbox is used 2. and the behavior of tools is defined accordingly based on that sandbox setting.

src/agent/rt.rs
```diff
@@ -170,22 +172,25 @@ impl Agent {
...
-            );
-            // Stop immediately after setup — VM only runs during tool execution.
-            sandbox
-                .stop()
-                .await
-                .expect("Failed to stop sandbox after init");
-            state.sandbox = Some(sandbox);
+                    .expect("Failed to stop sandbox after init");
+                state.sandbox = Some(sandbox);
+            }
```
khj809: When the provider does not have matching SandboxConfig with `spec.sandbox`, it does not raise any error and agent is created without sandbox. However, the `ToolFactory::sandbox_aware` would select the function using sandbox since AgentSpec has `sandbox` key, so the tool function panics on getting sandbox from `ctx.sandbox` which is None.
Therefore, rather than panic during tool execution, I suggest to return Err if `provider.sandboxes` does not have `spec.sandbox` when it's not `None`.
```suggestion
            if let Some(key) = spec.sandbox {
                if let Some(config) = provider.sandboxes.get(&key).cloned() {
                    let sandbox = Arc::new(
                        Sandbox::new(config)
                            .await
                            .expect("Failed to initialize sandbox"),
                    );
                    // Stop immediately after setup — VM only runs during tool execution.
                    sandbox
                        .stop()
                        .await
                        .expect("Failed to stop sandbox after init");
                    state.sandbox = Some(sandbox);
                } else {
                    return Err(anyhow::anyhow!("Sandbox '{}' not registered", key));
                }
            }
```

jhlee525: This occurred because I didn’t directly review the code generated by the AI 😓

I think it would be better to proceed this way.

src/agent/rt.rs
khj809: Now every sandbox tests should have sandbox configuration.

```suggestion
        let mut provider = AgentProvider::new();
        provider
            .model_claude(api_key)
            .tool(ToolProvider::Builtin(BuiltinToolProvider::PythonRepl {}))
            .sandbox("my-sandbox", SandboxConfig::default());

        let spec = AgentSpec::new("anthropic/claude-haiku-4-5-20251001")
            .tool("python_repl")
            .instruction(
                "When asked to run Python code, always use the python_repl tool. \
                 Never skip the tool call.",
            )
            .sandbox("my-sandbox");
```

khj809: ditto

khj809: ditto

src/tool/rt.rs
```diff
@@ -3,11 +3,74 @@ use std::sync::Arc;
...
+    pub fn simple(desc: ToolDesc, f: ToolFunc) -> Self {
+        let f = Arc::new(f);
+        Self::new(
+            desc.name.clone(),
+            Arc::new(move |_| (desc.clone(), f.clone())),
+        )
+    }
+
+    // Sandbox-aware, if agent have sandbox: f1, or f2
+    pub fn sandbox_aware(desc: ToolDesc, f1: ToolFunc, f2: ToolFunc) -> Self {
```
khj809: The argument name is confusing. Suggest to rename as below:
```suggestion
    pub fn sandbox_aware(desc: ToolDesc, f_sandbox: ToolFunc, f_local: ToolFunc) -> Self {
```

khj809 (approved):
LGTM after resolving conflicts

PR #390, opened 2026-04-24
Sandbox feature revised

jhlee525:
- Gate all sandbox-related imports, struct fields, and call-sites behind `#[cfg(feature = "sandbox")]`, replacing the previous no-op stub `Sandbox` type. The `sandbox` module itself is now only compiled when the feature is enabled.
- Introduce a Handle struct in `runner.rs` to encapsulate sandbox-vs-local dispatch inside `PythonScriptRunner,` eliminating the - `Option<&Arc<Sandbox>>` argument that was threaded through every method (`ensure_ready`, `install_packages`, `run`, `run_with_timeout`, `run_shell`, `write_file`).
- Extract a shared `run_python(runner, args)` free function in `python_repl/mod.rs` so `f_local` and `f_sandbox` reduce to one-liners that differ only in how they construct the runner.
- Apply the same `#[cfg(feature = "sandbox")]` split to `bash.rs` and `tool/func.rs` (`ToolContext::sandbox` field, `ToolContext::sandbox()` builder method).


src/tool_impl/builtins/python_repl/mod.rs
```diff
@@ -33,145 +31,84 @@ pub async fn build_python_repl_tool() -> anyhow::Result<ToolFactory> {
...
-                            "stdout": "",
-                            "stderr": format!("pip install error: {e}").as_str(),
-                            "exit_code": 1,
-                            "phase": "pip_install"
-                        });
-                    }
-                    Ok(_) => {}
-                }
-            }
+    let f_local = ToolFunc::new(|args: Value| run_python(PythonScriptRunner::new(), args));
```
khj809: This will make a fresh PythonScriptRunner whenever the tool is executed, which is not a desired behavior since `SETUP_CMD` will be called every time. I think it should take the initialized runner outside the closure, to prevent calling `SETUP_CMD` after its `initialized` become true.

src/tool_impl/builtins/python_repl/mod.rs
```diff
@@ -33,145 +31,84 @@ pub async fn build_python_repl_tool() -> anyhow::Result<ToolFactory> {
...
+    #[cfg(not(feature = "sandbox"))]
+    {
+        Ok(ToolFactory::simple(desc, f_local))
+    }
+
+    #[cfg(feature = "sandbox")]
+    {
+        let f_sandbox = ToolFunc::new(|args: Value, ctx: crate::tool::ToolContext| {
+            run_python(PythonScriptRunner::with_sandbox(ctx.sandbox), args)
+        });
```
khj809: Same as before, but more tricky since it requires sandbox at initialization time. I'd like to suggest to make PythonScriptRunner to be initialized without sandbox, and lazily run `SETUP_CMD` on tool execution time when `initialized` is false.

src/tool_impl/builtins/python_repl/runner.rs
```diff
@@ -3,7 +3,126 @@ use std::sync::Arc;
...
+
+/// Execution result from a shell command.  Mirrors `crate::sandbox::ExecResult` but is
+/// always compiled so non-sandbox builds don't need the sandbox module.
+#[derive(Debug)]
+pub struct ExecResult {
+    pub stdout: String,
+    pub stderr: String,
+    pub exit_code: i32,
+    pub timed_out: bool,
+}
```
khj809: Rather than mirroring this, how about creating a feature-independent mod to share this struct?

src/tool_impl/builtins/python_repl/runner.rs
```diff
@@ -3,7 +3,126 @@ use std::sync::Arc;
...
+    }
+
+    #[cfg(feature = "sandbox")]
+    async fn write_file_sandbox(&self, path: &str, contents: &[u8]) -> anyhow::Result<()> {
+        let sb = self.sandbox.as_ref().unwrap();
+        sb.write_file(path, contents).await
+    }
+
+    async fn write_file_local(&self, path: &str, contents: &[u8]) -> anyhow::Result<()> {
+        std::fs::write(path, contents).context("failed to write script to host")?;
```
khj809: This would block the async function. Suggest to replace with `tokio::fs::write`.

jhlee525:
Felt that the growing complexity in `bash` and `python_repl` called for a fundamental redesign.

I replaced `sandbox` with `runenv` as the universal execution environment, so all tools always go through it — eliminating the conditional gating logic entirely.

src/agent/rt.rs
```diff
@@ -167,25 +175,15 @@ impl Agent {
...
 
         #[allow(unused_mut)]
-        let mut state = AgentState::with_history(history);
+        let mut state = AgentState::new().history(history);
         #[cfg(feature = "sandbox")]
         {
-            use std::sync::Arc;
-
-            use crate::sandbox::Sandbox;
             if let Some(key) = spec.sandbox {
```
ljhh-0611: As I know, spec.sandbox is revised to spec.runenv

src/runenv/sandbox.rs
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    async fn write(&self, path: &Path, content: &[u8]) -> anyhow::Result<()> {
+        let path_str = path.to_string_lossy().into_owned();
+        let sb = self.acquire().await?;
+        sb.fs().write(&path_str, content).await?;
+        self.release(sb).await?;
+        Ok(())
+    }
+}
```
ljhh-0611, line 165: If fs().read/write in read/write function results Error, sandbox couldn't be released normally

src/tool_impl/builtins/python_repl/runner.rs
```diff
@@ -61,154 +60,95 @@ const AILOY_PYTHON: &str = r#""${XDG_CACHE_HOME:-$HOME/.cache}/ailoy/venv/bin/py
...
+            r.exit_code,
+            if r.stderr.trim().is_empty() {
+                r.stdout.trim()
+            } else {
+                r.stderr.trim()
+            }
+        );
     }
+    Ok(())
+}
```
ljhh-0611: Was there a specific issue with the previous initialized guard approach?

If the intention is for ensure_ready() to run on every invocation, would it make sense to strengthen the fast-path checks inside SETUP_CMD so already-initialized environments can return as cheaply as possible?

src/agent/spec.rs
```diff
@@ -40,10 +40,11 @@ pub struct AgentSpec {
...
     pub card: Option<AgentCard>,
 
-    /// Sandbox to run
-    /// If specified, it'll run in sandbox environment
+    /// Run environment.
+    /// If not specified, the tools will be run in local machine.
+    /// If specified, it'll use the run environment
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub sandbox: Option<String>,
+    pub runenv: Option<String>,
```
ljhh-0611: Would it be clearer to keep this field named `sandbox` for now?
I think that this name invoke users hesitation for using this field.
<runenv == None -> Local environment > is implicit in this context. Also any Some(_) value is treated as sandbox mode.

Or we have to use real enum type for this field, but in context of Serializing/Deserializing, An enum type adoption will be required, I think that is over engineering.

This is just my little opinion, so feel free to blame😁👍

src/tool/rt.rs
ljhh-0611: Could we add a small test for `ToolFactory::sandbox_aware` selection behavior?

It currently chooses the sandbox function based on `spec.runenv.is_some()`, which is exactly the boundary affected by this rename. A test for both `None` and `Some(_)` would help prevent this from regressing.

src/tool_impl/builtins/bash.rs
```diff
@@ -40,71 +37,27 @@ pub async fn build_bash_tool() -> anyhow::Result<ToolFactory> {
...
-                "stderr": r.stderr.as_str(),
-                "exit_code": r.exit_code as i64,
-                "timed_out": r.timed_out
-            }),
-            Err(e) => crate::to_value!({
-                "stdout": "",
-                "stderr": format!("execution error: {e}").as_str(),
+        let Ok(out) = ctx
+            .runenv
+            .exec("sh".to_string(), vec!["-c".to_string(), cmd], None)
```
ljhh-0611, line 43: The timeout_secs property is not being used. It is being silently ignored in the build bash tool.

src/runenv/sandbox.rs
```diff
@@ -0,0 +1,537 @@
...
+
+    /// Number of virtual CPUs. Default: `2`.
+    pub cpus: u8,
+
+    /// Guest memory in MiB. Default: `2048`.
+    pub memory_mib: u32,
+
+    /// Default working directory inside the sandbox. Default: `"/workspace"`.
+    /// Created automatically after boot if it does not already exist.
+    pub workdir: String,
```
ljhh-0611, line 66: As far as I can tell, workdir hasn't been working even before this PR.
We need to add code to pass workdir (as cwd for ExecOptionsBuilder) to the exec function, or if possible, in Sandbox::new function, switch default workspace directory.
Sandbox::new(config) only creates the workdir directory — it doesn't actually set it as the default working directory.

** This seems pre-existing from the old `src/sandbox/mod.rs`, so not necessarily a blocker for this PR. But since this PR moves the sandbox abstraction into `runenv`, it may be a good chance to clarify the contract. **

src/agent/rt.rs
jhlee525: it's a bug. I made changes, PTAL

src/runenv/sandbox.rs
jhlee525: Yes, that’s true. I tried various approaches to fix the sandbox lifecycle issue, but there wasn’t a simple solution without making significant changes. I’ll think further about how to resolve this problem.

src/tool_impl/builtins/python_repl/runner.rs
jhlee525: I thought if `runenv` itself provide syncronization protection between the callers, explicit guard in here is not necessary.

src/agent/spec.rs
jhlee525: As you may already know, with the recent changes, "sandbox" has become a trait that applies to both local and sandboxed environments, from the sandbox would only apply to non-local environments.

The idea in here is to specifying a “running environment key” here, and then let the provider define how to execute based on the given key.

However, this concept likely needs further discussion.

src/runenv/sandbox.rs
```diff
@@ -0,0 +1,537 @@
...
+                timed_out: false,
+            }),
+        }
+    }
+
+    async fn read(&self, path: &Path) -> anyhow::Result<Vec<u8>> {
+        let path = path.to_string_lossy().into_owned();
+        let sb = self.acquire().await?;
+        let rv = sb.fs().read(&path).await?.to_vec();
+        self.release(sb).await?;
```
khj809, line 154: When the L153 fails, it will be terminated without calling `self.release(sb)`, which leads to every future `self.acquire()` to be deadlocked.
But I'll handle it on #391, so let's merge it first.

khj809 (approved):

PR #391, opened 2026-04-24
feat: Sandbox Lifetime/Sharing Refactor + AgentBuilder

khj809:
## Background

The goal is to let the agent-k backend attach a microsandbox per session and have a parent chat agent and its subagents **share the same `Arc<Sandbox>`**. This PR fixes three defects in the existing ailoy API.

1. **No public API to clean up a `persist=true` sandbox from outside.** `Sandbox::shutdown()` skips `remove_persisted` when `persist=true`, and `Drop` does the same. A session DELETE leaves the sandbox as an orphan.

2. **`Agent::run`'s batch envelope causes nested lifetime contention.** `run` wraps each tool-call batch with `sb.start() / sb.stop()`. When a subagent tool recursively calls its own `Agent::run`, the inner `stop()` kills the outer sandbox, and the remaining ops in the outer batch fail with "sandbox is not running".

3. **No build-time path for parent and subagent to selectively share a sandbox.** `ToolSet::from_providers` calls `Sandbox::new(config)` for each subagent, creating a separate `Arc<Sandbox>` with its own `Mutex`. Callers need a per-agent opt-in to choose between "share an existing `Arc`" and "create a new sandbox from a config".

> **Note:** This branch was rebased onto PR #390 ("Sandbox feature revised"), which introduced the `RunEnv` trait and the `src/runenv/` module. The implementation here builds on that foundation.

---

### Change A — `pub async fn remove_persisted(name: &str)`

The only public API for cleaning up a `persist=true` sandbox from outside the sandbox. The cleanup logic previously buried in `Drop` is extracted into a free function in `src/runenv/sandbox.rs`.

- Calling with an unregistered name returns `Ok(())` (idempotent)
- If running/draining: `stop_and_wait()` then `remove_persisted()`
- If stopped: removes disk state without restarting the sandbox

At session teardown: `ailoy::runenv::remove_persisted(&format!("session-{sid}")).await?` reclaims the sandbox cleanly.

**Tests added:**
- `test_remove_persisted_idempotent` — unknown name still returns `Ok`
- `test_remove_persisted_cleans_up` — create `persist=true` sandbox → write file → drop → `remove_persisted` → recreate with same name → file is gone

### Change B — `RunEnv` trait + per-op sandbox lifetime (eliminates nested contention)

PR #390 introduced a `RunEnv` trait (`src/runenv/mod.rs`) that abstracts over local and sandbox execution:

```rust
pub trait RunEnv: Send + Sync {
    async fn exec(&self, program: String, args: Vec<String>, timeout: Option<u64>) -> anyhow::Result<ExecResult>;
    async fn read(&self, path: &Path) -> anyhow::Result<Vec<u8>>;
    async fn write(&self, path: &Path, content: &[u8]) -> anyhow::Result<()>;
}
```

Two implementations live in `src/runenv/`:
- **`Local`** — executes directly on the host, no sandbox involved.
- **`Sandbox`** — each op lazy-starts the sandbox, executes, then stops. `AgentState.sandbox: Option<Arc<Sandbox>>` is replaced by `AgentState.runenv: Arc<dyn RunEnv>`, which defaults to `Arc::new(Local {})` for agents without a sandbox.

All tool functions use `ctx.runenv` uniformly — no `with_sandbox: bool` toggle is needed.

**`Agent::run` no longer manages sandbox lifetime at all.** The batch `start/stop` envelope is removed. The nested-call race where an inner `stop()` kills the outer sandbox is eliminated at the root.

**Bug fix — error-path sandbox leak in file I/O methods (`src/runenv/sandbox.rs`):**

All 7 file I/O methods (`read`, `write`, `write_file`, `read_file`, `read_file_bytes`, `copy_from_host`, `copy_to_host`) previously used `?` before calling `stop_and_wait()`, leaving the sandbox running on failure. Fixed by storing the result first and calling `stop_and_wait()` unconditionally:

```rust
// before (sandbox leaks on error)
let bytes = guard.fs().read(&guest_path).await?.to_vec();
let _ = guard.stop_and_wait().await;

// after
let result = guard.fs().read(&guest_path).await.map(|b| b.to_vec());
let _ = guard.stop_and_wait().await;
Ok(result?)
```

`exec`, `shell`, and `shell_with_timeout` already used the correct pattern.

Additional changes to `SandboxConfig`:
- `name` changed from `String` to `Option<String>`. `None` defers name generation to `Sandbox::new` time via UUID, so every deserialized config automatically gets a fresh unique name.

### Change C — `AgentBuilder`

**🚧 This implementation needs further architecture design discussions. Currently its main purpose is to verify the scenario where parent and subagent share the same sandbox.**

#### `AgentBuilder` (`src/agent/builder.rs`) — new

A spec-free construction path for `Agent` that takes only materialised runtime objects, bypassing `AgentSpec`, `AgentProvider`, and `ToolFactory` entirely.

```rust
use std::sync::Arc;
use ailoy::{
    agent::{AgentBuilder, AgentCard},
    runenv::{Sandbox, SandboxConfig},
    tool::{BuiltinToolProvider, make_builtin_tool},
};

// Sharing a sandbox between parent and subagent
let sandbox = Arc::new(Sandbox::new(SandboxConfig {
    persist: true,
    name: Some(format!("session-{sid}")),
    ..Default::default()
}).await?);

let sub_agent = AgentBuilder::new(sub_model)
    .instruction("...")
    .tool(make_builtin_tool(&BuiltinToolProvider::Bash {}).await?)
    .runenv(sandbox.clone())
    .build().await?;

let parent = AgentBuilder::new(parent_model)
    .instruction("...")
    .tool(make_builtin_tool(&BuiltinToolProvider::Bash {}).await?)
    .runenv(sandbox.clone())       // same Arc — shared sandbox
    .subagent(card, sub_agent)
    .build().await?;

// At session teardown
drop(parent);
ailoy::runenv::remove_persisted(&format!("session-{sid}")).await?;
```

Key design decisions:
- Does not accept `ToolFactory`. Callers who need a factory-derived tool call `factory.make(&spec)` first to get a `Tool`.
- `.subagent(card, agent)` takes an already-built `Agent`. Sandbox sharing is achieved simply by passing the same `Arc` to both builders.
- The existing `Agent::try_new` / `try_with_provider` / `try_with_tools` paths are **completely unchanged**. `AgentBuilder` is an independent parallel path, not a replacement.

#### `Agent::from_parts` (`src/agent/rt.rs`) — crate-internal

An internal assembly constructor used exclusively by `AgentBuilder::build()`. Not exposed as public API.

#### Public tool builder functions (`src/tool/builder.rs`)

`src/tool/builder.rs` exposes two convenience constructors under `ailoy::tool::*`:

```rust
pub async fn make_builtin_tool(provider: &BuiltinToolProvider) -> anyhow::Result<Tool>
pub async fn make_a2a_tool(url: Url) -> anyhow::Result<Tool>
```

These are thin wrappers around the internal `ToolFactory`-returning functions and are intended for the `AgentBuilder` path (where you need a `Tool`, not a `ToolFactory`).

#### Split `make_subagent_tool` (`src/tool_impl/subagent.rs`)

Naming convention: **short name = returns `Tool`, `_factory` suffix = returns `ToolFactory`**.

| Function | Returns | Use case |
|----------|---------|----------|
| `make_subagent_tool(card, agent)` | `Tool` | `AgentBuilder` path — registers an already-built `Agent` as a tool |
| `make_subagent_tool_factory(card, agent)` | `ToolFactory` | spec/provider path — used by `ToolSet::from_providers` |

The shared desc/func assembly logic is extracted into a `make_subagent_parts` helper. The call site in `ToolSet::from_providers` is updated to `make_subagent_tool_factory`.

---

## Public API Changes

| Item | Change |
|------|--------|
| `ailoy::runenv::remove_persisted(name)` | **new** free function |
| `ailoy::runenv::RunEnv` | **new** trait (from PR #390). `exec`, `read`, `write` |
| `ailoy::runenv::Local` | **new** (from PR #390). Host-native `RunEnv` impl, default for non-sandbox agents |
| `AgentState.sandbox: Option<Arc<Sandbox>>` | → `runenv: Arc<dyn RunEnv>` (from PR #390). Always present, defaults to `Local` |
| `SandboxConfig::name` | `String` → `Option<String>` |
| `Sandbox` file I/O methods | Error-path sandbox leak fixed: `stop_and_wait()` now called on both success and failure |
| `ailoy::agent::AgentBuilder` | **new**. `new(model)` → `.instruction / .tool / .tools / .subagent / .runenv / .build` |
| `Agent::from_parts` | **new** `pub(crate)` constructor. Builder-only |
| `make_subagent_tool(card, agent)` | **return type changed**: `ToolFactory` → `Tool` |
| `make_subagent_tool_factory(card, agent)` | **new** (carries the old `make_subagent_tool` ToolFactory-returning behaviour) |
| `make_a2a_tool(url)` | **return type changed**: `Result<ToolFactory>` → `Result<Tool>` |
| `make_a2a_tool_factory(url)` | **new** (carries the old `make_a2a_tool` ToolFactory-returning behaviour) |
| `build_bash_tool()` | Now `async`, returns `anyhow::Result<ToolFactory>`. No `with_sandbox` param — all execution goes through `ctx.runenv` |
| `build_python_repl_tool()`, `build_web_search_tool()` | Same pattern as above |
| `ailoy::tool::make_builtin_tool(provider)` | **new public path** → `anyhow::Result<Tool>` |
| `ailoy::tool::make_a2a_tool(url)` | **new public path** → `anyhow::Result<Tool>` |
| `Agent::run` sandbox envelope | Removed (no observable API impact) |
| `Agent::{try_new, try_with_provider, try_with_tools}` | **Unchanged** |
| `ToolSet::from_providers` and all other methods | **Unchanged** |


src/agent/builder.rs
```diff
@@ -0,0 +1,313 @@
...
+///
+/// Use [`crate::agent::Agent::try_with_provider`] / [`Agent::try_new`] when you
+/// want the spec/provider path (YAML-friendly, string-keyed).
+pub struct AgentBuilder {
+    model: LangModel,
+    instruction: Option<String>,
+    tools: Vec<Tool>,
+    #[cfg(feature = "sandbox")]
+    sandbox: Option<SandboxSource>,
+}
```
jhlee525: Rather than having the `LangModel` and `Tool` here, it seems much simpler to use an AgentSpec.

The actual runtime construction would then be handled by AgentBuilder::build().

src/agent/builder.rs
```diff
@@ -0,0 +1,313 @@
...
+
+/// How the sandbox is supplied to an [`AgentBuilder`].
+///
+/// Pass an `Arc<Sandbox>` to share an existing VM across agents, or a
+/// `SandboxConfig` to create a dedicated VM during [`AgentBuilder::build`].
+#[cfg(feature = "sandbox")]
+pub enum SandboxSource {
+    Shared(Arc<dyn RunEnv>),
+    Fresh(SandboxConfig),
+}
```
jhlee525: Why do we need this enum?

within the builder function

```
fn sandbox_config(&self, config: SansboxConfig) {
  self.runenv = Arc::new(Sandbox::from_config())
}

fn sandbox(&self, sb: Arc<Sandbox>) {
  self.runenv = Sandbox::from_config()
}
```

Will takes same effect

src/agent/builder.rs
```diff
@@ -0,0 +1,313 @@
...
+/// runtime objects (e.g. from an existing session or an external DI container).
+///
+/// Use [`crate::agent::Agent::try_with_provider`] / [`Agent::try_new`] when you
+/// want the spec/provider path (YAML-friendly, string-keyed).
+pub struct AgentBuilder {
+    model: LangModel,
+    instruction: Option<String>,
+    tools: Vec<Tool>,
+    #[cfg(feature = "sandbox")]
+    sandbox: Option<SandboxSource>,
```
jhlee525: cfg gating and using "sandbox" will make unnecessary complexity.

I think using `runenv` here will be more straightforward.

khj809: I removed the feature gate and let builder receive `Arc<dyn RunEnv>` directly.

src/agent/builder.rs
khj809: Now it's unified as `fn runenv(mut self, runenv: Arc<dyn RunEnv>)`, so removed `SandboxSource` accordingly.

src/agent/builder.rs
khj809: The AgentBuilder was intentionally designed to take "instantiated runtime objects", so users can provide their own runtime objects without considering AgentSpec or AgentProvider. 
If this initialization path looks redundant in favor of `Agent::from_parts`, then we can rather consider making it as public API and remove builder.

src/agent/builder.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn subagent(mut self, card: AgentCard, agent: Agent) -> Self {
+        let sub = Arc::new(Mutex::new(agent));
+        self.tools.push(make_subagent_tool(card, sub));
+        self
+    }
+
+    /// Set the sandbox for this agent.
+    ///
+    /// Pass an `Arc<Sandbox>` to share an existing VM, or a `SandboxConfig` to
+    /// create a dedicated VM during [`build`](Self::build).
```
ljhh-0611: p.s.) It'd better to update comment too

src/agent/builder.rs
```diff
@@ -0,0 +1,237 @@
...
+            .unwrap_or_default();
+
+        let mut state = AgentState::new().history(history);
+        if let Some(runenv) = self.runenv {
+            state = state.runenv(runenv);
+        }
+
+        Ok(Agent::from_parts(self.model, self.tools, state))
+    }
+}
```
ljhh-0611, line 133: How about add real chat history optionally for AgentBuilder, not just instruction?

ljhh-0611 (approved):
I just commented some chore things. You can just pass away.
Thank you for your works!

src/agent/builder.rs
jhlee525: `AgentSpec` is to provide portability for agents in Ailoy. We’ve been thinking deeply about portability, and regardless of whether this approach has issues or not, I think backdoor path of defining `Agent` in other way feels makes the library more difficult to understand.

And also, We don't have to make the API complex.

```
let lm1 = LangModel::new().instruction()...
let tool1 = ...
let tool2 = ...
AgentBuilder::new().model(LangModel::new("claude")).tool(tool1).tool(tool2)
```

jhlee525: I think this can be easy way to defining `AgentBuilder`.

```rust
struct AgentBuilder {
  spec: AgentSpec,
  provider: Option<AgentProvider>,
  ...
}
```
keeping all the public functions are same,

and just defining the `build` function:

```rust
impl AgentBuilder {
...
pub fn build(self) -> Agent {
  ....
  Agent::new(self.spec)
}
}
```

jhlee525 (approved):

src/agent/builder.rs
khj809: Resolved in [01f63ba](https://github.com/brekkylab/ailoy/pull/391/commits/01f63ba695d5df96ee35b12208928d99765d3120)

src/agent/builder.rs
khj809: Resolved in [01f63ba](https://github.com/brekkylab/ailoy/pull/391/commits/01f63ba695d5df96ee35b12208928d99765d3120)

PR #392, opened 2026-04-28
feat: add TokenUsage and ContextManager

khj809:
## Background

As agents take on longer, multi-tool tasks, context window management is becoming a first-class concern. Without explicit control, conversations silently grow until they hit the model's input limit — either erroring out or degrading in ways the application can't observe or compensate for. This PR introduces the primitives needed to build agents that stay within bounds across extended runs.

### `ContextManager` — automatic history truncation

`ContextManager` monitors `input_tokens` from each LM response and truncates history when the count exceeds `max_input_tokens`. Rather than dropping messages, it replaces `Role::Tool` messages outside the preserve window with `"[context truncated]"` placeholders — **preserving the original `tool_use_id`**, since the Anthropic API rejects requests where a tool-use id in an assistant message has no matching tool result.

| Field | Default | Description |
|---|---|---|
| `max_input_tokens` | 30 000 | Truncation threshold |
| `preserve_recent_turns` | 3 | Recent user turns kept verbatim |

The system message is always preserved regardless of window size.

### `TokenUsage` in `MessageOutput`

`MessageOutput` now carries `usage: Option<TokenUsage>` populated for every top-level LM call. `TokenUsage` includes `input_tokens`, `output_tokens`, and optional cache token counts.

### 429 rate-limit retry in `LangModel`

All API-backed `LangModel` implementations now automatically retry on HTTP 429 responses. The retry logic respects the `Retry-After` header when present; if absent, it falls back to exponential backoff (1 s, 2 s, 4 s). Retries are capped at 3 attempts with a maximum wait of 10 s per retry. If all retries are exhausted the error is surfaced to the caller as before.


src/lang_model_impl/api/anthropic.rs
```diff
@@ -331,9 +331,33 @@ impl Unmarshal<MessageDeltaOutput> for AnthropicUnmarshal {
...
+                cache_read_input_tokens: u
+                    .get("cache_read_input_tokens")
+                    .and_then(|v| v.as_integer())
+                    .map(|v| v as u64),
+            });
+
         Ok(MessageDeltaOutput {
             delta,
             finish_reason,
+            usage,
```
grf53, line 371: It be would worth adding unit tests for `usage` unmarshalling to each `lang_model_impl/api/*.rs`.

src/lang_model/rt.rs
```diff
@@ -103,18 +103,51 @@ impl LangModel {
...
+                        last_text = Some(text);
+                        break;
+                    }
+                    (last_status.unwrap(), last_text.unwrap())
+                };
+
+                // Check response status
                 if !status.is_success() {
                     anyhow::bail!("API request failed with status {status}: {response_text}");
                 }
```
grf53, line 153: Would some new live-API tests be needed in the `lang_model/rt.rs`?
like:
- confirms that we get `usage` in the resp
- confirms that retries with mocked 429 response

src/agent/context.rs
```diff
@@ -0,0 +1,285 @@
...
+    datatype::Value,
+    message::{Message, Part, Role},
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize, schemars::JsonSchema)]
+pub struct ContextManager {
+    /// Triggers truncation when input_tokens from the previous API call exceeds this value.
+    pub max_input_tokens: u64,
+    /// Number of recent user turns to preserve after truncation (system message is always preserved separately).
+    pub preserve_recent_turns: usize,
```
nuri-yoo, line 10: The struct doc and PR body describe this as "Recent user turns kept verbatim", but `find_preserve_boundary` actually counts `Role::Assistant` messages. In a typical tool-use session a single user input expands into multiple assistant messages (`user → asst(tool_call) → tool → asst(text)`), so `preserve_recent_turns = 3` may keep only one user input's worth of history.

Concrete case:

```
history = [sys, u1, asst("a1"), u2, asst(tool_call), tool, asst("a2")]
preserve_recent_turns = 2  →  boundary lands at u2 (u1 falls outside the window)
```

Counting user messages instead would match the field name and the documented behavior. Alternatively, renaming the field to something like `preserve_recent_assistant_turns` would keep the current implementation but make the semantics explicit.

src/agent/context.rs
```diff
@@ -0,0 +1,285 @@
...
+            1
+        } else {
+            0
+        };
+
+        // ── Replace Tool messages outside the preserve window with placeholders ────
+        for i in start_idx..preserve_from {
+            if history[i].role == Role::Tool {
+                let original_id = history[i].id.clone();
+                let placeholder = Message::new(Role::Tool).with_contents([Part::value(
```
nuri-yoo: When this placeholder is sent through the Anthropic marshaller, it hits the `Part::Value` branch, which runs `serde_json::to_string(&value)` and JSON-encodes the inner string a second time. The wire body ends up as:

```
{"type":"tool_result","tool_use_id":"...","content":"\"[context truncated]\""}
```

The API accepts it (verified against `api.anthropic.com` — returns 200), but the model sees a quoted string rather than the bare placeholder. Switching to `Part::text("[context truncated]")` routes through the Text branch and produces `content: [{"type":"text","text":"[context truncated]"}]`, which is the more natural shape.

khj809: You're right. `Part::text("[context truncated]")` should be enough for tool contents, but still there's one more need to be fixed. The current marshaller converts the upper result as `{"type":"text","text":"[context truncated]"}`(by `part_to_value`), but it's rejected since [Anthropic accepts string or "array" of TextBlockParam](https://platform.claude.com/docs/en/api/messages/create#tool_result_block_param.content) (OpenAI has similar issue, and Gemini just gets arbitrary object so it's okay).
So I managed to marshal the tool result always to be a plain text (don't need to consider about array), and prevent `Part::Value` from being double-encoded. This fix is applied to Anthropic and OpenAI marshal, and Gemini does not need it.

src/agent/context.rs
khj809: Thanks, I didn't consider this scenario. Fixed to count user messages as you suggested.

src/lang_model_impl/api/anthropic.rs
khj809: Added `usage` unmarshal tests for anthropic, gemini and openai.

src/lang_model/rt.rs
khj809: Added tests for
- `usage` field inspection on real API response
- retrial on 429 by using mock axum server

nuri-yoo (approved):
Both review points addressed in 9461299. LGTM.

ljhh-0611 (approved):

PR #393, opened 2026-04-29
feat: add builtin `file_read` tool for multimodal files

khj809:
## Summary

Adds a `file_read` builtin tool that lets the LLM autonomously read files from the runenv(both local and sandbox) — both text and image files — within the agent loop. This enables multimodal file analysis (charts, screenshots, photos) without the caller needing to pre-load content into the message history.
The implementation mostly follows Claude Code's [FileReadTool](https://github.com/yasasbanukaofficial/claude-code/tree/main/src/tools/FileReadTool) with limited file types.

### New: `BuiltinToolProvider::FileRead` / `file_read` tool (`src/tool_impl/builtins/file_read.rs`)

- Detects file type by extension: PNG/JPEG/GIF/WEBP return `Part::Image` (embedded base64); all other readable files return `Part::Text`
- Unsupported binary types (PDF, ZIP, EXE, WASM, …) return a structured error
- Non-UTF-8 text files return a structured error
- Supports `offset` and `limit` parameters for windowed text reads (character-based)
- Reads via `RunEnv::read()` so it works in both local and sandbox environments
- Registered as `BuiltinToolProvider::FileRead {}` in `src/tool/provider.rs`

### Provider marshal fixes for `Part::Image` in `Role::Tool` messages

Previously all providers only used `contents[0]` of a tool-result message and only as plain text/JSON. Now each provider correctly handles images in tool results:

| Provider | Fix |
|---|---|
| **Anthropic** (`anthropic.rs`) | `tool_result.content` is now the full `contents` array; `Part::Value` is wrapped as `{"type":"text","text":"..."}` content block (required by the Anthropic API) |
| **Gemini** (`gemini.rs`) | Images go as sibling `inline_data` parts in the outer `parts` array alongside `functionResponse` — the `FunctionResponse` proto has no `parts` field, so nested `functionResponse.parts` was wrong |
| **OpenAI Responses** (`openai.rs`) | `function_call_output.output` becomes a content array with `input_image` blocks; `image_url` is a plain string (not `{"url":"..."}` object) |
| **ChatCompletion** (`chat_completion.rs`) | Images in tool results are substituted with `[image: mime_type]` / `[image at url]` text placeholders, since ChatCompletion backends don't reliably accept inline images in tool results |

### Gemini thinking model support (`gemini.rs`)

Gemini 3 thinking models return a `thoughtSignature` that must be replayed verbatim in subsequent turns. Added:
- Capture `thoughtSignature` from `thought` text parts and from `functionCall` parts (sibling field) during unmarshal
- Include `thoughtSignature` back in `thought` text parts and alongside `functionCall` parts during marshal
- `marshal_messages` now always passes `include_thinking = true` so signatures are never dropped mid-conversation

### Shared utility (`src/util/truncate.rs`)

Extracted `middle_truncate` into `src/util/truncate.rs`, removing local copies from `bash.rs` and `sandbox.rs`.

## Test plan

- [x] `cargo test --lib tool_impl::builtins::file_read` — 8 unit tests covering text, image, offset/limit, UTF-8 error, unsupported extension, missing path, nonexistent path
- [x] `cargo test --lib lang_model_impl::api::gemini::tests::test_tool_result_with_image` — 2-turn Gemini 3 interaction: step 1 gets real `functionCall` + `thoughtSignature`, step 2 sends JPEG tool result and verifies text response (model correctly describes Jensen Huang photo)
- [x] `cargo test --lib lang_model_impl::api` — all 13 provider unit tests pass; `test_tool_result_with_image` for Anthropic and OpenAI skipped without API keys (require compile-time `test_with::env`)
- [x] Manual smoke test with Anthropic/OpenAI/Gemini `test_tool_result_with_image` (requires API keys at compile time)


src/tool_impl/builtins/file_read.rs
```diff
@@ -0,0 +1,278 @@
...
+            Some("gif") => FileKind::Image("image/gif"),
+            Some("webp") => FileKind::Image("image/webp"),
+            Some(
+                e @ ("pdf" | "ipynb" | "zip" | "tar" | "gz" | "exe" | "bin" | "so" | "dylib"
+                | "wasm"),
+            ) => return error_message(id, format!("unsupported file type: .{e}")),
+            _ => FileKind::Text,
+        };
+
+        let bytes = match ctx.runenv.read(path).await {
```
nuri-yoo, line 94: file_read has no size cap. Verified locally with a 50MB text file: returns the full 49MB string in 9ms, which then marshals to a ~50MB JSON body across all four providers (49.99MB anthropic / openai / gemini / chat_completion). 5MB JPEG is 6.67MB after base64 + JSON wrapper.

If the model picks a path like `/var/log/system.log` or any GB-scale file, the ailoy process OOMs before the API call even leaves. The bash tool in this same directory has `MAX_OUTPUT_CHARS = 30_000`. Claude Code's FileReadTool (cited in the PR body) caps at 2000 lines / per-line truncation by default — that part of the pattern isn't ported here.

Suggest a default byte cap; offset/limit then becomes the way to read past it:

```rust
const MAX_FILE_BYTES: usize = 10 * 1024 * 1024;

let bytes = match ctx.runenv.read(path).await {
    Ok(b) if b.len() > MAX_FILE_BYTES =>
        return error_message(id, format!(
            "file too large: {} bytes (limit: {}); use offset/limit to read in chunks",
            b.len(), MAX_FILE_BYTES
        )),
    Ok(b) => b,
    Err(e) => return error_message(id, format!("read failed: {e}")),
};
```

For images, base64 grows by 1.33x, so a separate image cap (e.g. 5MB raw) is worth considering.

src/lang_model_impl/api/gemini.rs
```diff
@@ -70,21 +79,57 @@ fn marshal_message(msg: &Message, include_thinking: bool) -> Value {
     if msg.role == Role::Tool {
         let tool_call_id = msg.id.clone().expect("Tool call id must exist.");
         let (tool_name, _) = tool_call_id.split_once("/").unwrap();
```
nuri-yoo: nit: this unwrap relies on `tool_call_id` always containing `/`, which isn't enforced by the type. The slash is only present when an id synthesized by this file's unmarshal at line 384 (`format!("{}/call-{uuid8}", ...)`) round-trips back.

In practice the agent runtime preserves the synthesized id end-to-end, so this doesn't fire on the normal path. It does fire when a `Message` with a non-Gemini id is routed here — verified locally:

```rust
let assistant = Message::new(Role::Assistant).with_tool_calls([Part::function(
    "toolu_01ABCdef", "file_read", to_value!({"path": "/x"}),
)]);
let tool_msg = Message::new(Role::Tool).with_id("toolu_01ABCdef")
    .with_contents([Part::text("hello")]);
GeminiMarshal::default().marshal(&tool_msg);
// panicked at src/lang_model_impl/api/gemini.rs:81:59:
// called `Option::unwrap()` on a `None` value
```

The PR's own anthropic.rs / openai.rs tests use exactly this pattern (`with_id("toolu_test_001")`); routing the same shape to Gemini panics. Three-line fix:

```rust
let tool_name = tool_call_id
    .split_once('/')
    .map(|(name, _)| name)
    .unwrap_or(tool_call_id.as_str());
```

src/lang_model_impl/api/openai.rs
```diff
@@ -58,11 +58,29 @@ fn marshal_message(msg: &Message, include_thinking: bool) -> Vec<Value> {
     };
 
     if msg.role == Role::Tool {
```
nuri-yoo, line 60: (Note: this comment is anchored on the changed Role::Tool branch but actually concerns line 55 in the user-message branch above, which is unchanged in this PR.)

The tool-result path below was changed to use `image_url` as a string. The user-message branch at line 55 still emits the object form `{"url": "..."}`. Live verified that OpenAI Responses rejects the object form there too:

```
USER message, image_url={"url": "data:..."}: HTTP 400
   "Invalid type for 'input[0].content[1].image_url':
    expected an image URL, but got an object instead."
USER message, image_url="data:...":          HTTP 200
```

Same fix mechanically applies on line 55:

```rust
to_value!({"type": "input_image", "image_url": url})
```

src/tool_impl/builtins/file_read.rs
```diff
@@ -0,0 +1,278 @@
...
+    };
+    chars[start..end].iter().collect()
+}
+
+pub async fn build_file_read_tool() -> anyhow::Result<ToolFactory> {
+    let desc = ToolDescBuilder::new("file_read")
+        .description(
+            "Reads a file from the filesystem. Supports text files (returned as plain text) \
+             and image files (PNG/JPEG/GIF/WEBP, returned as an image the model can see). \
+             Binary or unsupported file types return an error. Paths must be absolute.",
```
nuri-yoo, line 42: nit: `Paths must be absolute.` is in the description but not enforced in code — relative paths fall through to `ctx.runenv.read(path)` and either succeed or fail with a generic OS error.

The description does influence the model: tested 22 prompts across Haiku 4-5 and Sonnet 4-5 with a system prompt giving the cwd. With this description ~50% of prompts get the model to construct an absolute path, but explicit user instructions ("pass `foo.txt` verbatim", "use `../../etc/passwd` exactly") bypass it.

Two ways to make doc and code agree:

```rust
let path = Path::new(&path_str);
if !path.is_absolute() {
    return error_message(id, "path must be absolute");
}
```

or relax the description to match current behavior:

```
Reads a file from the filesystem. Paths can be absolute or relative
(resolved against the agent's working directory).
```

src/lang_model_impl/api/gemini.rs
khj809: 1. Usually we don't assume that the tool call id is hand-crafted by users. If this is allowed, this is not a problem of Gemini only; another providers also fail on strange tool call ids (e.g. openai does not even allow hyphen).
2. Originally Gemini didn't return tool call id on `FunctionCall`, and that's why we added synthetic call id on unmarshal `functionCall`. However, it seems now they provide this one ([docs](https://ai.google.dev/api/caching?hl=ko&_gl=1*lnl903*_up*MQ..*_ga*NDMwOTI3MDg0LjE3NzgxMzk1MzU.*_ga_P1DBVKWT6V*czE3NzgxNDU2NDYkbzIkZzAkdDE3NzgxNDU2NDYkajYwJGwwJGg1Mzg0NDQ1OA..#FunctionCall)), so we can now eliminate such id synthesis eventually!
3. By using the provided call id, we don't need to assume the tool call id as slash-separated special form anymore. Let me fix this.

nuri-yoo: Got it, thanks for the heads-up. From a live check, only 3.x seems to populate it — 2.5 still ships without `id`.

khj809: Wow that's really bullshit..
According to their [model deprecation schedules](https://ai.google.dev/gemini-api/docs/deprecations?hl=ko), they're planning to stop the support for 2.5 models in October. So let's retain the current logic as fallback when `id` is not found in `functionCall` until they're fully deprecated.

khj809: Sorry I was totally misunderstanding.
The exact reason for synthesizing call id with function name was that `name` in `FunctionResponse` is mandatory (unlike other vendors) but there's no dedicated field to store it on our `Message`. Therefore, we should still keep the call id synthesis as before and fetch the `name` of function from that, regardless of using either provided or random call id. Please nevermind my comments so far.

Back to your suggestion, I think it's better to explicitly raise the error when the call id is out of the expected format rather than silently set `tool_name` as incorrect one. Instead, we can use `.expect("tool call id shuold be a form of {function name}/{random string}")` rather than just using `.unwrap()`.

src/tool_impl/builtins/file_read.rs
khj809: Applied limit as 10MB for files and 5MB for images.

src/lang_model_impl/api/openai.rs
khj809: Applied the fix.

src/tool_impl/builtins/file_read.rs
khj809: Applied the fix.

nuri-yoo (approved):

jhlee525 (approved):
I'll merge file_read tool & read tool in #395

PR #395, opened 2026-05-07
filesystem-editing builtin tools (read/write/edit/apply_patch) + RunEnv directory ops

jhlee525:
## Summary
- Extends `RunEnv` with directory operations (`ls`, `mkdir`, `rmdir`) and a recursive `Dirent` tree (Dir/File with owner-effective permission `u8` 4|2|1 and size). `Local` uses native `tokio::fs`; `Sandbox` shells out via `exec` with a portable POSIX script.
- Adds four file-editing builtins routed through `ctx.runenv` so they work in both `Local` and `Sandbox`:
  - **`read`** — `cat -n` formatted output with `offset`/`limit` slicing.
  - **`write`** — overwrite-or-create.
  - **`edit`** — exact string replacement; `replace_all` flag; rejects ambiguous matches.
  - **`apply_patch`** — `*** Begin Patch` envelope with `Add File` / `Update File` / `Delete File` ops and `@@`-delimited hunks; rejects when the `before` block doesn't match exactly once.

### DIFF
--- a/src/runenv/sandbox.rs
+++ b/src/runenv/sandbox.rs
@@ -16,6 +16,7 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
 use super::{Dirent, ExecResult, RunEnv};
+use crate::util::truncate::middle_truncate;
 
 fn fresh_sandbox_name() -> String {
     format!("ailoy-{}", Uuid::new_v4())
@@ -506,8 +507,8 @@ impl Sandbox {
     ) -> anyhow::Result<ExecResult> {
         match result {
             Ok(output) => {
-                let stdout = truncate_output(output.stdout().unwrap_or_default(), max_output_chars);
-                let stderr = truncate_output(output.stderr().unwrap_or_default(), max_output_chars);
+                let stdout = middle_truncate(output.stdout().unwrap_or_default(), max_output_chars);
+                let stderr = middle_truncate(output.stderr().unwrap_or_default(), max_output_chars);
                 Ok(ExecResult {
                     stdout,
                     stderr,
@@ -646,19 +647,6 @@ fn apply_volume_mount(builder: SandboxBuilder, mount: &VolumeMount) -> SandboxBu
             })
         }
     }
-}
-
-fn truncate_output(s: String, max_chars: usize) -> String {
-    let chars: Vec<char> = s.chars().collect();
-    if chars.len() <= max_chars {
-        return s;
-    }
-    let head = max_chars / 2;
-    let tail = max_chars - head;
-    let omitted = chars.len() - head - tail;
-    let head_str: String = chars[..head].iter().collect();
-    let tail_str: String = chars[chars.len() - tail..].iter().collect();
-    format!("{head_str}\n\n... [{omitted} characters omitted] ...\n\n{tail_str}")
 }
 
 #[cfg(test)]
--- a/src/runenv/sandbox.rs
+++ b/src/runenv/sandbox.rs
@@ -15,7 +15,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
-use super::{ExecResult, RunEnv};
+use super::{Dirent, ExecResult, RunEnv};
 
 fn fresh_sandbox_name() -> String {
     format!("ailoy-{}", Uuid::new_v4())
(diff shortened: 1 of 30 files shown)


src/runenv/sandbox.rs
```diff
@@ -202,6 +202,94 @@ impl RunEnv for Sandbox {
...
+  [ -w "$entry" ] && w=2
+  [ -x "$entry" ] && x=1
+  perm=$((r + w + x))
+  printf '%s\t%s\t%s\t%s\n' "$kind" "$perm" "$size" "$entry"
+done"#,
+            path = path_str
+        );
+        let result = self
+            .exec("sh".to_string(), vec!["-c".to_string(), script], None)
+            .await?;
```
khj809: Rather than executing a shell command, it would be better to use microsandbox's native filesystem API ([list a directory](https://docs.microsandbox.dev/sandboxes/filesystem#list-a-directory)).

src/runenv/sandbox.rs
```diff
@@ -202,6 +202,94 @@ impl RunEnv for Sandbox {
...
+    }
+
+    async fn mkdir(&self, path: &Path) -> anyhow::Result<()> {
+        let result = self
+            .exec(
+                "mkdir".to_string(),
+                vec!["-p".to_string(), path.to_string_lossy().into_owned()],
+                None,
+            )
+            .await?;
```
khj809: ditto ([mkdir](https://docs.microsandbox.dev/sdk/typescript/filesystem#mkdir))

src/runenv/sandbox.rs
```diff
@@ -202,6 +202,94 @@ impl RunEnv for Sandbox {
...
+    }
+
+    async fn rmdir(&self, path: &Path) -> anyhow::Result<()> {
+        let result = self
+            .exec(
+                "rmdir".to_string(),
+                vec![path.to_string_lossy().into_owned()],
+                None,
+            )
+            .await?;
```
khj809: ditto ([removeDir](https://docs.microsandbox.dev/sdk/typescript/filesystem#removedir))

src/tool_impl/builtins/write.rs
```diff
@@ -0,0 +1,137 @@
...
+        }))
+        .build();
+
+    let f = ToolFunc::new(|args: Value, ctx: ToolContext| async move {
+        let Some(path) = args.pointer("/file_path").and_then(|v| v.as_str()) else {
+            return crate::to_value!({
+                "error": "missing required parameter: file_path",
+                "phase": "validation"
+            });
+        };
```
khj809, line 42: `read` tool checks if the provided path is absolute, but `write` tool does not check it. Suggest to check consistently:

```suggestion
        let Some(path) = args.pointer("/file_path").and_then(|v| v.as_str()) else {
            return crate::to_value!({
                "error": "missing required parameter: file_path",
                "phase": "validation"
            });
        };
        let path = Path::new(path_str);
        if !path.is_absolute() {
            return crate::to_value!({
                "error": "file_path must be absolute",
                "phase": "validation"
            });
        }
```

src/tool_impl/builtins/edit.rs
```diff
@@ -0,0 +1,274 @@
...
+        }))
+        .build();
+
+    let f = ToolFunc::new(|args: Value, ctx: ToolContext| async move {
+        let Some(path) = args.pointer("/file_path").and_then(|v| v.as_str()) else {
+            return crate::to_value!({
+                "error": "missing required parameter: file_path",
+                "phase": "validation"
+            });
+        };
```
khj809, line 52: ditto

```suggestion
        let Some(path) = args.pointer("/file_path").and_then(|v| v.as_str()) else {
            return crate::to_value!({
                "error": "missing required parameter: file_path",
                "phase": "validation"
            });
        };
        let path = Path::new(path_str);
        if !path.is_absolute() {
            return crate::to_value!({
                "error": "file_path must be absolute",
                "phase": "validation"
            });
        }
```

grf53 (approved):
LGTM

src/tool_impl/builtins/apply_patch.rs
```diff
@@ -0,0 +1,396 @@
...
+                Ok(msg) => summary.push(Value::from(msg)),
+                Err(e) => {
+                    return crate::to_value!({
+                        "error": format!("{e}"),
+                        "applied": Value::Array(summary),
+                        "phase": "apply"
+                    });
+                }
+            }
+        }
```
khj809, line 230: When a patch contains multiple file operations and one fails mid-way, all previously applied operations are left on disk. To ensure the atomicity of the tool, how about adding some rollback mechanism or dry-run validation?

src/runenv/sandbox.rs
jhlee525: Applied

src/runenv/sandbox.rs
jhlee525: Applied

src/runenv/sandbox.rs
jhlee525: Applied

src/tool_impl/builtins/apply_patch.rs
jhlee525: Claude says:

Dry-run is the better fit here. The realistic failure modes — hunk-not-found, ambiguous-match, non-UTF-8, missing file for Update/Delete — are all deterministic and catchable by walking the ops once against in-memory copies before any write hits disk. Rollback adds state-tracking complexity (snapshot every touched file, recreate deleted ones, handle restore-itself-fails) for marginal benefit, since the only failures it would catch beyond dry-run are mid-write I/O errors which are rare and catastrophic either way.

This makes sense. Most cases(syntax errors) are caught upfront anyway, and the only thing rollback would catch is "runtime IO errors". The cost of implementing it seems too high relative to what you gain (OS IO errors).

khj809 (approved):

PR #396, opened 2026-05-10
refactor(tool): name-keyed registry, ToolDesc-driven specs, fs tools

jhlee525:
## Summary
- Redesigns the tool system around a name-keyed `ToolProvider` registry that resolves directly to `ToolFunc` via `provide(&[ToolDesc])`, retiring the `ToolFactory` / runtime-`Tool` / ordered-list model.
- Rebuilds `ToolFunc` on an HRTB closure so async tools can borrow `&dyn RunEnv` across `.await`, adds a `tool_func!` construction macro, and moves `ToolDesc` from `message/` to `tool/desc.rs`.
- `AgentSpec.tools` is now `Vec<ToolDesc>` (was `Vec<String>`) — specs describe what they expose, not just names. Looked up by `desc.name` against the provider.
- Adds builtin filesystem tools (`read`, `write`, `edit`, `apply_patch`, `find`, `grep`) routed through `RunEnv`, and folds the old `file_read` into the expanded `read`. `ToolProvider::new()` pre-registers every builtin; use `empty()` to opt out.
- `AgentSpec` / `AgentBuilder` gain `fs_tools()`, `web_search_tool()`, `bash_tool()`, `find_tool()`, `grep_tool()`. `fs_tools()` selects per model family: `openai/*` → `read + apply_patch` (codex-style envelope), others → `read + write + edit`.
- Extends `RunEnv` with `ls` / `mkdir` / `rmdir` returning a recursive `Dirent` tree (owner rwx bits + size); `Local` uses `tokio::fs`, `Sandbox` shells out via a portable POSIX script for matching semantics.
- `LangModelProvider::default()` now auto-loads `.env` and registers `openai/*` / `anthropic/*` / `gemini/*` from `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY`. `LangModelProvider::new()` is the empty-registry constructor. `make_runtime` renamed to `provide` for consistency with the tool side.
- Dependency changes: `dotenvy` promoted from dev-dep to runtime; adds `infer` and `encoding_rs` for the new read tool.

## Breaking changes
- `AgentBuilder::tool(key)` / `.tools(keys)` now take `ToolDesc` / `IntoIterator<Item = ToolDesc>` instead of strings. Use the new `*_tool()` helpers or `ToolDescBuilder` to migrate.
- `BuiltinToolProviderElem` and `ToolProvider::custom(factory)` / `.web_search()` / `.python_repl()` / `.bash()` / `.file_read()` / `.mcp_*()` / `.a2a()` chained builders are gone; register via `insert_func` / `insert_func_factory` / `insert_a2a` (MCP variants still `todo!()`).
- `LangModelProvider::make_runtime` → `provide`. `ToolProvider::make_runtime` → `provide(&[ToolDesc])` returning `HashMap<String, ToolFunc>`.
- `ToolDesc` is no longer re-exported from `message::`; import from `tool::`.
- `file_read` builtin removed — its capabilities are now in `read`.

## Test plan
- [ ] \`cargo build\` (default features)
- [ ] \`cargo build --features sandbox\`
- [ ] \`cargo test\`
- [ ] Smoke-run an agent against \`openai/*\`, \`anthropic/*\`, \`gemini/*\` with \`LangModelProvider::default()\` picking up env keys
- [ ] Exercise \`fs_tools()\` against an \`openai/*\` model (verify \`apply_patch\`) and against a non-\`openai\` model (verify \`read+write+edit\`)
- [ ] Exercise \`find\` / \`grep\` builtins under both \`Local\` and \`Sandbox\` runenvs
- [ ] Verify \`RunEnv::ls\` permission bits match between \`Local\` and \`Sandbox\` for the same tree


src/tool_impl/builtins/edit.rs
```diff
@@ -39,52 +36,48 @@ pub async fn build_edit_tool() -> anyhow::Result<ToolFactory> {
...
+                "error": "missing required parameter: path",
+                "phase": "validation",
             });
         };
-        if !Path::new(path).is_absolute() {
-            return crate::to_value!({
-                "error": "file_path must be absolute",
-                "phase": "validation"
-            });
-        }
```
khj809, line 58: Why this is removed?

src/tool_impl/builtins/write.rs
```diff
@@ -29,77 +23,68 @@ pub async fn build_write_tool() -> anyhow::Result<ToolFactory> {
...
+                "error": "missing required parameter: path",
+                "phase": "validation",
             });
         };
-        if !Path::new(path).is_absolute() {
-            return crate::to_value!({
-                "error": "file_path must be absolute",
-                "phase": "validation"
-            });
-        }
```
khj809, line 48: ditto

src/lang_model/provider.rs
```diff
@@ -56,16 +56,42 @@ pub enum LangModelProviderElem {
...
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 #[serde(transparent)]
 #[schemars(transparent)]
 pub struct LangModelProvider {
     inner: BTreeMap<String, LangModelProviderElem>,
 }
 
+impl Default for LangModelProvider {
+    fn default() -> Self {
+        dotenvy::dotenv().ok();
```
khj809: I think reading and exporting `.env` should be handled in application layer (including tests) rather than library layer.

src/agent/spec.rs
```diff
@@ -62,13 +73,43 @@ impl AgentSpec {
...
+        self.tools.push(tool);
         self
     }
 
-    pub fn tools(mut self, tools: impl IntoIterator<Item = impl Into<String>>) -> Self {
-        self.tools = tools.into_iter().map(|v| v.into()).collect();
+    pub fn tools(mut self, tools: impl IntoIterator<Item = ToolDesc>) -> Self {
+        self.tools = tools.into_iter().collect();
+        self
+    }
```
khj809, line 85: `tool()` pushes the new item, but `tools()` replaces the whole items. `.tool("a").tools(vec!["b"])` would remove "a".
How about changing `tools()` to append the items?

src/tool_impl/a2a.rs
```diff
@@ -2,14 +2,17 @@ use serde::{Deserialize, Serialize};
...
+    tool_func,
 };
 
 // ── Tool constructor ──────────────────────────────────────────────────────────
 
-async fn a2a_parts(url: Url) -> anyhow::Result<(crate::message::ToolDesc, ToolFunc)> {
+/// Discover the remote A2A agent at `url` and build its [`ToolDesc`].
+///
+/// Performs a network fetch of the agent card.
+pub(crate) async fn get_a2a_tool_desc(url: &Url) -> anyhow::Result<ToolDesc> {
```
khj809, line 15: There's a lot of unused warnings in `a2a.rs`.

src/tool_impl/builtins/python_repl/mod.rs
```diff
@@ -3,14 +3,12 @@ mod runner;
...
-    datatype::Value,
-    message::ToolDescBuilder,
-    tool::{ToolContext, ToolFactory, ToolFunc},
+    tool::{ToolDesc, ToolDescBuilder, ToolFunc},
+    tool_func,
 };
 
-pub async fn build_python_repl_tool() -> anyhow::Result<ToolFactory> {
-    let desc = ToolDescBuilder::new("python_repl")
+pub fn get_python_repl_tool_desc() -> ToolDesc {
```
khj809, line 10: unused warning

src/tool_impl/builtins/grep.rs
```diff
@@ -0,0 +1,614 @@
...
+                        break;
+                    }
+                }
+                OutputMode::FilesWithMatches => {
+                    if files_matched.len() >= limit {
+                        truncated = true;
+                        break;
+                    }
+                }
+                OutputMode::Count => {}
```
khj809, line 315: Why there's no limit check in Count mode?

src/tool_impl/builtins/read.rs
```diff
@@ -92,121 +93,116 @@ pub async fn build_read_tool() -> anyhow::Result<ToolFactory> {
...
+                            "validation",
+                        );
+                    }
+                    let (cow, _, _) = encoding_rs::UTF_8.decode(&bytes);
+                    let text = cow.into_owned();
+                    let offset = args
+                        .pointer("/offset")
+                        .and_then(|v| v.as_integer())
+                        .map(|n| n.max(1) as usize)
+                        .unwrap_or(1);
```
khj809: The default offset in tool description is 0, but it becomes 1 due to `n.max(1)` or `.unwrap_or(1)`.

src/tool_impl/builtins/grep.rs
```diff
@@ -0,0 +1,614 @@
...
+            '.' | '+' | '(' | ')' | '{' | '}' | '|' | '^' | '$' | '\\' => {
+                pat.push('\\');
+                pat.push(c);
+            }
+            _ => pat.push(c),
+        }
+    }
+    pat.push('$');
+    Regex::new(&pat).map_err(|e| format!("invalid glob {glob:?}: {e}"))
+}
```
khj809: There's separate `glob_to_regex` functions across `grep.rs` and `glob.rs`. Is there a reason not to use the unified one?

src/tool_impl/builtins/edit.rs
jhlee525: I take the **absolute constraints** to be conventions by Claude's read/write/edit tools. (Not should be general behavior)

If those conventions are required, the agent should supply them through a system prompt.

src/lang_model/provider.rs
jhlee525: Okay, we can remove the `.env` parsing logic from here.

However, using environment variables (e.g. `OPENAI_API_KEY` or `ANTHROPIC_API_KEY`) still seems useful. Is it okay to keep that?

src/agent/spec.rs
jhlee525: 👍

src/tool_impl/a2a.rs
jhlee525: Currently, a2a is not being used. Supporting a2a would require additional refactoring of the subagent tool, but the scope seems to be growing, so I've excluded it from this change for now.

One option would be to exclude the a2a logic from the build and bring it back later when it's needed.

src/tool_impl/builtins/python_repl/mod.rs
jhlee525: make it public, can user call this API

src/tool_impl/builtins/grep.rs
jhlee525: Count mode only returns aggregate numbers (`match_count` and `files_matched`), not a list — so there's nothing whose size needs capping. The limit parameter exists to cap the result payload: lines in Content mode, file paths in FilesWithMatches mode. Both can grow unbounded with matches, hence the early-break.

In Count mode, the return is always just two integers regardless of input size. Applying limit there would actually be incorrect — stopping early would produce an undercount, defeating the whole purpose of asking for a count. That's also why Count mode's response has no truncated field, unlike the other two.

src/tool_impl/builtins/read.rs
jhlee525: It's a bug. Thank you for clarifying.

src/tool_impl/builtins/grep.rs
jhlee525: There is no special reason. The only reason is that not existence of proper slot for putting common logic. (I thought it is not big issue...)

src/lang_model/provider.rs
khj809: Sure. This comment is not complaining about them.

src/tool_impl/builtins/grep.rs
khj809: Got it. Then it would be better to comment the reason briefly.

src/tool_impl/builtins/grep.rs
khj809: Since the function name is `glob_to_regex`, how about locating it only on `glob.rs` and import it from `grep.rs`?

src/lib.rs
```diff
@@ -9,5 +9,5 @@ pub(crate) mod macros;
 pub mod message;
 pub mod runenv;
 pub mod tool;
-pub(crate) mod tool_impl;
+pub mod tool_impl;
```
khj809, line 12: Does `tool_impl` mod meant to be public?
If there's something need to be exposed publicly in `tool_impl`(e.g. `get_python_repl_tool_desc`), I think they can be re-exported via `tool` mod.

src/tool_impl/builtins/edit.rs
khj809: However, the tool description says it's "the absolute path to the file to modify" (same for `write` tool).
Also, `read` tool is still checking if the path is absolute.
We should apply the unified behavior in either way.

src/tool_impl/a2a.rs
khj809: I see. Since you commented the whole `a2a` mod, it seems enough for now.

src/tool_impl/builtins/python_repl/mod.rs
khj809: Related to https://github.com/brekkylab/ailoy/pull/396#discussion_r3225286381

src/tool_impl/builtins/edit.rs
jhlee525: Oh, I removed it but it is reverted when I merge develop.

I'll fix it.

src/lib.rs
jhlee525: Yes, this change came about because `AgentSpec` now takes `ToolDesc` as input, and users now have occasion to work with `ToolDesc` for built-in tools.

Re-exporting from tool sounds like a good idea. 

P.S. Down the line, it might even make sense to move `tool_impl` inside of `tool`

jhlee525: I'll do it in following PRs.

src/tool_impl/builtins/grep.rs
jhlee525: Gread idea. I applied

src/tool_impl/builtins/edit.rs
khj809: If this is the way to go, then please remove `!path.is_absolute()` check from `read` tool too.

khj809 (approved):

PR #397, opened 2026-05-11
Introducing skills as a file within `RunEnv`

### LINKED ISSUE #398: Introducing Skills
## Description
Establish a foundation in Ailoy that allows the use of skills similar in level to those included in https://github.com/anthropics/skills based on `RunEnv`, and enable the creation of such `Agent`s based on `AgentSpec`.

## Reference
https://github.com/brekkylab/ailoy/blob/098a8289272ccce3cf5719e20732050cb047d04a/src/agent/rt.rs#L855

grf53:
## Summary

Introduces file-based skills as a first-class part of `AgentSpec`, so agent definitions are fully self-contained and reproducible across runenvs.

A *skill* is a directory containing `SKILL.md` (YAML frontmatter) plus any supporting files. The spec carries the file content in `files`, the declared list in `skills`, and an optional growth area in `skill_root`:

```
spec → build → run (mutate runenv) → snapshot → spec' → JSON → elsewhere → build → equivalent
```

## API

```rust
pub struct AgentSpec {
    // ... existing fields ...
    pub files: Vec<FileEntry>,        // content pre-filled into runenv at first run
    pub skills: Vec<PathBuf>,         // declared skill dirs (each contains SKILL.md)
    pub skill_root: Option<PathBuf>,  // optional growth area, scanned by snapshot
}

// AgentBuilder additions
builder.file(FileEntry::new(path, content));
builder.skill(dir);
builder.skills([d1, d2]);
builder.skill_root(root);

// FileEntry
FileEntry::new(path, content);
fe.write_to(runenv).await;
FileEntry::read_from(runenv, path).await;
```

## Classes
<details>
<summary>Class diagram</summary>

```mermaid
classDiagram
    class FileEntry {
        +PathBuf path
        +Bytes content
        +new(path, content)
        +write_to(runenv)
        +read_from(runenv, path)
    }

    class SkillMeta {
        +String name
        +String description
        +PathBuf dir
        +skill_md_path()
    }

    class AgentSpec {
        +Vec~FileEntry~ files
        +Vec~PathBuf~ skills
        +Option~PathBuf~ skill_root
        +file(FileEntry)
        +files(iter)
        +skill(dir)
        +skills(iter)
        +skill_root(root)
    }

    class AgentBuilder {
        +file(FileEntry)
        +files(iter)
        +skill(dir)
        +skills(iter)
        +skill_root(root)
        +build()
    }

    class Agent {
        +AgentSpec spec
        +snapshot() AgentSpec
        +skills() &[PathBuf]
        +skill_root() Option<&Path>
        +spec() &AgentSpec
    }

    AgentSpec o-- "*" FileEntry : declares
    AgentBuilder ..> AgentSpec : builds
    AgentBuilder ..> Agent : produces
    Agent --> AgentSpec : holds
    Agent ..> FileEntry : materialises / scans
    Agent ..> SkillMeta : discovers
```

</details>

## `skills` ~& `skill_root`~ — declared ~vs grown skills~

The two fields separate concerns:

- **declared skills** (`skills`) — known up front, content embedded in `files`. Always present, always round-tripped.
- ~**growth area** (`skill_root`)~ — where the agent creates *new* skills at runtime; `snapshot()` does one `ls` there to capture additions.

**`skill_root = None` is meaningful**: it declares "this agent is not expected to grow new skills at runtime." It does not prevent or forbid runtime skill creation — the agent technically can still write files anywhere — but snapshot will not look for them, so any such additions will not round-trip. The natural cases:

- **Sub-agents**: they're stateless `ToolFunc` closures rebuilt per call — no persistent handle, no runtime skill mutation. Sub-specs typically have `skill_root = None`.
- **Frozen agents**: any agent whose skill set is fixed by design. (No concrete example in the current codebase.)

Because sub-agent paths are **not rewritten** by the parent (sub-spec taken as-is), sub-specs are portable across parents. The author picks distinct absolute paths to avoid filesystem collisions.

When `skill_root = Some(path)`, the auto-rendered "Available Skills" block in the system instruction tells the model to create new skills under that exact path.

## Materialisation ~& snapshot~
- materialise: **Lazy** (first `run()` or `snapshot()`) and **write-once** — existing files in the runenv are preserved, so runtime mutations survive across builds.
- ~`snapshot()` rebuilds `files` from three sources~:
  1. **Walk each declared skill dir** — captures every file inside, including runtime additions (e.g. a new `helper.py` placed next to a declared `SKILL.md`).
  2. **Walk each newly discovered dir under `skill_root`** — captures the whole subtree of skills the agent created at runtime.
  3. **Re-read declared files outside any skill dir** — recovers non-skill seeded files; silently dropped if missing on disk.
- `skills` becomes declared list + new dirs found under `skill_root`. Declared entries are preserved even if their files were deleted on disk (intent is kept).
- Sub-spec is kept as declared (no recursion).

## Testing

- **15 skill tests** — frontmatter parsing, table rendering with/without `skill_root`, runenv discovery, in-memory projection.
- **26 agent tests** — builder API, per-agent isolation, same-name skills at different paths, snapshot scenarios (runtime add via `skill_root`, overwrite, delete), `skill_root` advertised in system instruction.
- **6 spec tests** — JSON round-trip including nested sub-agents.
- **End-to-end** (`test_convert_pdf_to_md_skill`, sandbox + Anthropic) — auto-rendered SKILL.md table → Docling install → PDF→Markdown. Gated by `#[ignore]`.

5 pre-existing failures (`lang_model::*`) require API keys and are unrelated.

## Test plan
- [x] `cargo test --lib skill::` → 15/15
- [x] `cargo test --lib agent::` → 26/26 (7 ignored, API-gated)
- [x] `cargo test --lib` → 189 pass / 5 pre-existing env failures
- [x] `ANTHROPIC_API_KEY=sk-ant-... cargo test --features sandbox --lib skill::skill::tests::test_benchmark_python_snippet_skill -- --ignored`


jhlee525:
I think it was my mistake to suggest managing skills as list of FilePath. I didn't consider the case of "updating a newly created skill."

I think it would be fine to go back to using just `skill_dir` as we originally planned. That said, if you think it's necessary, it's also fine to keep using both `skill` and `skill_dir` as-is.

jhlee525 (approved):
👍

grf53:
> I think it would be fine to go back to using just `skill_dir` as we originally planned. That said, if you think it's necessary, it's also fine to keep using both `skill` and `skill_dir` as-is.

It gave me an opportunity to rethink the paths of skills.

Using only `skill_dir` can make things complicated considering sub-agents:
1. Common skills must always be copied (or symlinked) and placed under each agent's `skill_dir`.
2. Case where the paths of subagents' `skill_dir` are always set based on the top-level agent's `skill_dir`
  - implicit rules regarding the structure are required.
  - The directory structure must follow the sub-agent tree structure.
3. Case where specifying the skill_dir for each sub-agent separately
  - the isolation boundaries become blurred, such as when different agents overwrite each other.
  - more exception handling and implicit rules are included to prevent crashes.

There will be trade-offs, but it has been considered the most stable method for now.

grf53:
@khj809 I just removed the test `test_convert_pdf_to_md_skill`.
You can see the details in the commit https://github.com/brekkylab/ailoy/pull/397/changes/f5e52ec298c5f1c093d2dfac493b1dc9b8123e87.

src/agent/builder.rs
```diff
@@ -143,6 +143,41 @@ impl AgentBuilder {
...
+    pub fn skill(mut self, dir: impl Into<PathBuf>) -> Self {
+        self.spec = self.spec.skill(dir);
+        self
+    }
+
+    /// Declare multiple skill directories at once.
+    pub fn skills(mut self, dirs: impl IntoIterator<Item = PathBuf>) -> Self {
+        self.spec = self.spec.skills(dirs);
+        self
+    }
```
khj809: Since files and skills exist separately, users need to proceed two steps manually, and it can be easily broken in the following mis-usages:
- Calling `builder.file()` without `.skill()`: files are materialized but skill does not appear in system prompt
- Calling `builder.skill()` without `.file()`: skill appears in system prompt but agent can never find its `SKILL.md`.

Why don't we just use `builder.skill(dir, entries)` which can minimize such mistakes?

src/skill/skill.rs
```diff
@@ -0,0 +1,683 @@
...
+}
+
+async fn read_skill_meta(runenv: &dyn RunEnv, dir: &Path) -> Option<SkillMeta> {
+    let bytes = runenv.read(&dir.join(SKILL_FILE)).await.ok()?;
+    let raw = String::from_utf8(bytes).ok()?;
+    let (_n, description, _body) = parse_skill_frontmatter(&raw).ok()?;
+    let name = dir
+        .file_name()
+        .map(|s| s.to_string_lossy().into_owned())
+        .unwrap_or_default();
```
khj809: I'm curious why the directory name segment is preferred over the name from frontmatter (which is silently discarded).

src/skill/skill.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+        out.push('\n');
+    }
+    if let Some(root) = skill_root {
+        out.push_str(&format!(
+            "**Creating new skills at runtime:** new skills MUST be created \
+             under `{}/<skill_name>/`, with a `SKILL.md` carrying `name:` \
+             and `description:` frontmatter (plus any supporting files).\n",
+            root.display()
+        ));
```
khj809, line 92: I think "creating skills" should be handled by the dedicated tool or another skill (e.g. https://github.com/anthropics/skills/blob/main/skills/skill-creator/SKILL.md). Just adding this system prompt seems too weak to create skills properly. Moreover, dynamically created skills are not applied in the above "Available Skills" section since the system prompt is currently fixed at agent creation time.

In that sense, I'd like to suggest to focus on handling static skills only in this PR, and design more sophisticated approach for dynamic skills in the following PR. How do you think about it? @jhlee525

khj809 (commented):
Sorry for the late review..

selloriwoo:
It took me a little while to understand it, but I think the overall direction looks good.

selloriwoo (approved):

src/skill/skill.rs
grf53: Thank you for pointing this out.

Prioritizing the directory name over frontmatter was not intended; rather, I believe the value of frontmatter should be prioritized.

Since I do not think it is necessary to cover cases where retrieving the frontmatter name fails, it seems that retrieving the directory name itself is unnecessary.

Please let me know if you have any other opinions.

khj809: I agree. If the frontmatter parsing fails, we should consider the provided SKILL.md is not spec-compliant and should return the explicit error.

src/agent/builder.rs
grf53: Since the API for adding pre-filled files to RunEnv is completely separate from skills (it can be used to pre-load files unrelated to skills), I believe it is meaningful to retain that API.

However, the fact that a skill might be added but the corresponding file might not be inserted needs improvement. The interface you suggested looks good, so I will refer to it.

src/skill/skill.rs
grf53: The system message in question was intended to leave the possibility open as a guideline to "do this if you are creating a skill," but it did not assume the skill creation function itself.

I agree that skill creation should be provided as a separate skill, and I have been considering it as a separate task.

Since the current implementation is not complete, it seems reasonable to focus on enabling the use of static skills in this PR, as you suggested.

I will be able to add this later along with the actual usable skillify feature, and you will be able to refer to the commit history of this PR at that time.

grf53:
## Updates (31db204)

### Removed Functions

**`FileEntry`**
- `read_from(runenv, path)`

**`AgentSpec`**
- `skill_root: Option<PathBuf>`
- `skill_root(root)`

**`AgentBuilder`**
- `skill_root(root)`

**`Agent`**
- `snapshot() -> AgentSpec`
- `skill_root() -> Option<&Path>`
- `discover_skills() -> Vec<SkillMeta>` (async)

**in `skill.rs`**
- `discover_skills(runenv, &[PathBuf])`
- `discover_new_skills(runenv, root, declared)`
- `read_skill_meta`
- `render_skills_table(skills: &[SkillMeta], skill_root: Option<&Path>)`
  -> `render_skills_table(skills: &[SkillMeta])`

**in `rt.rs`**
- `scan_files_recursive` helper

### Classes
<details>
<summary>Class diagram</summary>

```mermaid
classDiagram
    class FileEntry {
        +PathBuf path
        +Bytes content
        +new(path, content)
        +write_to(runenv)
    }

    class SkillMeta {
        +String name
        +String description
        +PathBuf dir
        +skill_md_path()
    }

    class AgentSpec {
        +Vec~FileEntry~ files
        +Vec~PathBuf~ skills
        +file(FileEntry)
        +files(iter)
        +skill(dir, entries)
    }

    class AgentBuilder {
        +file(FileEntry)
        +files(iter)
        +skill(dir, entries)
        +build()
    }

    class Agent {
        +AgentSpec spec
        +skills() &[PathBuf]
        +spec() &AgentSpec
    }

    AgentSpec o-- "*" FileEntry : declares
    AgentBuilder ..> AgentSpec : builds
    AgentBuilder ..> Agent : produces
    Agent --> AgentSpec : holds
    Agent ..> FileEntry : materialises on first run
    AgentSpec ..> SkillMeta : parsed at build
```
</details>

src/skill/skill.rs
grf53: See the [updates](https://github.com/brekkylab/ailoy/pull/397#issuecomment-4449760644) by this decision.

khj809 (approved):
Thanks for the comprehensive fixes. Looks good!

PR #399, opened 2026-05-12
Move tool impl into tool

jhlee525:


src/lib.rs
khj809: How about applying the same approach to lang_model?

jhlee525: ok if it looks natural, i'll do it

khj809 (approved):

PR #400, opened 2026-05-12
feat: add temperature and top_k sampling params to AgentSpec

jhlee525:
## Summary
- Add `temperature` / `top_k` fields to `AgentSpec` (with matching builder setters on `AgentBuilder`).
- Introduce `LangModelOptions` and thread per-call sampling params through `LangModel::run`.
- Forward params in each API marshal where supported: Anthropic and Gemini emit both; ChatCompletion and OpenAI Responses emit `temperature` only (`top_k` is not in those specs, silently ignored).

Grammar / structured output is intentionally deferred — Anthropic has no native equivalent, so it needs a separate design pass.

## Test plan
- [x] `cargo check --all-features --tests`
- [x] `cargo test --lib lang_model_impl` (17 passed, 4 ignored due to missing API keys)
- [x] Manual smoke test against a real provider with non-default temperature/top_k


src/agent/rt.rs
```diff
@@ -63,6 +63,8 @@ impl AgentState {
 pub struct Agent {
     model: LangModel,
 
+    lm_options: LangModelOptions,
```
khj809: I remember that we decided to use `model` instead of `lm`, so should we use `model_options` here?

src/lang_model/rt.rs
```diff
@@ -22,6 +22,18 @@ pub(crate) struct LangModelRequest<'a> {
...
+    pub top_k: Option<u64>,
+}
+
+/// Per-call sampling parameters supplied by the agent on top of the static
+/// [`LangModelProviderElem`] configuration.
+#[derive(Clone, Debug, Default)]
+pub struct LangModelOptions {
+    pub temperature: Option<f64>,
+    pub top_p: Option<f64>,
+    pub top_k: Option<u64>,
```
khj809: It's good to see this struct has resurrected. But why `max_tokens` does not belong to this struct?

src/agent/rt.rs
jhlee525: ok I'll change it.

src/lang_model/rt.rs
jhlee525: I moved `max_tokens` into the `LangModelOptions`

khj809:
@jhlee525 I can't see the changes you mentioned. Did you push your changes?

nuri-yoo (commented):
Empirically, `gpt-5-mini` rejects `temperature`/`top_p` with HTTP 400
(`"Unsupported parameter: 'temperature' is not supported with this model."`).

Affected OpenAI reasoning models:
`o1`, `o1-mini`, `o1-pro`, `o3`, `o3-mini`, `o3-pro`, `o4-mini`,
`gpt-5`, `gpt-5-mini`, `gpt-5-nano`, `gpt-5-pro`.

Is reasoning-model support intentionally out of scope right now? The codebase signals both ways:

- `OpenAIMarshal` already round-trips `reasoning` items (`openai.rs:92` marshal, `:327` unmarshal), `Message.thinking` + `signature` fields exist
- but no `reasoning_effort` / `verbosity` exposure, and this PR adds `temperature`/`top_p` without addressing that those are 400-rejected on the same reasoning models

If reasoning models are in scope, would prefer folding two more fields into `LangModelOptions` in this PR:

- `reasoning_effort: Option<String>` — {minimal, low, medium, high} ([docs](https://developers.openai.com/api/docs/guides/reasoning))
- `verbosity: Option<String>` — {low, medium, high} ([docs](https://developers.openai.com/cookbook/examples/gpt-5/gpt-5_new_params_and_tools))

Verified both work on gpt-5-mini against raw API (effort scales reasoning_tokens 0 → 64; verbosity scales output length 483 → 2218 → 3993 chars). `OpenAIMarshal` inserts nested keys (`reasoning.effort`, `text.verbosity`); other marshalers ignore — same pattern as `top_k` on OpenAI today. `Option<String>` over enums since OpenAI narrows allowed values per model (e.g. non-reasoning models only accept `verbosity=medium` — [Azure docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/reasoning)).

khj809:
Here's kuse-cowork's solution of handling temperature on openai: https://github.com/kuse-ai/kuse_cowork/blob/main/src-tauri/src/llm_client.rs#L475-L494

nuri-yoo:
> Here's kuse-cowork's solution of handling temperature on openai: https://github.com/kuse-ai/kuse_cowork/blob/main/src-tauri/src/llm_client.rs#L475-L494

For comparison, I searched how openwork handles this issue.
Opencode (which openwork relies on) uses 'models.dev' as metadata (https://github.com/sst/models.dev)
They take the opposite route. Gather capability metadata instead of hardcoded matching. For this
PR, silent drop (the way how kuse-cowork does) seems reasonable. Maybe we can utilize models.dev later considering fast update.

grf53:
> 'models.dev' as metadata (https://github.com/sst/models.dev)

It seems like a good approach for future application.

This is unrelated to the PR, but it is a bit disappointing that its model definitions don't include a list of selectable 'reasoning efforts' or the range of the `ThinkingBudget` (of Gemini).

src/agent/spec.rs
```diff
@@ -58,6 +58,38 @@ pub struct AgentSpec {
...
+    /// Honoured by all supported providers (Anthropic, Gemini, OpenAI).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub top_p: Option<f64>,
+
+    /// Top-k sampling parameter passed to the language model on every call.
+    /// Rarely needed in practice — see [`temperature`](Self::temperature).
+    /// Only honoured by providers that support it (e.g. Anthropic, Gemini);
+    /// silently ignored by providers that do not (e.g. OpenAI).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub top_k: Option<u64>,
```
grf53: While looking at the `LangModelOptions` struct, I wondered if it was really necessary for `LangModel`-related settings to be included so 'flatly' within the `AgentSpec`.

I mean, we could consider placing them within a child object(with name `"model_options"` or sth) to adjust the priority of concerns.

Do you have some opinions on this?

jhlee525:
> > 'models.dev' as metadata (https://github.com/sst/models.dev)
> 
> It seems like a good approach for future application.
> 
> This is unrelated to the PR, but it is a bit disappointing that its model definitions don't include a list of selectable 'reasoning efforts' or the range of the `ThinkingBudget` (of Gemini).

Yes, it was exist in main, but it dropped in `develop` since it started with "minimum reproducible" of ailoy.

We can add it later

src/agent/spec.rs
jhlee525: Ok it seems reasonable. I think it can be applied.

khj809:
@jhlee525 What's the conclusion about openai's reasoning models? If you're planning to fix in the following PR, I'm gonna approve it right now.

jhlee525:
For now, in this PR, let's handle it with a hardcoded approach like Kuse Cowork does.

nuri-yoo (approved):

src/lang_model/impl/api/chat_completion.rs
```diff
@@ -32,6 +32,18 @@ impl LangModelProvider {
...
+pub(super) fn is_openai_reasoning_model(model: &str) -> bool {
+    let m = model.to_ascii_lowercase();
+    if m.starts_with("gpt-5") {
+        return true;
+    }
+    // OpenAI reasoning families share an `o<digit>` prefix: o1, o3, o4, ...
+    let bytes = m.as_bytes();
+    bytes.len() >= 2 && bytes[0] == b'o' && bytes[1].is_ascii_digit()
+}
+
```
khj809: nit: would be better to locate it under `openai.rs`

khj809 (approved):

src/agent/spec.rs
grf53: @jhlee525 are you gonna apply this? It's not necessary, so we could just defer.

jhlee525: Applied

grf53 (approved):
LGTM

src/agent/spec.rs
grf53: Looks more neat!

PR #401, opened 2026-05-12
feat(sandbox): microsandbox 0.3→0.4 upgrade + Sandbox::fork()

khj809:
## Summary

- **microsandbox 0.3 → 0.4 upgrade**: bumps the dependency and fixes a regression where filesystem writes were not persisted across stop/start cycles. Root cause: 0.4 changed the writable upper layer from a host directory (virtiofs) to an ext4 block device (`upper.ext4`); dirty pages live in the guest kernel's page cache and must be flushed explicitly via `sync` before `stop_and_wait()`.
- **`Sandbox::fork()`**: new method that copies a stopped sandbox's complete disk state (including system paths, `apt install` side-effects, and `/workspace`) into a brand-new sandbox. Internally uses microsandbox 0.4's snapshot API (`SandboxHandle::snapshot` + `SandboxBuilder::from_snapshot`). The intermediate snapshot artifact is always removed after fork, whether it succeeds or fails.

## Changed files

| File | Change |
|---|---|
| `Cargo.toml` | `microsandbox = "0.3"` → `"0.4"` |
| `Cargo.lock` | updated transitive deps |
| `src/runenv/sandbox.rs` | `sync` before every write-path `stop_and_wait()`; `Sandbox::fork()`; `create_from_snapshot()` helper; 4 new tests |

## Test plan

- [x] `test_fork_running_returns_error` — fork on running sandbox returns `Err`
- [x] `test_fork_copies_workspace_file` — file written before fork is visible in child
- [x] `test_fork_is_isolated` — child mutation does not affect source
- [x] `test_fork_snapshot_cleaned_up` — snapshot artifact is removed after fork
- [x] All 15 pre-existing sandbox tests still pass (including the two 0.4 regressions: `test_filesystem_persists_across_stop_start`, `test_sequential_ops_on_shared_arc`)

Run with:
```
cargo test --features sandbox runenv::sandbox
```


grf53 (approved):
LGTM

src/runenv/sandbox.rs
```diff
@@ -348,6 +357,7 @@ impl Sandbox {
             let raw = guard.shell(&script).await;
             Self::handle_exec_result_static(raw, max_output_chars)
         };
+        let _ = guard.shell("sync").await;
```
ljhh-0611, line 360: ~~Is doing sync again in this shell function not considered a recursive sync?~~
Ah, all of guard is inner sandbox, right?

src/runenv/sandbox.rs
```diff
@@ -416,6 +429,62 @@ impl Sandbox {
...
     }
 
+    /// Copy this sandbox's full disk state into a new sandbox.
+    ///
+    /// The source must be stopped. Internally uses microsandbox's snapshot API
+    /// (`upper.ext4` capture + `from_snapshot` boot) so that all filesystem
+    /// changes — including system paths modified by `apt install` — are
+    /// included. The snapshot artifact is removed after the fork completes,
+    /// whether the fork succeeded or failed.
+    pub async fn fork(&self, new_cfg: SandboxConfig) -> anyhow::Result<Sandbox> {
```
ljhh-0611, line 439: ~~Instead of syncing before every write function stop, couldn’t we call sync at the very beginning of the fork function?~~

src/runenv/sandbox.rs
khj809: Yes 🙂

ljhh-0611 (approved):

PR #403, opened 2026-05-13
feat(langmodel): add structured output capability

khj809:
## Summary

Adds structured output support (`ResponseFormat`) across all four provider backends, wired end-to-end from `AgentSpec` down to the provider wire format.

### `ResponseFormat` type (`lang_model/rt.rs`)

- New `ResponseFormat` enum with a single variant `JsonSchema(Value)`.
- `ResponseFormat::json_schema(schema)` validates the schema against JSON Schema Draft 7 via the `jsonschema` crate at construction time and returns `Err` on structurally invalid schemas (e.g. `"type": 123`).
- Derives `Serialize`, `Deserialize`, `PartialEq` for serde round-trips, and implements `schemars::JsonSchema` manually so the field is visible in generated API schemas.
- Added `response_format: Option<ResponseFormat>` to `LangModelOptions` and `LangModelRequest`.

### `AgentSpec` wiring (`agent/spec.rs`, `agent/rt.rs`)

- Sampling parameters (`max_tokens`, `temperature`, `top_p`, `top_k`) and `response_format` are carried via `model_options: Option<LangModelOptions>` on `AgentSpec` — not as flat fields.
- Builder methods (`AgentSpec::response_format`, `AgentBuilder::response_format`) lazily initialise `model_options` via `get_or_insert_with`.
- `response_format` is serialised normally with `#[serde(skip_serializing_if = "Option::is_none")]` so agent configs round-trip correctly.
- `AgentProvider` passes `model_options` through to `LangModelOptions` on every run.

### `ResponseSchemaMarshal` trait (`lang_model/impl/response_format.rs`)

Provider-specific schema adaptation is decoupled from `ResponseFormat` itself via the `ResponseSchemaMarshal` trait:

- **Default implementation** — recursively adds `"additionalProperties": false` to every object sub-schema that omits it. Required by both Anthropic and OpenAI strict mode. Covers `properties`, `items`, `prefixItems`, `anyOf`/`oneOf`/`allOf`, `\$defs`, and `definitions`.
- **Gemini override** — strips JSON Schema keywords Gemini does not accept (`additionalProperties`, `\$schema`, `\$defs`, `definitions`) since its `responseSchema` field only accepts a restricted OpenAPI 3.0 subset. Does not add `additionalProperties`.

### Provider wire formats

| Provider | Field | Format |
|---|---|---|
| OpenAI Responses API | `body.text.format` | `{type: "json_schema", name: "response", strict: true, schema: ...}` |
[...]

PR #404, opened 2026-05-13
feat(lang_model): opt-in structured outputs across OpenAI, Anthropic, Gemini adapters

nuri-yoo:
## Summary

Add opt-in JSON-schema-shaped output enforcement across all four provider adapters (OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, Gemini generateContent). When the caller does nothing, request bodies are byte-identical to before this change.

- `ResponseFormat::JsonSchema { name, schema, strict }` is a new enum exposed from `lang_model::rt`. The single shape covers all four adapters; each adapter translates it into its provider-specific body field.
- `AgentSpec.response_format: Option<ResponseFormat>` (`#[serde(skip)]` + `#[schemars(skip)]`) carries the choice through the spec. `AgentBuilder::response_format(name, schema_value)` is the convenience setter (`strict = true` by default).
- `LangModel::run` is preserved as a wrapper over a new `run_with_response_format` so existing call sites and external users of the public `LangModel` API are not affected.
- `Agent` stores the option once at construction and forwards it on every `run_with_response_format` call from the agent runtime.

## Why

Instruction-only "respond in this JSON shape" prompts are not enforced at decode time, so models occasionally output values outside the declared enum (e.g. `"agent":"fallback"` instead of one of the declared agent names). Provider Structured Outputs APIs constrain the decoder and eliminate those escapes — but only if the adapter actually passes the schema through, which ailoy did not.

This PR adds the pass-through. agent-k will use it in a follow-up to remove the ~2% enum escape we measured on the router; other ailoy users get the same lever for any agent that needs structured output.

## Per-provider mapping

- **OpenAI Chat Completions** — `response_format.json_schema.{name,schema,strict}`.
- **OpenAI Responses API** — `text.format.{name,schema,strict}` (different field path from Chat Completions; both are handled).
- **Anthropic Messages** — `response_format.json_schema.{name,schema,strict}` on the request body. Newer Claude models accept this directly; older models return 400 and callers should target a supported model or omit `response_format`.
- **Gemini generateContent** — `generationConfig.responseMimeType = "application/json"` plus `generationConfig.responseSchema = <schema>`, coexisting with `maxOutputTokens` in the same `generationConfig` object.

## Backward compatibility

- `LangModelRequest` is `pub(crate)`; the added field is invisible to external users.
- `AgentSpec` field is `#[serde(skip)]` and `#[schemars(skip)]`, so saved/restored specs round-trip unchanged.
- `LangModel::run(...)` retains the same signature and behaviour; `run_with_response_format` is additive.
- When `response_format` is `None` (the default), every adapter takes the existing code path and produces the same body bytes as before.

## Test plan

- [x] `cargo build` — no warnings (including dead-code).
- [x] `cargo test --lib` — 165 existing tests pass with the change applied. (Two pre-existing failures on `develop` HEAD — `test_context_manager_truncates_tool_results_when_threshold_exceeded` and `test_read_relative_path_rejected` — reproduce on a clean checkout and are unrelated to this PR.)
- [x] New `test_response_format_enforces_enum` (gated by `OPENAI_API_KEY`): feeds a Korean meta/identity-shaped adversarial prompt with an enum-constrained schema and asserts the response parses as the expected struct with a value inside the enum set. Passes.

## Out of scope (follow-up)

- Anthropic older-model fallback via the tool-use route (the workaround that worked before `response_format` shipped). Callers targeting those models can keep using instruction-only schemas for now.
- A `strict = false` lax mode setter. The current setter defaults to `strict = true`; adding the lax variant is a one-method addition if a real use case appears.
- Streaming responses with schema enforcement (the integration test exercises the non-streaming path).

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use crate::{
     agent::{Agent, AgentProvider, AgentSpec, ContextManager},
+    lang_model::ResponseFormat,
     message::Message,
     runenv::RunEnv,
     tool::ToolDesc,
@@ -140,6 +141,18 @@ impl AgentBuilder {
     /// exceeds `spec.max_input_tokens`, preserving the most recent turns.
     pub fn context_manager(mut self, spec: ContextManager) -> Self {
         self.context_manager = Some(spec);
+        self
+    }
+
+    /// Constrain the model's output to a JSON schema (opt-in, strict).
+    /// `schema` is the raw JSON Schema value (e.g. `schemars::schema_for!(T)`).
+    /// When not called, the request body is unchanged.
+    pub fn response_format(mut self, name: impl Into<String>, schema: serde_json::Value) -> Self {
+        self.spec.response_format = Some(ResponseFormat::JsonSchema {
+            name: name.into(),
+            schema,
+            strict: true,
+        });
         self
     }
 
--- a/src/agent/rt.rs
+++ b/src/agent/rt.rs
@@ -72,6 +72,10 @@ pub struct Agent {
     pub state: AgentState,
 
     context_manager: Option<ContextManager>,
+
+    /// Carried from [`AgentSpec::response_format`]; passed to every
+    /// `LangModel::run_with_response_format` call. `None` = unconstrained.
+    response_format: Option<crate::lang_model::ResponseFormat>,
 }
 
 impl Agent {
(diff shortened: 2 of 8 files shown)


nuri-yoo:
Closing in favor of #403 by @khj809, which covers the same area more broadly. This branch is left for reference if any specific piece (e.g. the per-provider mapping notes in the description, or the meta/identity-shaped integration test) turns out to be useful there.

PR #407, opened 2026-05-18
Runenv revised

jhlee525:
## Summary

Redesign `RunEnv` around a two-trait split and a minimal `exec`-based surface, then migrate the tool/agent stack onto it.

- **`Container` + `Console` split.** `Container` owns the boot/shutdown lifecycle and returns a `Console` handle on `boot()`. `RunEnv` is a cheap-to-construct, cheap-to-clone wrapper that lazily boots the container on the first `get()`, hands out `Arc<RunEnvHandle>` clones that share one booted instance, and shuts the container down when the last handle drops.
- **Minimal trait surface.** `Console` only requires `get_os` + `exec`. `exec_shell`, `get_cwd`, `read`, and `write` have default implementations that shell out via `exec` (base64 round-trip for I/O, here-docs for `write`, `pwd` / `(Get-Location).Path` for `get_cwd`). Backends override for efficiency where it matters — most don't need to. The old `stat`/`FSEntry` surface is gone; `grep` and `glob` builtins now shell out instead (`rg` with `grep -E -r -I` fallback; `zsh` for globstar on macOS where bash 3.2 lacks it).
- **Tool/agent migration.** `ToolFunc::call` now takes an owned `Arc<RunEnvHandle>`; tool streams become `'static` and the HRTB on the trait disappears. `AgentState::runenv` is the cloneable v2 `RunEnv`. The VM only boots when a tool actually runs (lazy boot inside `execute_tool_calls`), and parallel tool calls share a single VM via the cached handle.
- **`SharedSandbox` removed.** Sharing a sandbox between agents is now just `RunEnv::clone()`. Sandbox-feature tests updated to `RunEnv::sandbox(...)` + `.get().await`.

See the module docs in [src/runenv/mod.rs](src/runenv/mod.rs) for the full design, lifecycle semantics, and an "extending with a new runenv" skeleton.

## Test plan

- [x] `cargo test` (default features) passes
- [x] `cargo test --features sandbox` passes — covers the lazy-boot path and the dropped-handle shutdown path
- [x] Manual: an agent with multiple parallel tool calls boots the VM once (verify in logs / via sandbox start counter)
- [x] Manual: `read` / `write` round-trip through the default `exec`-only blanket impl on a backend that doesn't override them (binary file, file with single quotes in path, file larger than typical ARG_MAX)
- [x] Manual: `glob`/`grep` builtins on macOS (zsh globstar, `rg` present and absent) and Linux


src/tool/impl/builtins/shell.rs
```diff
@@ -40,10 +40,7 @@ pub fn get_bash_tool_func() -> ToolFunc {
             }
         };
 
-        let Ok(out) = runenv
-            .exec("sh".to_string(), vec!["-c".to_string(), cmd], None)
-            .await
-        else {
+        let Ok(out) = runenv.exec_shell(cmd, None).await else {
```
nuri-yoo, line 45: This call switched from `runenv.exec("sh", ["-c", cmd], None)` to `runenv.exec_shell(cmd, None)` here (commit 2bc5019), but the description on line 17 ("interpreted by sh -c") was not updated to match. The default impl of `exec_shell` branches on `Console::get_os()`:

```
"linux" | "macos" → bash -c <cmd>
"windows"         → powershell -Command <cmd>
```

`LocalConsole::get_os()` returns `std::env::consts::OS`, so on a Windows host with the Local backend the call enters the PowerShell branch. The description still claims sh, so the LM generates POSIX sh syntax from the description while PowerShell ends up interpreting it. That mismatch is the root cause.

Reproduced on PowerShell 7.6 with sh-style commands an LM would likely emit:

```
[ -f /tmp/x ] && echo yes
  exit=1
  stderr: ParserError: Missing type name after '['
  cause: `[` in PowerShell starts a type literal

[[ 1 -eq 1 ]] && echo ok
  exit=1
  stderr: ParserError: Missing type name after '['
  cause: same as above

if [ -d /tmp ]; then echo yes; fi
  exit=1
  stderr: ParserError: Missing '(' after 'if' in if statement
  cause: PowerShell's if syntax is `if (cond) { ... }`; `then` and `fi`
         are not keywords

FOO=bar echo hi
  exit=1
  stderr: 'FOO=bar' is not recognized as a name of a cmdlet, function,
          script file, or executable program
  cause: PowerShell has no one-shot env-var prefix syntax. `FOO=bar` is
         parsed as the first token and looked up as a cmdlet, which fails

echo hello > /dev/null
  exit=0, stdout=""
  cause: PowerShell 7.x doesn't special-case /dev/null. It treats it as
         a real file path, so on Windows `/dev/null` gets created as a
         file. Silent side effect — the riskiest case

true && echo yes
  exit=0, stdout="yes"
  cause: PowerShell 7.x supports &&. 5.x (the default on Windows) does
         not, so this would fail with a parse error there

echo hello
echo $HOME
printf 'a\nb\na\n' | grep a | wc -l
  All three happen to be equivalent. `echo` is an alias for Write-Output,
  `$HOME` is a built-in PowerShell variable, and grep/wc are external
  binaries that happen to be on PATH (absent on Windows native, present
  only when WSL/Git Bash is installed)
```

Categories:

- Syntax-level mismatch: exit_code and stderr surface the error clearly, so the LM can retry (`[`, `[[`, `if/then/fi`, `FOO=bar`)
- Silent semantic divergence: PowerShell parses successfully but does something different from what was intended (`/dev/null` becoming a real file, alias collisions). The LM has no signal to detect this

Suggestion — update the description on line 17 so the LM is aware of the OS-level divergence:

```rust
"description": "Shell command. Interpreted by bash on Linux/macOS and by powershell on Windows. Prefer OS-agnostic syntax (e.g. `mkdir`, simple pipes) when possible."
```

This gives the LM enough information to branch on the host OS when crafting commands.

khj809:
I merged develop to resolve conflicts and use the latest microsandbox.
Also, there was an issue that the sandbox unit tests persist msb instances even the config is set as `persist: false`, so fixed it as well.

src/tool/impl/builtins/shell.rs
jhlee525: Thank you for pointing out. I made changes

- `bash` tool -> `shell` tool
- running `bash` -> running `sh` (in lightweight linux (like busybox), bash may not exists)

jhlee525:
@khj809 Thank you for your works!

src/tool/impl/builtins/shell.rs
nuri-yoo: Cleaner than I suggested. Verified all call sites are updated and tests pass on the latest head.

nuri-yoo (approved):
LGTM

src/runenv/mod.rs
```diff
@@ -110,34 +171,248 @@ pub struct ExecResult {
...
+    fn drop(&mut self) {
+        let machine = self.machine.clone();
+        let state = self.state.clone();
+        tokio::spawn(async move {
+            // Hold the state lock through shutdown so concurrent `get()` calls
+            // block until we transition back to `Idle`.
+            let mut s = state.lock().await;
+            machine.lock().await.shutdown().await;
+            *s = RunEnvState::Idle;
+        });
```
khj809: This may panic when there's no tokio runtimes ("there is no reactor running").
As I already implemented in Sandbox's `Drop`, suggest to use `std::thread::spawn` to spin up a temporary runtime in a separate thread.

src/runenv/sandbox.rs
```diff
@@ -275,295 +153,221 @@ impl Sandbox {
...
-        let result = {
-            let raw = guard.shell(&script).await;
-            Self::handle_exec_result_static(raw, max_output_chars)
-        };
-        let _ = guard.shell("sync").await;
-        let _ = guard.stop_and_wait().await;
-        result
+/// `Console` wrapping a running microsandbox VM.
+pub struct SandboxConsole {
+    inner: MsbSandbox,
```
khj809, line 329: Since `inner` is not Mutex anymore, it will allow running multiple commands at once on parallel tool calling. Is that an intended behavior?

src/runenv/local.rs
```diff
@@ -1,30 +1,47 @@
...
+    pub fn new() -> Self {
+        Self {}
+    }
 }
 
-#[cfg(not(unix))]
-fn owner_permission(meta: &std::fs::Metadata) -> u8 {
-    if meta.permissions().readonly() { 4 } else { 6 }
+#[async_trait]
+impl Container for Local {
```
khj809: Since "Container" is often used as Docker/OCI container, it's quite awkward for me to implement `Container` for `Local`, which is definitely not a container environment.
Since you used `machine` for `dyn ContainerDyn`, I think `Machine` can also be a good candidate.

src/runenv/sandbox.rs
jhlee525: I think actually using a `Mutex` would prevent parallel tool calling.

A `Mutex` is a thing to block concurrent access to the inner state, thereby preventing parallel execution.

In real use cases, SandboxConsole will be wrapped into `RunEnvHandle`, and it is `Sync`

src/runenv/local.rs
jhlee525: Ok i'll change it

src/runenv/mod.rs
jhlee525: Ok i'll change it

src/runenv/mod.rs
```diff
@@ -110,34 +171,258 @@ pub struct ExecResult {
...
+        let (program, args) = match self.get_os() {
+            "linux" | "macos" => ("sh".to_string(), vec!["-c".to_string(), script]),
+            "windows" => (
+                "powershell".to_string(),
+                vec!["-Command".to_string(), script],
+            ),
+            other => anyhow::bail!("exec_shell: unsupported OS '{other}'"),
+        };
+        self.exec(program, args, timeout).await
+    }
```
selloriwoo, line 219: Switched the Windows path from `-Command` to `-EncodedCommand` (UTF-16LE base64).

The current internal callers (`get_cwd` / `read` / `write`) are already safe with `-Command`, but `exec_shell` is a public entry point that accepts arbitrary external scripts. This change is mainly for better handling and defense against multiline input, mixed quoting, and CMD metacharacters.

Only the input transport mechanism changed, the semantics of the return values (e.g. stdout / exit code) remain unchanged.

```
    /// Run `script` through the system shell
    ///
    /// `sh -c` on Linux/macOS, `powershell -EncodedCommand` on Windows.
    ///
    /// On Windows the script is transported as UTF-16LE base64 to bypass
    /// CMD/PowerShell argv quoting; quotes, newlines, and metacharacters in
    /// `script` pass through verbatim. Output encoding (BOM, code page) is
    /// unaffected and must still be handled by callers if relevant.
    async fn exec_shell(&self, script: String, timeout: Option<u64>) -> anyhow::Result<ExecResult> {
        let (program, args) = match self.get_os() {
            "linux" | "macos" => ("sh".to_string(), vec!["-c".to_string(), script]),
            "windows" => {
                let utf16le: Vec<u8> = script
                    .encode_utf16()
                    .flat_map(|u| u.to_le_bytes())
                    .collect();
                let b64 = STANDARD.encode(&utf16le);
                (
                    "powershell".to_string(),
                    vec![
                        "-NoLogo".to_string(),
                        "-NoProfile".to_string(),
                        "-EncodedCommand".to_string(),
                        b64,
                    ],
                )
            }
            other => anyhow::bail!("exec_shell: unsupported OS '{other}'"),
        };
        self.exec(program, args, timeout).await
    }
```

src/runenv/sandbox.rs
khj809: What I worried is the situation running a command which depends on the previous command (for example, `echo "asdf" > asdf.txt` and `cat asdf.txt`). If you think this is not a practical scenario, then we can use `inner` without mutex for now.

src/runenv/mod.rs
jhlee525: it seems good, I applied.

However, I intentionally dropped `-NoProfile`, since in local environment the profile should be overrided. (In linux / macos version, it does)

selloriwoo (approved):

src/runenv/sandbox.rs
jhlee525: keeping `inner: Mutex<MsbSandbox>` is not related to this problem.

1. If you meant race condition in tool calling, mutex cannot resolve this problem (also happens)
2. If you meant parallel access, that's what rust's `Sync` does. In here, dropping `Mutex` does not collapse `Sync`.

Also, `Console` is just trait so it does not related to `Sync` implementation. The user of `Console` can make it syncable even `Console` is not `Sync`, by

```
let a = Arc<Mutex<dyn Console>>
```

khj809:
Unit tests are still failing, and I'm trying to fix it.

khj809 (approved):
I fixed the failed tests and added more tests for uncovered logics.
The current latest version (v0.4.6) of microsandbox does not flush pending writes on stop, and this is fixed in https://github.com/superradcompany/microsandbox/pull/746. In `Cargo.toml`, microsandbox version is pinned to the latest commit of that PR.
However, it still downloads v0.4.6's prebuilt `agentd` binary which does not apply the fix, so there's a workaround to call `sync -f` before stop.

jhlee525:
Thank you for your help!

PR #410, opened 2026-05-20
revise glob and grep tool using `exec_shell`

jhlee525:


src/tool/impl/builtins/glob.rs
```diff
@@ -12,21 +12,60 @@ fn sh_single_quote_inner(s: &str) -> String {
...
+    let mut i = 0;
+    while i < chars.len() {
+        let c = chars[i];
+        if c == '*' {
+            if i + 1 < chars.len() && chars[i + 1] == '*' {
+                out.push_str(".*");
+                i += 2;
+                if i < chars.len() && chars[i] == '/' {
+                    i += 1;
+                }
```
khj809: `**` is translated to `.*`, which matches any sequence of characters — not just complete path segments. Consider the pattern `**/foo.rs`:

`glob_to_ere("**/foo.rs")` -> `.*foo\.rs`

This incorrectly matches `./barfoo.rs` because `.*` greedily consumes `bar` and then `foo\.rs` matches the rest. `**` semantics require that a path separator precede the matched segment (or the match is at the root). The correct ERE should be:

`(.*\/|)foo\.rs` -> zero or more path segments + separator, or at root

src/tool/impl/builtins/glob.rs
```diff
@@ -12,21 +12,60 @@ fn sh_single_quote_inner(s: &str) -> String {
...
+            out.push('[');
+            i += 1;
+            if i < chars.len() && (chars[i] == '!' || chars[i] == '^') {
+                out.push('^');
+                i += 1;
+            }
+            while i < chars.len() && chars[i] != ']' {
+                out.push(chars[i]);
+                i += 1;
+            }
```
khj809: This loop terminates at the first `]` unconditionally. It fails for:
1. POSIX character classes — `[[:alpha:]]`: the inner `]` closes the bracket early, producing the invalid ERE `[[:alpha:]` followed by a stray `]`.
2. Bracket-first syntax — `[]abc]`: a `]` as the first character inside `[` is valid POSIX, but here it immediately closes the class, producing `[]` (likely an ERE error on most implementations).
3. Unterminated class — `[abc` with no closing `]`: the outer if guard prevents a push, leaving an unclosed `[` in the ERE, which causes `grep -E` to return a parse error.

src/tool/impl/builtins/glob.rs
```diff
@@ -12,21 +12,60 @@ fn sh_single_quote_inner(s: &str) -> String {
...
+            if i < chars.len() {
+                out.push(']');
+                i += 1;
+            }
```
khj809, line 63: When a pattern contains an unterminated `[` (e.g. `[abc` without a closing `]`), the loop exhausts the input without finding `]`, so the closing bracket guard silently produces an unclosed `[` in the ERE:

`glob_to_ere("[abc")` -> "[abc"

`grep -E` treats an unclosed `[` as a parse error and exits with code 2, which the caller surfaces as a tool-level `"phase": "io"` error — indistinguishable from a real I/O failure. The user gets a confusing error with no indication that their pattern is malformed.

The fix is straightforward: if the inner loop exhausts input without finding `]`, either push a synthetic `]` to close the class, or detect and reject the pattern early in the validation phase with `"phase": "validation"`.

src/tool/impl/builtins/grep.rs
```diff
@@ -142,13 +145,21 @@ pub fn get_grep_tool_func() -> ToolFunc {
             });
         }
 
+        let os = runenv.get_os();
+        if os != "linux" && os != "macos" {
+            return crate::to_value!({
+                "error": format!("grep: unsupported OS '{os}'"),
+                "phase": "io",
```
khj809: I think this should be `validation` rather than `io`, since it's still validating OS capability.

src/tool/impl/builtins/grep.rs
```diff
@@ -10,12 +10,15 @@ use crate::{
 const DEFAULT_LIMIT: usize = 1000;
 const MAX_OUTPUT_CHARS: usize = 200_000;
 
-/// Probe for ripgrep by running `rg --version`. A non-zero exit (or spawn
-/// failure, which `Local::exec` surfaces as `exit_code = -1`) is treated as
-/// "not available".
+/// POSIX-safe shell quoting: wrap `s` in single quotes, escaping embedded `'`.
+fn sh_quote(s: &str) -> String {
+    format!("'{}'", s.replace('\'', "'\\''"))
+}
```
khj809, line 16: Duplicated function `sh_single_quote_inner` exists in glob.rs.

jhlee525: I think it's one-liner utility and ok to make it duplicate for loosen their relationship.

khj809 (approved):

PR #412, opened 2026-05-21
feat(tool): add web_fetch builtin (single)

nuri-yoo:
## Summary

Adds a `web_fetch` builtin that fetches a URL and returns the readable body. Complements `web_search`: search gives metadata, fetch gives the actual page so a caller can verify what a snippet claimed.

## Surface

- `AgentSpec::web_fetch_tool()` / `AgentBuilder::web_fetch_tool()` add the tool to a spec. Opt-in, like `web_search_tool`; not included in `system_tools()`.
- One URL per call. ailoy already parallelises tool calls, so the model can issue several `web_fetch` calls in one turn without the tool needing its own batch path.
- Returns `{url, status, title, content_type, body, body_length_total, next_offset, complete, retrieved_at}`.

## Behavior

- `format` parameter, default `"text"`:
  - `"text"` — visible text only, no markup. Smallest token cost.
  - `"markdown"` — CommonMark output, keeps headings, lists, tables, and link targets.
  - `"html"` — response body unchanged.
- HTML responses go through `html-to-markdown-rs`, which parses with `html5ever` and drops `<script>`/`<style>`/inline `<svg>`/etc. Non-HTML responses (JSON, plain text, …) pass through verbatim regardless of `format`.
- Title comes from `metadata.document.title` with `og:title` as a fallback.
- `skip_images=true` so inline base64 data-URIs don't dominate the body.
- Response body cap: 2 MiB downloaded, streamed chunk by chunk.
- Readable body clamped to 30 KiB by default (hard cap 60 KiB). `offset` pages past the first window on the same URL; `next_offset` echoes the next char index.
- URL length cap: 2048 characters. Request timeout: 10 seconds.
- Per-host rate limit: one request per second.

## Tests

Default unit tests (run on every `cargo test`):

- `slice_body` math (ASCII + unicode boundaries) and the `next_offset` contract that callers rely on.
- `BodyFormat::parse` accepts `text`/`plain`/`markdown`/`md`/`html`/`raw`.
- HTML conversion: `text` strips `<script>` + `<style>`, `markdown` preserves headings + link targets, `html` returns the body verbatim, non-HTML content types pass through.
- Title extraction prefers `<title>` and falls back to `og:title`.
- Tool descriptor shape.

Network-required tests, `#[ignore = "requires network"]` to match the convention used by `web_search` (run with `cargo test -- --ignored`):

- Single-URL `format="text"` fetch against `https://example.com/` — HTTP 200, body contains `Example Domain`, `retrieved_at` populated.
- Same endpoint with `format="html"` — body contains raw `<html…>` markup.

## Notes

- New dependency: `html-to-markdown-rs = "3.4"` with the `metadata` feature. This is what powers the `text`/`markdown` modes and the title extraction. MSRV is `1.85` (the crate is edition 2024). One transitive cost: a second copy of `html5ever` (0.39 alongside `scraper`'s 0.35) gets compiled until either crate moves.
- The factory closes over a process-wide `WebFetchState` (HTTP client, rate-limit ledger), matching how `web_search` shares one `MetaSearcher` across calls.

## Changelog during review

- Dropped robots.txt handling entirely (parser + cache + pre-fetch check), in response to maintainer preference. ailoy follows the host's response and rate limits per-host.
- Replaced the hand-written `walk()` flattener with `html-to-markdown-rs` and added the `markdown` format option. The earlier description's HTML→text conversion mechanics are now an implementation detail.
- Dropped the `urls` batch path. ailoy's parallel tool calling covers the same use case without nudging the model toward batched calls.
- Rewrote `slice_body` to a single `char_indices()` pass that returns `(slice, total_chars, Option<next_offset>)`, dropping the `Vec<char>` copy and the redundant `slice.chars().count()` in `fetch_one`.

### DIFF
--- a/src/tool/impl/builtins/web_fetch.rs
+++ b/src/tool/impl/builtins/web_fetch.rs
@@ -143,14 +143,7 @@ fn convert_with_crate(html: &str, output_format: OutputFormat) -> Converted {
                 .document
                 .title
                 .clone()
-                .or_else(|| {
-                    result
-                        .metadata
-                        .document
-                        .open_graph
-                        .get("title")
-                        .cloned()
-                })
+                .or_else(|| result.metadata.document.open_graph.get("title").cloned())
                 .unwrap_or_default()
                 .trim()
                 .to_string();
@@ -224,11 +217,7 @@ fn slice_body(text: &str, offset: usize, len: usize) -> (String, usize, Option<u
     };
     let e = end_byte.unwrap_or(text.len());
     let slice = text[s..e].to_string();
-    let next_offset = if e < text.len() {
-        Some(end_char)
-    } else {
-        None
-    };
+    let next_offset = if e < text.len() { Some(end_char) } else { None };
     (slice, total_chars, next_offset)
 }
 
@@ -335,7 +324,10 @@ async fn fetch_one(
         Ok(t) => t,
         Err(e) => return crate::to_value!({"url": url_str, "error": e}),
     };
-    let Converted { body: readable, title } = convert(&body, &content_type, format);
+    let Converted {
+        body: readable,
+        title,
+    } = convert(&body, &content_type, format);
     let (slice, total_chars, next_offset) = slice_body(&readable, offset, length);
     let complete = next_offset.is_none();
     let next_offset = match next_offset {
@@ -483,16 +475,11 @@ mod tests {
 
     #[test]
     fn convert_markdown_preserves_structure() {
-        let html =
-            "<html><body><h1>Title</h1><p>See <a href=\"https://example.com/x\">docs</a></p></body></html>";
+        let html = "<html><body><h1>Title</h1><p>See <a href=\"https://example.com/x\">docs</a></p></body></html>";
         let out = convert(html, "text/html", BodyFormat::Markdown);
         // Markdown should keep the heading sigil and the anchor URL.
         assert!(out.body.contains("# Title"), "{}", out.body);
-        assert!(
-            out.body.contains("(https://example.com/x)"),
-            "{}",
-            out.body
-        );
+        assert!(out.body.contains("(https://example.com/x)"), "{}", out.body);
     }
 
     #[test]
@@ -595,5 +582,4 @@ mod tests {
             &body.chars().take(200).collect::<String>()
         );
     }
-
 }
--- a/src/tool/impl/builtins/web_fetch.rs
+++ b/src/tool/impl/builtins/web_fetch.rs
@@ -1,0 +1,698 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use parking_lot::Mutex;
+use reqwest::Client;
+use scraper::{Html, Selector};
+use url::Url;
+
+use crate::{
+    datatype::Value,
+    tool::{ToolDesc, ToolDescBuilder, ToolFunc},
+    tool_func,
+};
+
+const DEFAULT_BODY_CHARS: usize = 30 * 1024;
+const MAX_BODY_CHARS: usize = 60 * 1024;
+const MAX_DOWNLOAD_BYTES: usize = 2 * 1024 * 1024;
+const MAX_URL_CHARS: usize = 2048;
+const REQUEST_TIMEOUT_SECS: u64 = 10;
+const PER_HOST_MIN_INTERVAL: Duration = Duration::from_millis(1000);
+const MAX_BATCH_URLS: usize = 5;
+const USER_AGENT: &str =
+    "Mozilla/5.0 (compatible; ailoy/web_fetch; +https://github.com/brekkylab/ailoy)";
+
+#[derive(Clone)]
+struct WebFetchState {
+    client: Client,
+    last_hit: Arc<Mutex<HashMap<String, Instant>>>,
+    robots: Arc<Mutex<HashMap<String, RobotsRules>>>,
+}
+
+impl WebFetchState {
+    fn new() -> Self {
+        let client = Client::builder()
+            .user_agent(USER_AGENT)
+            .timeout(Duration::from_secs(REQUEST_TIMEOUT_SECS))
+            .build()
+            .expect("reqwest::Client builder cannot fail with these settings");
+        Self {
+            client,
+            last_hit: Arc::new(Mutex::new(HashMap::new())),
+            robots: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+}
+
+#[derive(Default, Clone)]
+struct RobotsRules {
+    disallow: Vec<String>,
+    allow: Vec<String>,
+    // Per RFC 9309 §2.3.1.4: when robots.txt cannot be fetched, allow all.
+    fetch_failed: bool,
+}
+
+impl RobotsRules {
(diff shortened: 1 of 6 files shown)


src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+fn extract_readable(body: &str, content_type: &str) -> String {
+    if content_type.to_ascii_lowercase().contains("html") {
+        html_to_text(body)
+    } else {
+        body.to_string()
+    }
+}
+
+fn slice_body(text: &str, offset: usize, len: usize) -> (String, usize, bool) {
+    let total: Vec<char> = text.chars().collect();
```
khj809: This copies the whole `text` into `Vec<char>`, which doubles the memory usage.
Use `text.char_indices()` to iterate without additional memory allocation.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+    let total_chars = total.len();
+    if offset >= total_chars {
+        return (String::new(), total_chars, true);
+    }
+    let end = offset.saturating_add(len).min(total_chars);
+    let slice: String = total[offset..end].iter().collect();
```
khj809: Related to upper comment, it would be better to slice from `text` directly.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+        Ok(t) => t,
+        Err(e) => return crate::to_value!({"url": url_str, "error": e}),
+    };
+    let title = extract_title(&content_type, &body);
+    let readable = extract_readable(&body, &content_type);
+    let (slice, total_chars, complete) = slice_body(&readable, offset, length);
+    let next_offset = if complete {
+        Value::Null
+    } else {
+        Value::from((offset + slice.chars().count()) as i64)
```
khj809: `slice.chars().count()` iterates over `slice` again, which was already done in `slice_body`.
I think `next_offset` can also be calculated inside `slice_body` during the iteration.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+            }
+        }
+        for a in &self.allow {
+            if !a.is_empty() && path.starts_with(a.as_str()) && a.len() >= best_len {
+                best_len = a.len();
+                best_allow = true;
+            }
+        }
+        best_allow
+    }
```
khj809: According to RFC 9309 §2.2.2, `*` is considered as random string and `$` is considered as end-anchor, but current impl does not consider the following cases:
- `Disallow: /*` should disallow every paths, but `/any/path` is still allowed
- `Disallow: /private/*.pdf` should disallow `/private/doc.pdf`, but it's still allowed
- `Disallow: /search$` should disallow `/search`, but it's still allowed

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+        let line = raw.split('#').next().unwrap_or("").trim();
+        if line.is_empty() {
+            continue;
+        }
+        let (key, value) = match line.split_once(':') {
+            Some((k, v)) => (k.trim().to_ascii_lowercase(), v.trim().to_string()),
+            None => continue,
+        };
+        match key.as_str() {
+            "user-agent" => applies_to_us = value == "*",
```
khj809: Consecutive User-agent lines with no blank line form one record whose rules apply to all listed agents. If `*` appears first, Disallow must block us even though a named bot follows on the next line.
For example, in the following robots.txt, 
```
User-agent: *
User-agent: Googlebot
Disallow: /blocked
```
it should disallow `/blocked` for any user agents, but it's still allowed.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+}
+
+fn parse_robots(body: &str) -> RobotsRules {
+    let mut rules = RobotsRules::default();
+    let mut applies_to_us = false;
+    for raw in body.lines() {
+        let line = raw.split('#').next().unwrap_or("").trim();
+        if line.is_empty() {
+            continue;
+        }
```
khj809: Only the first record matching a given user-agent should be used.
For example, in the following robots.txt,
```
User-agent: *
Disallow: /blocked

User-agent: *
Allow: /blocked
```
it should disallow `/blocked` for any user agents, but it's overwritten by the second rule and allows the path.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+    (y, m, d)
+}
+
+pub fn get_web_fetch_tool_desc() -> ToolDesc {
+    ToolDescBuilder::new("web_fetch")
+        .description(concat!(
+            "Fetch one or more URLs and return readable bodies. HTML is reduced ",
+            "to plain text: script/style/svg/template/noscript subtrees are ",
+            "removed and <a href> anchors render as `text [url]` so outbound ",
+            "links survive. Use this after `web_search` to read the full body ",
```
jhlee525: 1. This assumes that it is paired with a `web_search` tool. While it may be the usual case, but theoretically it is not paired with a `web_search` tool, so I think this assumption does not need to be included in the default behavior. (One can add them on their instruction)
2. The logic explaination above feels too specific and may not be something “users of this tool” need to know. I think it would be sufficient to simply state that the tool returns results in markdown format.

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,698 @@
...
+            "of a result instead of relying on the snippet.\n\n",
+            "Two call shapes:\n",
+            "  - single: pass `url` (string), get one result back.\n",
+            "  - batch: pass `urls` (array of strings) to fetch up to 5 URLs in ",
```
jhlee525: Giving "batched option" to LLM can lead to bias. (They feel pressure to use batch query)

I think we already have parallel tool calling, so it would be good to just use single fetch per one call.

nuri-yoo:
@jhlee525 @khj809 all feedback addressed in lastest commit [b5c3b24](https://github.com/brekkylab/ailoy/pull/412/commits/b5c3b247a9fcdab46d2557a965f5ebddf756bc8e)

src/tool/impl/builtins/web_fetch.rs
```diff
@@ -0,0 +1,575 @@
...
+
+async fn rate_limit_for(state: &WebFetchState, host: &str) {
+    let wait = {
+        let mut m = state.last_hit.lock();
+        let now = Instant::now();
+        let wait = m
+            .get(host)
+            .and_then(|when| PER_HOST_MIN_INTERVAL.checked_sub(now.duration_since(*when)))
+            .unwrap_or(Duration::ZERO);
+        m.insert(host.to_string(), now + wait);
```
khj809, line 53: There's no eviction in `last_hit`. How about removing the old ones using `m.retain()` after insert?

nuri-yoo: Will fix shortly.

khj809 (approved):
Just applied the formatting. LGTM!

jhlee525 (approved):
👍

PR #413, opened 2026-05-21
chore: fix failing context_manager test and silence clippy warnings on develop

nuri-yoo:
## Summary

Cleans up two pre-existing issues on `develop`:

- One unit test (`agent::rt::tests::test_context_manager_truncates_tool_results_when_threshold_exceeded`) was panicking when run with `OPENAI_API_KEY` set in the environment.
- `cargo clippy --lib` was reporting 41 warnings on default features (and 42 with `--all-features`).

After this PR, `cargo test --lib` passes with zero failures and `cargo clippy --lib` reports zero warnings.

The PR is split into three commits so each layer can be reviewed independently.

## Commit 1 — `fix(test): align context_manager truncation test with prod boundary`

`ContextManager::truncate_history` calls `find_preserve_boundary`, which walks history backwards and counts every `User` message it sees, including the new query message that `Agent::run` pushes at the start of the turn.

The test was using `preserve_recent_turns: 1` and the docstring said the boundary would land at `u2`. In practice the boundary landed on `u3` (the just-pushed new query), so `tr_recent` was truncated and the test panicked at `'recent tool result must still be a Value part'`.

Change: bump the test to `preserve_recent_turns: 2` so the boundary actually lands on `u2` as the docstring describes, and spell out in a comment that the just-pushed query counts as a turn. No production code change.

## Commit 2 — `chore(lint): apply cargo clippy --fix and cargo fmt`

Runs `cargo clippy --lib --fix` and then `cargo fmt` on the workspace using the rustc/clippy versions matching CI (rustc 1.93.1). All edits are mechanical idiom upgrades suggested by clippy:

- `needless_borrow`, `redundant_closure`, `manual_filter`, `useless_format`, `single_match`, `unnecessary_lazy_evaluations`, `bind_instead_of_map`, `from_over_into`, `default_constructed_unit_structs`, `unnecessary_lazy_evaluations`, `collapsible_if`.

Diff scope: 11 files, +29 / −37. No runtime behavior changes.

## Commit 3 — `chore(lint): silence remaining clippy warnings`

Manual fixes for the lints `cargo clippy --fix` cannot rewrite automatically:

- `match_like_matches_macro`: rewrite five `Value::` predicate methods (`is_null`, `is_number`, `is_string`, `is_array`, `is_object`), four `Part::` predicates (`is_text`, `is_function`, `is_value`, `is_image`), and five `PartDelta::` predicates as `matches!(...)` invocations.
- `ptr_arg` + `needless_range_loop`: `ContextManager::truncate_history` now takes `&mut [Message]` and walks the preserve window with `iter_mut().take().skip()` instead of range indexing.
- `manual_strip`: `web_search/engines/google.rs::clean_url` uses `strip_prefix("/url?q=")` instead of `starts_with` plus byte-index slicing.
- `module_inception`: `src/message/mod.rs` and `src/skill/mod.rs` get `#[allow(clippy::module_inception)]` on the inner module to keep the established naming.

After this commit `cargo clippy --lib` is clean on default features.

## Verification

- `cargo clippy --lib` → 0 warnings.
- `cargo clippy --lib --all-features` → 0 warnings from non-test code in this PR's scope (the pre-existing web_search engine warnings under `--all-features` were already addressed by commit 2).
- `cargo test --lib` → 234 passed, 0 failed, 16 ignored (ignored = network-required tests, plus a few env-gated sandbox tests).
- `cargo test --lib -- --ignored` → no new failures introduced by this PR.

## Notes

- No new dependencies.
- No public API changes. `ContextManager::truncate_history` widens its argument from `&mut Vec<Message>` to `&mut [Message]`, but the method is `pub(crate)` so this is internal-only.

### DIFF
--- a/src/agent/context.rs
+++ b/src/agent/context.rs
@@ -56,6 +56,16 @@ impl ContextManager {
             0
         };
 
+        // `preserve_from = 0` means "fewer turns than requested — preserve everything":
+        // `.take(0).skip(start_idx)` is always empty, so nothing is truncated.
+        // When `preserve_from > 0` the System message always lands at index 0 and
+        // the oldest preserved User turn is at index >= 1, so start_idx <= preserve_from.
+        debug_assert!(
+            preserve_from == 0 || start_idx <= preserve_from,
+            "start_idx ({start_idx}) > preserve_from ({preserve_from}): \
+             truncation window would overlap the system message"
+        );
+
         // ── Replace Tool messages outside the preserve window with placeholders ────
         for msg in history.iter_mut().take(preserve_from).skip(start_idx) {
             if msg.role == Role::Tool {
--- a/src/agent/context.rs
+++ b/src/agent/context.rs
@@ -42,7 +42,7 @@ impl ContextManager {
     /// post-truncation token estimate that is not yet available here.  For now,
     /// placeholder replacement alone is sufficient to keep the context window
     /// manageable for most workloads.
-    pub(crate) fn truncate_history(&self, history: &mut Vec<Message>) {
+    pub(crate) fn truncate_history(&self, history: &mut [Message]) {
         if history.is_empty() {
             return;
         }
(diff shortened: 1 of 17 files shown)


nuri-yoo:
@grf53 @jhlee525 @khj809 @ljhh-0611 @selloriwoo

khj809:
I fixed the remaining 2 warnings, and added debug assert on `truncate_history`.

khj809 (approved):

nuri-yoo:
> I fixed the remaining 2 warnings, and added debug assert on `truncate_history`.

Thank you for the fix.

jhlee525 (approved):

PR #414, opened 2026-05-21
feat(lang_model): add DeepSeek and Kimi (Moonshot) provider support

jhlee525:
## Summary

Adds first-class DeepSeek and Kimi (Moonshot) support to `LangModelProvider`, including thinking-mode round-tripping for DeepSeek and a payload fix for Kimi.

- **Convenience constructors** in [src/lang_model/impl/api/chat_completion.rs](src/lang_model/impl/api/chat_completion.rs):
  - `LangModelProvider::deepseek(api_key)` → ChatCompletion schema at `https://api.deepseek.com/chat/completions`
  - `LangModelProvider::kimi(api_key)` → ChatCompletion schema at `https://api.moonshot.ai/v1/chat/completions`
- **Default registry wiring** in [src/lang_model/provider.rs](src/lang_model/provider.rs): `LangModelProvider::default()` now also registers `deepseek/*` and `moonshot/kimi-*` when `DEEPSEEK_API_KEY` / `KIMI_API_KEY` is set, matching the existing pattern for OpenAI/Anthropic/Gemini.
- **DeepSeek thinking-mode round-trip** in `ChatCompletionMarshal` / `ChatCompletionUnmarshal`: parse `reasoning_content` from assistant responses into the canonical `Message.thinking` field, and replay it as `reasoning_content` on follow-up turns (required by DeepSeek's thinking endpoint).
- **Kimi payload fix**: drop empty-text parts from the content array and omit the `content` key entirely when no parts remain — Kimi's ChatCompletion backend rejects content arrays that contain an empty-text block.

## Test plan

- [x] `cargo check --lib` passes
- [x] Smoke test with `DEEPSEEK_API_KEY` set — non-thinking and thinking models, multi-turn conversation
- [x] Smoke test with `KIMI_API_KEY` set — basic completion and tool calling

### DIFF
--- a/src/lang_model/provider.rs
+++ b/src/lang_model/provider.rs
@@ -54,8 +54,9 @@ pub enum LangModelProviderElem {
 ///
 /// [`Default::default`] (and therefore [`AgentProvider::new`]) returns a
 /// registry pre-populated from the environment:  registers `openai/*`,
-/// `anthropic/*`, and/or `gemini/*` for every `OPENAI_API_KEY` / `ANTHROPIC_API_KEY`
-/// / `GEMINI_API_KEY` that is set.
+/// `anthropic/*`, `gemini/*`, `deepseek/*`, and/or `moonshot/kimi-*` for every
+/// `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` / `DEEPSEEK_API_KEY`
+/// / `KIMI_API_KEY` that is set.
 /// Use [`new`](Self::new) for an empty registry.
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 #[serde(transparent)]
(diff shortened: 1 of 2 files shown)


src/lang_model/provider.rs
```diff
@@ -76,6 +77,12 @@ impl Default for LangModelProvider {
         if let Ok(key) = std::env::var("GEMINI_API_KEY") {
             p.insert("gemini/*".into(), Self::gemini(key));
         }
+        if let Ok(key) = std::env::var("DEEPSEEK_API_KEY") {
+            p.insert("deepseek/*".into(), Self::deepseek(key));
+        }
+        if let Ok(key) = std::env::var("KIMI_API_KEY") {
+            p.insert("moonshot/kimi-*".into(), Self::kimi(key));
```
khj809: Following the openrouter's provider prefix, suggest to fix as below:
```suggestion
            p.insert("moonshotai/kimi-*".into(), Self::kimi(key));
```

khj809:
1. Found that there's no provider support for grok(`x-ai/*`), so added it.
2. `gemini` is not a provider name, so changed it to `google`.

khj809 (approved):

PR #415, opened 2026-05-21
fix(sandbox): decouple VM removal from Machine::shutdown

khj809:
## Problem

When an agent executed multiple tool calls against the same sandbox, the second tool call would fail with a \"sandbox not found\" error. The root cause was a combination of three behaviors:

1. **`Sandbox::stop`** called `MsbSandbox::remove(name)` whenever `persist=false`, permanently deleting the VM definition from the microsandbox DB.
2. **`RunEnvHandle::drop`** triggered `Machine::shutdown` (→ `Sandbox::stop`) every time the strong count of `Arc<RunEnvHandle>` hit zero.
3. **Agents hold handles briefly**: `ensure_files_materialised` takes and immediately drops a handle before the LLM call, and each tool task in `execute_tool_calls` independently calls `runenv.get()` and drops the handle when its stream ends.

This meant: tool A finishes → handle drops → `Machine::shutdown` → VM **removed** → tool B starts → `start_and_connect` cannot find the registered name → error.

## Fix

Decouple VM *removal* from VM *stop* by separating responsibilities:

- **`Machine::shutdown`** (called on every handle drop): stops the VM only — never removes its definition.
- **`Sandbox::Drop`**: sole owner of VM definition removal for `persist=false` sandboxes. The VM definition now survives multiple boot/shutdown cycles within the same `Sandbox` lifetime and is only deleted when the `Sandbox` struct itself is dropped.

### Key code changes (`src/runenv/sandbox.rs`)

- `Sandbox::stop(name)` — removed `persist: bool` parameter and the trailing `MsbSandbox::remove` call. Now stop-only.
- New private `Sandbox::remove(name)` — stops then removes the VM definition. Called exclusively from `Drop`.
- `Machine::shutdown` — now calls `Self::stop` only; no longer reads `self.config.persist`.
- `Drop for Sandbox` — calls `Self::remove` instead of `Self::stop(…, false)`. Drop is now the *sole* owner of VM removal, not just a safety net.
- `Sandbox::remove_persisted(name)` — public associated function for explicit cleanup of `persist=true` sandboxes when the `Sandbox` instance is no longer available (e.g. after a process restart).

### `persist` semantics redefined

`SandboxConfig::persist` now controls whether the **VM definition survives `Sandbox` struct drop**, independent of `Machine::boot`/`Machine::shutdown` cycles:

| | `persist = false` (default) | `persist = true` |
|---|---|---|
| On `Machine::shutdown` | VM stopped | VM stopped |
| On `Sandbox` drop | VM stopped + **removed** | VM left intact |
| Cleanup | automatic | explicit `Sandbox::remove_persisted(name)` call |

## Tests

- **New**: `test_vm_definition_survives_machine_shutdown_with_persist_false` — directly reproduces the bug scenario: two sequential boot/shutdown cycles with `persist=false`, asserting the VM definition survives between them and is removed only after `Sandbox` drop.
- **Updated**: `test_start_stop_idempotent` — switched to `persist=false` to directly validate the new guarantee (VM definition survives repeated boot/drop cycles within one `Sandbox` lifetime).
- **Updated**: `test_filesystem_persists_across_stop_start` — removed `persist=true` dependency; filesystem state persists across stop/start cycles regardless of the `persist` flag, as long as the `Sandbox` is alive.

### DIFF
--- a/src/runenv/sandbox.rs
+++ b/src/runenv/sandbox.rs
@@ -1491,7 +1491,7 @@ mod tests {
 
         drop(_console);
         src.shutdown().await;
-        remove_persisted(&src_name).await.ok();
+        Sandbox::remove_persisted(&src_name).await.ok();
     }
 
     #[tokio::test]
@@ -1539,8 +1539,8 @@ mod tests {
 
         drop(child_console);
         child.shutdown().await;
-        remove_persisted(&src_name).await.ok();
-        remove_persisted(&child_name).await.ok();
+        Sandbox::remove_persisted(&src_name).await.ok();
+        Sandbox::remove_persisted(&child_name).await.ok();
     }
 
     #[tokio::test]
@@ -1597,8 +1597,8 @@ mod tests {
 
         drop(src_console);
         src.shutdown().await;
-        remove_persisted(&src_name).await.ok();
-        remove_persisted(&child_name).await.ok();
+        Sandbox::remove_persisted(&src_name).await.ok();
+        Sandbox::remove_persisted(&child_name).await.ok();
     }
 
     #[tokio::test]
@@ -1635,7 +1635,7 @@ mod tests {
         );
 
         child.shutdown().await;
-        remove_persisted(&src_name).await.ok();
-        remove_persisted(&child_name).await.ok();
+        Sandbox::remove_persisted(&src_name).await.ok();
+        Sandbox::remove_persisted(&child_name).await.ok();
     }
 }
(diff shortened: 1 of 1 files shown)


jhlee525 (approved):

jhlee525:
How about just removing persisted option?

khj809:
> How about just removing persisted option?

If you meant to always persist the sandbox and let users handle them, I don't think users are such diligent as we expect, and they'll encounter tons of unexpected obsolete msb instances eventually. The current minimal guard (removing at Sandbox's Drop) seems sufficient to prevent such situation without harming user experiences so much.

PR #416, opened 2026-05-22
fix(runenv): unblock Drop, persist /tmp, rename Machine verbs, bump msb 0.5.0

nuri-yoo:
## Summary

Four facets of the same handle-lifecycle problem in sandbox-backed `Agent::run`. Each is a thin change; they ship together because they share the same code path and the same reviewer thread.

**(1) `RunEnvHandle::Drop` and `Sandbox::Drop` block the caller thread on a sync `recv_timeout(30s)`.** When `Drop` fires on a tokio worker the worker stalls for up to 30 seconds and later-scheduled futures on the same runtime starve. Measured on a 4-parallel `api_search` turn against a sandbox-backed agent on `claude-opus-4-7`: 97 to 99 seconds wall time with 2 of 4 calls failing on `error sending request`. With the sync wait removed: 6.4 seconds, 0 failures.

**(2) `/tmp` in an ailoy sandbox does not survive between tool-call batches.** For OCI rootfs sandboxes, microsandbox auto-attaches a tmpfs at `/tmp` (`apply_runtime_defaults` in `microsandbox/lib/sandbox/config.rs`). That tmpfs lives in VM memory; the VM stops every time `RunEnvHandle::Drop` runs, and the tmpfs evaporates with it. Files an agent writes with `curl > /tmp/foo` disappear before the next tool call can read them. This is sandbox-only — `Local::stop` is a no-op.

**(3) `Machine::boot` and `Machine::shutdown` use different verbs than the microsandbox API they wrap.** microsandbox itself uses `start` / `stop` (`MsbSandbox::start_detached`, `MsbSandbox::stop_and_wait`, `SandboxHandle::stop`). The `boot` / `shutdown` vocabulary also reads to a language model as a full restart, which biases against the `/tmp`-persists-between-starts contract this PR establishes.

**(4) An ad-hoc `sync -f /` workaround sits in `Sandbox::stop_vm` to compensate for v0.4.6 agentd not flushing the rootfs before poweroff.** Once v0.5.0 lands (superradcompany/microsandbox#746), agentd does sync + clean poweroff itself, so the workaround becomes dead code.

## Changes

**Commit 1 — `fix(runenv): unblock caller thread in Drop handlers`** (`src/runenv/mod.rs`, `src/runenv/sandbox.rs`)

Two lines are commented out (kept as a record, not deleted):

```rust
// let _ = rx.recv_timeout(std::time::Duration::from_secs(30));
let _ = rx;
```

The background cleanup thread still does the actual cleanup (sync filesystem, stop VM, transition `RunEnvState` to `Idle`, remove the VM definition on `Sandbox` drop). The change is that the caller no longer waits for it synchronously.

Two existing tests in `src/runenv/sandbox.rs` (`test_stop_and_start`, `test_start_stop_idempotent`) assumed `drop(handle)` returned only after stop had completed. They now poll `vm_is_running` with a 30s deadline through a small `wait_until_stopped` helper, expressing the same intent without the race the immediate assert had against the (now async) stop.

`recv_timeout(30s)` never actually guaranteed cleanup completion: if cleanup took longer it returned anyway with no diagnostic. Replacing the asserts with bounded polling keeps the coverage of the intended behaviour ("the VM stops once the last handle is dropped") and removes the runtime stall.

**Commit 2 — `refactor(runenv): rename Machine::boot/shutdown to start/stop`** (`src/runenv/mod.rs`, `src/runenv/local.rs`, `src/agent/rt.rs`, `src/agent/builder.rs`)

`Machine::boot` → `Machine::start`, `Machine::shutdown` → `Machine::stop` (and the same on `MachineDyn`, the `Local` impl, the `RunEnv::get` / `RunEnvHandle::Drop` call sites). Doc comments and the `expect("runenv boot failed")` strings in agent tests are updated for consistency. No behaviour change.

**Commit 3 — `fix(sandbox): mount /tmp on a named volume by default + Machine rename`** (`src/runenv/sandbox.rs`)

Two changes in the sandbox `Machine` implementation:

1. `Sandbox::new` injects a per-sandbox named volume at `/tmp` by default. The escape-hatch branch in `apply_runtime_defaults` fires when `mounts` already contains a `/tmp` entry, so microsandbox stops attaching its tmpfs. Volume names are `ailoy-tmp-{sandbox_name}` so concurrent sandboxes stay isolated, and every provisioned volume carries the `ailoy.managed=true` label. A user-supplied `/tmp` mount in `SandboxConfig::volumes` takes precedence; the injection is skipped. Cleanup runs in `Sandbox::remove` (drop path) and `Sandbox::remove_persisted`. Volumes provisioned for a sandbox that fails to register are rolled back so a failed `Sandbox::new` does not leak.

   A startup sweep gated by a `tokio::sync::OnceCell` runs at most once per process. It lists volumes via `microsandbox::Volume::list()`, intersects with `MsbSandbox::list()`, and removes only volumes that carry the `ailoy.managed` label, match the `ailoy-tmp-` prefix, and have no live owner sandbox. Volumes without the label are left alone, so volumes from other tools (and from an in-flight `Sandbox::new` in a parallel process) are never touched.

2. `impl Machine for Sandbox` is renamed to match commit 2. The private helper `Sandbox::stop` is renamed to `stop_vm` so the trait method `stop` is free for the new vocabulary. `test_vm_definition_survives_machine_shutdown_with_persist_false` from #415 is renamed in lockstep.

`persist=true` semantics are preserved exactly as #415 defined them: `Machine::start`/`Machine::stop` cycles do not depend on `persist`, and the auto-injected `/tmp` volume is reused across `Sandbox::new` calls for the same name (the `VolumeAlreadyExists` from a second `Volume::create` is silently absorbed). Pre-existing `persist=true` sandboxes created before this change keep the microsandbox auto-tmpfs at `/tmp`; remove and recreate them to pick up the new default.

**Commit 4 — `chore(deps): bump microsandbox to v0.5.0, drop sync-flush workaround`** (`Cargo.toml`, `Cargo.lock`, `src/runenv/sandbox.rs`)

microsandbox v0.5.0 ships the upstream fix this commit depends on. The relevant upstream change is superradcompany/microsandbox#746 (`fix(sandbox): flush root-fs writes on stop`), which routes `SandboxHandle::stop` through agentd's own sync + clean poweroff. The explicit `sync -f /` we ran before `handle.stop()` is no longer needed; the workaround block inside `Sandbox::stop_vm` collapses to a single `handle.stop()` call (17 lines deleted) and the TODO comment that tracked the upstream fix is removed.

`Cargo.toml` switches to `tag = "v0.5.0"` so the dependency reads as a versioned release rather than an opaque commit hash. `microsandbox::setup::install` auto-downloads the v0.5.0 prebuilt on the first `Sandbox::new`.

## Test plan

|                                | sandbox module | notes |
| ---                            | ---            | --- |
| develop baseline (v0.4.6)      | 34 / 34        | pre-existing tests, no new ones |
| commits 1+2+3 on v0.4.6        | 38 / 38        | + 4 new regression tests added in commit 3 |
| commits 1+2+3+4 on v0.5.0      | 38 / 38        | same 4 + 34 against the v0.5.0 prebuilt |

Workspace-level `cargo test --workspace --all-features` outside the sandbox feature is unchanged from develop (OpenAI 429 quota failures from the local environment, unrelated to this PR).

Command: `cargo test --workspace --all-features --release` for the sandbox feature, run separately on v0.4.6 and v0.5.0 to confirm parity.

The four regression tests added in commit 3, running on v0.5.0:

```
test runenv::sandbox::tests::test_tmp_persists_across_starts ... ok
test runenv::sandbox::tests::test_user_tmpfs_overrides_default_tmp ... ok
test runenv::sandbox::tests::test_tmp_volume_removed_on_sandbox_drop ... ok
test runenv::sandbox::tests::test_startup_sweep_targets_only_labelled_orphans ... ok
```

Full sandbox test module on v0.5.0: 38 passed, 0 failed (23.56s).

## Measurement detail

Reproducer: a deep-research-style agent in `brekkylab/agent-k` issuing a 4-parallel `api_search` call inside a single `agent.run` turn, sandbox-backed, on `claude-opus-4-7`. A second probe writes `/tmp/probe` in tool call 1 and reads it in tool call 2 of the same `agent.run`.

| condition                                   | wall time            | reqwest fail | /tmp persists across calls |
| ---                                         | ---                  | ---          | ---                        |
| develop baseline                            | 97 - 99s             | 2 / 4        | no                         |
| commit 1 only (Drop unblock)                | 6.4s                 | 0 / 4        | no                         |
| sandbox.rs Drop wait removed only           | 99s (first at 35s)   | 1 / 4        | no                         |
| commits 1+2+3+4 (this PR head)              | ~6s                  | 0 / 4        | yes                        |

VM lifecycle is unchanged: the VM still stops between batches, and the agent still releases the VM during LLM inference. The `/tmp` data is what now survives, not the VM.

## Cross-reference

* `persist` semantics: PR #415 redefined `persist` as orthogonal to `Machine::start` / `Machine::stop`. This PR keeps that invariant intact — the auto-injected volume lifecycle keys off `Sandbox` drop, not off VM start/stop.
* microsandbox escape hatch: `apply_runtime_defaults` in `microsandbox/lib/sandbox/config.rs` returns early when `mounts` already contains `/tmp`. That is the contract commit 3 relies on.
* microsandbox upstream fix: superradcompany/microsandbox#746 (`fix(sandbox): flush root-fs writes on stop`), landed in v0.5.0. That is the contract commit 4 relies on.
* microsandbox volume API: `microsandbox::Volume::builder().label().create()`, `Volume::list()`, `Volume::remove()`, `VolumeHandle::labels()`.

## Known limitation

The startup sweep covers process abort and the `RunEnvHandle::Drop` race introduced by commit 1, but it cannot rescue a sandbox that was *mid-`Sandbox::new`* when a parallel process started its sweep. The `ailoy.managed` label keeps that case from clobbering other tools' volumes; a parallel ailoy process's in-flight volume is the residual race window.

Upstream fix in flight: superradcompany/microsandbox#801 adds `SandboxBuilder::auto_volume`, which provisions the volume and the sandbox in one DB transaction. Once that lands and we bump to the release that carries it, this PR's two-step `Volume::create` + `VolumeMount::Named` can collapse into a single builder call and the race window closes at the microsandbox layer. Until then, the pid + start-timestamp label mitigation stays a follow-up if it turns out to bite in practice.

## Trade-off

`/tmp` in an ailoy sandbox no longer matches POSIX `/tmp` semantics: it survives the start/stop cycle that `RunEnvHandle::Drop` drives. Callers who want strict POSIX semantics override `SandboxConfig::volumes` with their own `VolumeMount::Tmpfs { guest: "/tmp", .. }`. The named volume has no size cap (disk quota only) where the auto-tmpfs was clamped to `[1, 512]` MiB; volume removal is best-effort and covered by the startup sweep.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg",
 indexmap = { version = "2", features = ["serde"] }
 infer = "0.16"
 log = "0.4"
-microsandbox = { git = "https://github.com/superradcompany/microsandbox", rev = "1294b4d7f850741e4d7fa91624226fcb45aa848a", optional = true }
+microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.0", optional = true }
 ordered-float = { version = "5.0.0", features = ["serde"] }
 parking_lot = "0.12.4"
 png = "0.18"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1904,6 +1904,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "fslock"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2504,12 +2513,12 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "system-configuration 0.7.0",
  "tokio",
  "tower-service",
  "tracing",
- "windows-registry 0.6.1",
+ "windows-registry",
 ]
 
 [[package]]
@@ -2761,6 +2770,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "inotify"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bd5b3eaf1a28b758ac0faa5a4254e8ab2705605496f1b1f3fbbc3988ad73d199"
+dependencies = [
+ "bitflags 2.11.0",
+ "inotify-sys",
+ "libc",
+]
+
+[[package]]
+name = "inotify-sys"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e05c02b5e89bff3b946cedeca278abc628fe811e604f027c45a8aa3cf793d0eb"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "inout"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2953,6 +2982,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4933f3f57a8e9d9da04db23fb153356ecaf00cbd14aee46279c33dc80925c37"
 
 [[package]]
+name = "kqueue"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eac30106d7dce88daf4a3fcb4879ea939476d5074a9b7ddd0fb97fa4bed5596a"
+dependencies = [
+ "kqueue-sys",
+ "libc",
+]
+
+[[package]]
+name = "kqueue-sys"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07293a4e297ac234359b510362495713f75ea345d5307140414f20c69ffeb087"
+dependencies = [
+ "bitflags 2.11.0",
+ "libc",
+]
+
+[[package]]
 name = "kvm-bindings"
 version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3115,6 +3164,9 @@ name = "lru"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "227748d55f2f0ab4735d87fd623798cb6b664512fe979705f829c9f81c934465"
+dependencies = [
+ "hashbrown 0.15.5",
+]
 
 [[package]]
 name = "lru"
@@ -3248,13 +3300,15 @@ dependencies = [
 
 [[package]]
 name = "microsandbox"
-version = "0.4.6"
-source = "git+https://github.com/superradcompany/microsandbox?rev=1294b4d7f850741e4d7fa91624226fcb45aa848a#1294b4d7f850741e4d7fa91624226fcb45aa848a"
+version = "0.5.0"
+source = "git+https://github.com/superradcompany/microsandbox?tag=v0.5.0#fa3b2a563e8a519eae0edc84a79d2e325acb0c49"
 dependencies = [
  "astral-tokio-tar",
  "async-compression",
+ "base64 0.22.1",
  "bytes",
  "chrono",
+ "ciborium",
  "crossterm",
  "dbus",
  "dirs",
@@ -3267,12 +3321,14 @@ dependencies = [
  "microsandbox-db",
  "microsandbox-filesystem",
  "microsandbox-image",
+ "microsandbox-metrics",
  "microsandbox-migration",
  "microsandbox-network",
  "microsandbox-protocol",
  "microsandbox-runtime",
  "microsandbox-utils",
  "nix 0.30.1",
+ "notify",
  "rand 0.10.1",
  "reqwest 0.13.2",
  "scopeguard",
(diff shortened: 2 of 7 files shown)


jhlee525:
Pinning the `runenv` into the `Agent::run` function is certainly the most reliable solution.

However, the problem is that the `runenv` remains active throughout the agent's lifecycle — or at least for the entire duration of an agent run.

In most cases, agent runs spend the majority of their time on LLM inference, while tool calls themselves execute only for a relatively short period. Keeping the `runenv` alive the whole time can therefore significantly reduce parallelism when running Ailoy in a server environment.

Because of this, I think we need to find a solution where the `runenv` does not stay booted for the entire duration of the agent run.

nuri-yoo:
> Pinning the `runenv` into the `Agent::run` function is certainly the most reliable solution.
> 
> However, the problem is that the `runenv` remains active throughout the agent's lifecycle — or at least for the entire duration of an agent run.
> 
> In most cases, agent runs spend the majority of their time on LLM inference, while tool calls themselves execute only for a relatively short period. Keeping the `runenv` alive the whole time can therefore significantly reduce parallelism when running Ailoy in a server environment.
> 
> Because of this, I think we need to find a solution where the `runenv` does not stay booted for the entire duration of the agent run.

Addressed in commits f95412f and 9c0bf54.

jhlee525 (approved):
If replacing the /tmp directory with a mounted directory does not cause any issues, then this PR should be fine. (I think it probably will be okay.)

For now, let's proceed with this approach.

Cargo.toml
```diff
@@ -43,7 +43,7 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg",
 indexmap = { version = "2", features = ["serde"] }
 infer = "0.16"
 log = "0.4"
-microsandbox = { git = "https://github.com/superradcompany/microsandbox", rev = "1294b4d7f850741e4d7fa91624226fcb45aa848a", optional = true }
+microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.0", optional = true }
```
khj809, line 46: nit: It seems microsandbox has failed to release 0.5.0 to crates.io.
Not sure when they'll fix the release error, but we need to change it as below in the future:
```suggestion
microsandbox = { version = "0.5.0", optional = true }
```

khj809 (approved):
Thanks for the hard work!
I understood the current situation that the parallel sandbox tests fail randomly due to the non-atomic named volume creation. Hope that https://github.com/superradcompany/microsandbox/pull/801 will be merged soon.

PR #417, opened 2026-05-26
Agent state

jhlee525:
- Collapse `ContextManager` and the standalone history vector into a single `AgentState` that owns the message transcript, truncation policy (`max_input_tokens`, `preserve_recent_turns`, `last_input_tokens`), and the shared `RunEnv`. The old `src/agent/context.rs` is deleted; the truncation algorithm lives on `AgentState::{truncate, truncate_if_needed}` alongside the data it acts on.
- Share `RunEnv` by reference instead of value: drop `#[derive(Clone)]` and switch agent / subagent constructors to `impl Into<Arc<RunEnv>>`, so the same VM is reference-counted across parent and sub-agents rather than cloned.
- Slim `AgentBuilder` to a single `state: AgentState` field. New builder surface: `.state(AgentState)` to swap wholesale, plus direct `.messages(...)`, `.max_input_tokens(n)`, `.preserve_recent_turns(n)` shortcuts. The old `.history(...)` / `.context_manager(...)` methods are gone.
- Add `AgentState::{list_skills, get_skill}` backed by new `skill::{list_skills, get_skill}` helpers that walk `SKILL.md` files through the runenv shell — so skill discovery works against any backend (local FS or sandbox).


src/agent/state.rs
```diff
@@ -0,0 +1,388 @@
...
+/// the [`preserve_recent_turns`](Self::preserve_recent_turns) window are
+/// reduced to `"[context truncated]"` placeholders.
+pub struct AgentState {
+    pub messages: Vec<Message>,
+
+    /// Triggers truncation when input_tokens from the previous API call exceeds this value.
+    pub max_input_tokens: u64,
+
+    /// Number of recent user turns to preserve after truncation (system message is always preserved separately).
+    pub preserve_recent_turns: usize,
```
khj809, line 23: I can't understand why these values should be inside `AgentState`, which is for storing dynamic state data.

src/agent/rt.rs
```diff
@@ -425,18 +383,15 @@ impl Agent {
...
-            self.state.history.push(query);
+            self.state.messages.push(query);
 
             loop {
                 // Truncation check based on previous call's token usage.
-                if let Some(cm) = &self.context_manager
-                    && self.state.last_input_tokens.unwrap_or(0) > cm.max_input_tokens {
-                        cm.truncate_history(&mut self.state.history);
-                    }
+                self.state.truncate_if_needed();
```
khj809, line 387: I don't agree with placing truncation logics inside `AgentState`. Context compaction will become highly complex than the current implementation, and that's why I made the dedicated `ContextManager`. I think `AgentState` should work as a data type for storing the dynamic states, not a struct which has a lot of special logics.

selloriwoo:
Were list_skills and get_skill added with future use cases in mind — for example, skills being dynamically added to the RunEnv at runtime?

PR #418, opened 2026-05-28
chore(deps): bump microsandbox to v0.5.2

nuri-yoo:
## Summary

Bumps the microsandbox git tag from `v0.5.0` to `v0.5.2`.

Notable upstream changes shipped between `v0.5.0` and `v0.5.2`:

- superradcompany/microsandbox#810 — `fix(upgrade): support 0.5 upgrade path for old sandboxes`. Pre-0.5 sandboxes left over on disk now get a graceful restart-required path instead of a handshake failure on `exec`.
- superradcompany/microsandbox#820 — `fix(runtime): support long sandbox names for agent sockets`. The agent relay socket is derived from a fixed-length hash under `~/.microsandbox/run/agent/` instead of `<sandboxes>/<name>/runtime/agent.sock`, so long sandbox names no longer break `SUN_LEN`.

The other 0.5.x deltas — agentd absorbed into the root workspace (superradcompany/microsandbox#804), SSH enabled in release binaries (superradcompany/microsandbox#812), DNS / network-policy fixes (superradcompany/microsandbox#824, #825), SDK doc links (superradcompany/microsandbox#811, #826) — are transparent at this layer.

## Dep form

The dependency moves to a crates.io version requirement, replacing the previous `git + tag` form used while the v0.5.0 publish was still pending. Every `microsandbox-*` member is on crates.io 0.5.2.

```diff
-microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.0", optional = true }
+microsandbox = { version = "0.5.2", optional = true }
```

`validate_sandbox_name`'s `SUN_PATH_MAX` check is left as-is.

## Verification

The local `~/.microsandbox/{bin,lib}` caches were wiped, then the v0.5.2 prebuilt was cold-pathed through `microsandbox::setup::install` on the first `Sandbox::new`. The full sandbox test module was then run.

```
cargo test --release --all-features runenv::sandbox::tests -- --test-threads=1
...
test result: ok. 38 passed; 0 failed; 0 ignored; 0 measured; 269 filtered out; finished in 21.33s
```

The 38 tests include the four regression tests added in #416 (`test_tmp_persists_across_starts`, `test_user_tmpfs_overrides_default_tmp`, `test_tmp_volume_removed_on_sandbox_drop`, `test_startup_sweep_targets_only_labelled_orphans`) plus the 34 pre-existing sandbox-feature tests.

`cargo check --all-features` also passes.

## Cross-reference

- superradcompany/microsandbox v0.5.1 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.1
- superradcompany/microsandbox v0.5.2 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.2
- The pending upstream PR for race-free per-sandbox auto-volumes (superradcompany/microsandbox#801) is unaffected by this bump and remains open against `main`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg",
 indexmap = { version = "2", features = ["serde"] }
 infer = "0.16"
 log = "0.4"
-microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.0", optional = true }
+microsandbox = { version = "0.5.2", optional = true }
 ordered-float = { version = "5.0.0", features = ["serde"] }
 parking_lot = "0.12.4"
 png = "0.18"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2513,7 +2513,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "system-configuration 0.7.0",
  "tokio",
  "tower-service",
@@ -3164,9 +3164,6 @@ name = "lru"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "227748d55f2f0ab4735d87fd623798cb6b664512fe979705f829c9f81c934465"
-dependencies = [
- "hashbrown 0.15.5",
-]
 
 [[package]]
 name = "lru"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -43,7 +43,7 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg",
 indexmap = { version = "2", features = ["serde"] }
 infer = "0.16"
 log = "0.4"
-microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.0", optional = true }
+microsandbox = { git = "https://github.com/superradcompany/microsandbox", tag = "v0.5.2", optional = true }
```
khj809: The git+tag approach makes sense as a workaround when a publish is broken or pending, but now that v0.5.2 is fully available, `version = "0.5.2"` is the more idiomatic and maintainable path forward.

A couple of notes on the PR description's rationale:

**On lockfile determinism:** `Cargo.lock` pins exact versions and checksums for registry deps just as it records the exact git commit SHA for git deps — either way, builds are fully reproducible once the lockfile is committed.

**On build speed:** crates.io serves compressed tarballs via CDN, which is noticeably faster than git operations, especially for a workspace with this many crates.

nuri-yoo: Addressed in https://github.com/brekkylab/ailoy/commit/92cff9473c3d80bb5655f4b0d194627941aa0107 — switched to version = "0.5.2".

khj809 (approved):

PR #421, opened 2026-05-29
feat(lang_model): per-call thinking toggle for Moonshot kimi-k2

nuri-yoo:
## Summary

Adds a `thinking` option on `LangModelOptions` that the ChatCompletion marshal forwards as Moonshot's `\"thinking\": {\"type\": \"enabled\" | \"disabled\"}` request-body field. Today `kimi-k2.5` and `kimi-k2.6` default to thinking-enabled and emit responses where every streamed delta carries `reasoning_content` and `content` is left empty until reasoning finishes. Callers that just want a short answer (e.g. ReAct tool-call decisions on small queries) have no way through ailoy to flip that off; this PR adds one.

## Changes

`src/lang_model/options.rs`: new `ThinkingMode { Enabled, Disabled }` enum and a `thinking: Option<ThinkingMode>` field on `LangModelOptions`, following the same `Option` + `skip_serializing_if` pattern as the existing `max_tokens` / `temperature` / `top_p` / `top_k` / `response_format` fields.

`src/lang_model/rt.rs`: `LangModelRequest` carries the new option; `LangModel::run` forwards `options.thinking` into it next to the other model options.

`src/lang_model/impl/api/chat_completion.rs`: new private helper `is_moonshot_thinking_model(model: &str) -> bool` matching `kimi-k2.5` and `kimi-k2.6` (mirrors `is_openai_reasoning_model`). The marshal inserts `\"thinking\": {\"type\": ...}` into the request body only when both the option is set and the model passes that check. Other ChatCompletion-schema vendors (OpenAI, x.ai, DeepSeek) silently drop the field, matching how `temperature` / `top_p` are already dropped on OpenAI reasoning models.

`src/agent/spec.rs`, `src/agent/builder.rs`: new fluent setters `AgentSpec::thinking(mode)` and `AgentBuilder::thinking(mode)`, parallel to the existing `temperature` / `top_p` / `response_format` setters.

The four marshals (`chat_completion`, `openai`, `anthropic`, `gemini`) all construct `LangModelRequest` literals in test fixtures, so each gets `thinking: None` added to keep the struct literal exhaustive. No behaviour change in the latter three.

## Tests

`src/lang_model/impl/api/chat_completion.rs` adds four unit tests:

- `test_marshal_thinking_kimi_disabled`: `kimi-k2.6` + `Disabled` produces `\"thinking\": {\"type\": \"disabled\"}` in the wire body.
- `test_marshal_thinking_kimi_enabled`: `kimi-k2.5` + `Enabled` produces `\"thinking\": {\"type\": \"enabled\"}`.
- `test_marshal_thinking_ignored_on_other_models`: `gpt-4.1-mini` + `Disabled` produces no `thinking` field in the wire body.
- `test_marshal_thinking_absent_when_none`: `kimi-k2.6` with `thinking: None` produces no `thinking` field.

Full result on a clean develop tip:

\`\`\`
cargo test --features sandbox --lib lang_model
test result: ok. 54 passed; 0 failed; 5 ignored; 0 measured; 281 filtered out; finished in 2.51s
\`\`\`

The 5 ignored tests are the pre-existing `test_run_*` integration tests that require provider env keys.

## Motivation, observed behaviour

Direct streaming probe against `https://api.moonshot.ai/v1/chat/completions` for `kimi-k2.6` with `\"Say hi in 3 words\"` and `max_tokens: 50` emits 48 streamed chunks, every one of which is a `reasoning_content` delta; the model never reaches the `content` field within that budget. The same prompt with `\"thinking\": {\"type\": \"disabled\"}` in the body returns `\"Hello there, friend!\"` in 1.07s, with `content` populated and no `reasoning_content`. In a ReAct loop over a multi-turn agent (`Agent::run` with corpus tools), the thinking-on flavour amplifies into roughly 60-130s per turn before the first `tool_call` is emitted; flipping the toggle off brings it into the same range as the other small light-tier models.

## Trade-off

This PR keeps Moonshot's default behaviour (thinking enabled). Callers must opt out explicitly via `LangModelOptions { thinking: Some(ThinkingMode::Disabled), ... }` or `AgentSpec::thinking(ThinkingMode::Disabled)`. The `is_moonshot_thinking_model` helper is intentionally narrow (the two literal model prefixes) so future Moonshot models, or vendors that adopt the same shape, get an explicit decision rather than silent capture.

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -174,6 +174,11 @@ impl AgentBuilder {
         self
     }
 
+    pub fn thinking(mut self, mode: crate::lang_model::ThinkingMode) -> Self {
+        self.spec = self.spec.thinking(mode);
+        self
+    }
+
     /// Append a single pre-fill [`FileEntry`] to `spec.files`.  The file is
     /// written into the runenv on the agent's first `run`.
     pub fn file(mut self, entry: FileEntry) -> Self {
--- a/src/agent/spec.rs
+++ b/src/agent/spec.rs
@@ -218,6 +218,13 @@ impl AgentSpec {
         self
     }
 
+    pub fn thinking(mut self, mode: crate::lang_model::ThinkingMode) -> Self {
+        self.model_options
+            .get_or_insert_with(LangModelOptions::new)
+            .thinking = Some(mode);
+        self
+    }
+
     pub fn file(mut self, file: FileEntry) -> Self {
         self.files.push(file);
         self
--- a/src/lang_model/impl/api/anthropic.rs
+++ b/src/lang_model/impl/api/anthropic.rs
@@ -434,6 +434,7 @@ mod tests {
             top_p: None,
             top_k: None,
             response_format: None,
+            thinking: None,
         };
         f(&req)
     }
@@ -655,6 +656,7 @@ mod tests {
             top_p: None,
             top_k: None,
             response_format: Some(&fmt),
+            thinking: None,
         };
         let val = AnthropicMarshal::default().marshal(&req);
         let body = val.as_object().unwrap().get("body").unwrap();
--- a/src/lang_model/impl/api/chat_completion.rs
+++ b/src/lang_model/impl/api/chat_completion.rs
@@ -52,6 +52,12 @@ impl LangModelProvider {
     }
 }
 
+/// Moonshot kimi-k2.5 / kimi-k2.6 accept a `thinking` toggle in the request
+/// body. Other ChatCompletion-schema vendors (OpenAI, x.ai, DeepSeek) ignore it.
+fn is_moonshot_thinking_model(model: &str) -> bool {
+    model.starts_with("kimi-k2.5") || model.starts_with("kimi-k2.6")
+}
+
 #[derive(Clone, Debug, Default)]
 pub struct ChatCompletionMarshal;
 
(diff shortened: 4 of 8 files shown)


khj809 (approved):
I think other guys will mention about using the existing reasoning_effort instead of introducing a new option, so I'll not cover that in my review.

What I'm worried is that `chat_completion.rs` is getting bloated by many provider-specific conditions. In the long term, we may need to generalize `ChatCompletionMarshal` into `ChatCompletionMarshal<P: ChatCompletionProvider>`, or applying provider-specific `pre_marshal_message()` or `post_unmarshal_message()` hooks.

nuri-yoo:
> I think other guys will mention about using the existing reasoning_effort instead of introducing a new option, so I'll not cover that in my review.
> 
> What I'm worried is that `chat_completion.rs` is getting bloated by many provider-specific conditions. In the long term, we may need to generalize `ChatCompletionMarshal` into `ChatCompletionMarshal<P: ChatCompletionProvider>`, or applying provider-specific `pre_marshal_message()` or `post_unmarshal_message()` hooks.

Totally agree. Currently, kimi is experiencing two issues. First is thinking mode, Second is timeout issue. 
Further, we need to decide whether to adopt "per call" settings or not.
Maintaining this PR is necessary to test kimi while avoiding infinite thinking, changing this pr to draft.

PR #423, opened 2026-06-01
feat(lang-model): add tool_choice option + per-turn override

nuri-yoo:
## Summary

Adds a `tool_choice` knob to `LangModelOptions` and a per-turn override on `Agent` so an agent loop can deterministically end runaway tool use.

- `LangModelOptions::tool_choice: Option<ToolChoice>` with `Auto / None / Required`
- Honored by all three providers: Anthropic Messages (`tool_choice.type` = `auto|none|any`), OpenAI Responses and Chat Completions (`tool_choice` = `auto|none|required`), Gemini (`tool_config.function_calling_config.mode` = `AUTO|NONE|ANY`)
- `Agent::tool_choice_handle()` returns an `Arc<Mutex<Option<ToolChoice>>>`. Writing `Some(_)` causes the next model call to use that policy in place of the static `model_options.tool_choice`. The override is one-shot — the loop clears it after each model call.

## Motivation

Running CUAD / FinanceBench Q&A ablations on PR #121's `speedwagon` agent, Gemini routinely enters a verify-loop where it calls `read_document` over and over on overlapping windows until the 120s per-QA timeout. Prompt-only guidance (`max 3 reads`, `do not re-read overlapping windows`) only partially helps because the model continues to choose tool calls when offered the choice.

A tool wrapper can already count calls and return an error after a budget, but the model is still free to call the tool again. The clean fix is to give the loop a way to say \"no more tools\" at the protocol level for the next request. `tool_choice = \"none\"` on the next call forces a text-only completion, which is what we actually want.

## Wire output (verified by unit tests)

| Choice    | Anthropic            | OpenAI Responses / ChatCompletion | Gemini                                              |
| --------- | -------------------- | ---------------------------------- | --------------------------------------------------- |
| Auto      | `{\"type\":\"auto\"}` | `\"auto\"`                          | (no `tool_config` — provider default)              |
| None      | `{\"type\":\"none\"}` | `\"none\"`                          | `{\"function_calling_config\":{\"mode\":\"NONE\"}}` |
| Required  | `{\"type\":\"any\"}`  | `\"required\"`                      | `{\"function_calling_config\":{\"mode\":\"ANY\"}}`  |

When `tools` is empty for the request, the field is omitted on every provider (existing behavior preserved).

## Tests

12 new unit tests cover the four adapters: `default → auto`, `None → forbidden`, `Required → mapped correctly`. The existing 275 lib tests still pass.

## Test plan

- [x] `cargo test --lib` — 287 passed
- [ ] End-to-end smoke: agent-k speedwagon read-budget guard sets `ToolChoice::None` on budget-hit; verify Gemini stops re-reading and emits a text answer on the next turn

### DIFF
--- a/src/agent/rt.rs
+++ b/src/agent/rt.rs
@@ -1,10 +1,15 @@
-use std::{collections::HashMap, path::PathBuf, pin::Pin};
+use std::{
+    collections::HashMap,
+    path::PathBuf,
+    pin::Pin,
+    sync::{Arc, Mutex as StdMutex},
+};
 
 use futures::{FutureExt as _, Stream, StreamExt as _};
 
 use crate::{
     agent::{AgentProvider, AgentSpec, ContextManager, default_provider},
-    lang_model::{LangModel, LangModelOptions},
+    lang_model::{LangModel, LangModelOptions, ToolChoice},
     message::{FinishReason, Message, MessageOutput, Part, Role},
     runenv::{Console, RunEnv},
     skill::{render_skills_table, scan_declared_skills},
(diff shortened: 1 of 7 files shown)


nuri-yoo:
This issue has been solved by changing model.

PR #424, opened 2026-06-05
Runenv revised

jhlee525:
## Summary

The `RunEnv` / `RunEnvHandle` indirection — an `Arc`-counted wrapper that wrapped a `Box<dyn MachineDyn>` and managed start-on-first-`get` / stop-on-last-drop ref-counting — is gone. Agents now share a `Machine` directly via a `SharedMachine = Arc<Mutex<dyn MachineDyn>>` handle, with explicit `start` / `stop` and no implicit lifecycle.

This removes a layer of state that always existed and was always a half-step out of sync with the underlying machine. The previous design hid two correctness pitfalls behind the ref-count:

- Pure tools (ones that never need a console) would still boot the machine on first `get`, even when nothing would touch the console.
- `RunEnvHandle::Drop`'s background stop kept `Arc` semantics surprising: dropping the *last* handle was the only way to actually stop the VM, but callers couldn't tell which `Arc::clone` was the last one without inspecting refcounts.

After this PR, the machine's running state is exactly what `Machine::is_running` reports, and `ToolFunc` carries enough type-level information for the runtime to skip starting the machine for tools that don't need it. The companion microsandbox bump to v0.5.4 lets us delete the ailoy-side `/tmp` volume management that #416 had to put in place to compensate for missing upstream behaviour.

### 1. `refactor(runenv): drop RunEnv wrapper, expose Machine via SharedMachine` (85e0e81)

- **`Machine` trait reshape.** Add `is_running`. `start` now returns `&Console` whose lifetime is tied to `&mut self`, so a console borrow holds the machine for the duration of a tool stream. `stop` returns `Result` and defaults to no-op.
- **`SandboxConfig` → `SandboxBuilder`.** The struct + builder split is collapsed into a single chainable `SandboxBuilder`. Per-process `/tmp` volume sweeping and the `ailoy.managed` label scheme introduced in #416 are gone — `microsandbox::SandboxConfig` is reused directly (see the dep bump below).
- **`ToolFunc` splits.** `ToolFunc::Pure` and `ToolFunc::WithConsole`, with a public `needs_console()`. The runtime skips locking and starting the machine for pure tools in `execute_tool_calls`. Builtins that need shell access — `shell`, `read`, `write`, `edit`, `apply_patch`, `glob`, `grep`, `python_repl` — take `&dyn Console` instead of `Arc<RunEnvHandle>`.
- **`AgentBuilder` / `Agent` API renames.** `.runenv(..)` → `.machine(..)` (owned) and `.shared_machine(..)` (pre-shared handle); `Agent::try_with_runenv` → `Agent::try_with_machine`; subagents inherit the parent machine via `Arc::clone`.

### 2. `chore(deps): bump microsandbox 0.5.2 → 0.5.4` (rolled into 85e0e81)

The runenv refactor depends on this bump for two reasons:

- **Drop ailoy-side `/tmp` volume management.** #416 added per-sandbox named volumes (`ailoy-tmp-{name}`) with an `ailoy.managed` label and a startup sweep, working around microsandbox not having a first-class auto-volume API. The 0.5.4 release ships that auto-volume work upstream, so the ~200 lines of label-tracking, sweep, and rollback logic in `runenv/sandbox.rs` collapse to a direct `microsandbox::SandboxConfig` reuse.
- **`SandboxBuilder` parity.** The chainable builder this PR introduces mirrors the upstream builder API surface that landed in 0.5.3/0.5.4.

`Cargo.toml`: `microsandbox = "0.5.2"` → `"0.5.4"`. `cargo update` regenerates lock-file entries for `microsandbox-*` crates and their transitive deps.

### 3. `refactor(runenv): promote SharedMachine to a struct` (891563c)

The handle moves out of `agent::rt` into `runenv`, alongside the `Machine` trait it wraps. The `Arc<Mutex<dyn MachineDyn>>` type alias and free `shared_machine()` helper are replaced with `#[derive(Clone)] struct SharedMachine` exposing `new`, `get`, and a `From<M: Machine>` impl. Agents pass `SharedMachine` around as an opaque handle; callers don't need to know about the `Arc<Mutex<>>` shape.

### 4. `refactor(agent): split subagent module, introduce AgentParts, drop spec field` (107b422)

Riding on the new `Machine` API, the agent-side wiring gets a parallel cleanup:

- **Sub-agent `ToolFunc` construction** moves from `tool/impl/subagent.rs` into `agent/subagent.rs` so it can share `Agent`-internal resolution helpers with `agent/rt.rs`.
- **`AgentParts`**: everything that depends on `AgentProvider` is resolved once at outer-call time (model factory + options, tool descs and funcs, the flattened file list, the system message). The sub-agent closure captures only this owned, `Clone`-able struct — no per-invocation `AgentProvider` clone.
- **`LangModelFactory`**: `LangModelProvider::provide` now returns a cheap factory that can be cloned freely and turned into a `LangModel` lazily via `.make()`. This is what lets `AgentParts` defer the actual `LangModel` construction to the moment a sub-agent is actually invoked.
- **`Agent` no longer retains `AgentSpec`.** Files are pre-flattened across the sub-spec subtree at construction time; `skills` and `card_name` are lifted onto the runtime struct directly. Adds `Agent::files()` accessor. `Agent::spec()` is removed.
- **`AgentState`** moves to its own module (`agent/state.rs`).

### 5. `chore(agent): remove dead subagent file` (012a854)

`src/tool/impl/subagent.rs` was already dead (its `mod` declaration was commented out in step 4). Deleting the file and dropping the leftover commented-out lines in `src/tool/impl/mod.rs`.


src/runenv/sandbox.rs
```diff
@@ -1,468 +1,412 @@
...
-        #[serde(default)]
-        readonly: bool,
-    },
-    /// Memory-backed temporary filesystem. Disappears when the sandbox stops.
-    Tmpfs {
-        /// Absolute guest path.
-        guest: String,
-        /// Size limit in MiB. `None` means no limit.
-        size_mib: Option<u32>,
-    },
```
khj809, line 73: agent-k already depends on this volume mount for mounting project/session files. Why did you decide to remove this, and how you're planning to mount the files without this?

jhlee525: I recovered this.

grf53:
## 1. Merge `develop` (`23f979b`)

Kept runenv-revised's refactor as authoritative (name-based provider
registries, `LangModelAPISchema`-centralized wire dispatch,
`SandboxBuilder`/`Snapshot`) and re-applied develop's fixes on top.

| develop commit | Resolution |
|---|---|
| don't retry permanent quota 429s (#431) | Dropped the `QuotaClassifier` trait; **moved the behavior only** into `LangModelAPISchema::is_permanent_quota_error()`, wired into the `rt.rs` retry loop |
| gemini missing `content.role` (#426) | Restored the default-to-`Assistant` behavior |
| shell_tool() (#427) | Kept |
| web_fetch Firefox emulation (#428) | Kept + re-enabled the `wreq` `stream` feature |
| `Sandbox::exists()` (#430) | Ported onto runenv-revised's design (+ `remove_persisted()`) |
| microsandbox 0.5.6 (#429) | Not taken here (superseded by the 0.6.1 bump below) |

- develop's `QuotaClassifier` impls that leaked in via auto-merge
  (openai/anthropic/chat_completion/gemini) referenced the removed trait, so
  those files were reset to the HEAD versions.
- Verified: `cargo check` (default + `--features sandbox --all-targets`);
  quota-classifier unit tests pass.

## 2. Bump microsandbox to 0.6.1 (`5453f46`)
(referred to #432)

Adapted `src/runenv/sandbox.rs` to the 0.6.x API (behavior unchanged):

- `SandboxConfig` fields moved under `spec.{name, image, resources.cpus,
  resources.memory_mib, runtime.workdir, pull_policy, env, network, mounts}`.
- `spec.env` is now `Vec<EnvVar>` → map `(String, String)` via `EnvVar: From`.
- `spec.network.policy` is now `Option<Value>` → `disable_network` clears it.
- Tests: `SandboxHandle::status()` → `status_snapshot()`.
- `microsandbox-network` not needed, so not added.

grf53:
There was an opinion that is the name 'Machine' is a little awkward for its concept before.

I suggest some alternatives here.
- Host: pairs with `Console`. means the one 'hosts' the console executions.
- Runtime: means 'runtime'(an instance) of the 'runenv'(module name) that can execute `Console` on it.
- RunEnv: `runenv`(module name) itself, more concrete definition of runenv.
- Machine (no change): Since the name is not really a non-sense, it is sufficient to choose to proceed as is without change if other alternatives are not selected.

src/runenv/sandbox.rs
```diff
@@ -40,29 +42,16 @@ pub enum VolumeMount {
...
-        /// row, so a parallel scan never observes an orphan-shaped row.
-        ///
-        /// When `false` (default), the volume must already exist.
-        #[serde(default)]
-        create_if_missing: bool,
-        /// Labels attached to the volume when it is created. When reusing an
-        /// existing volume, microsandbox requires these to match the
-        /// persisted labels.
-        #[serde(default, skip_serializing_if = "Vec::is_empty")]
-        labels: Vec<(String, String)>,
```
khj809, line 65: This is not restored yet (introduced in https://github.com/brekkylab/ailoy/pull/430). Is this intended?

grf53: Accidentally missed this, thanks! I am working on it with some other missed #430 changes.

grf53: Applied. plz take a look!

khj809: Looks fine.

khj809:
@grf53 `test_archive_and_restore` is not passing in my environment. Did it pass in yours?

grf53:
> @grf53 `test_archive_and_restore` is not passing in my environment. Did it pass in yours?

It is taking quite long time to run that test. Let me wait and see the result.

khj809:
@grf53 I've just fixed it, but still takes longer, and it'll be fixed by https://github.com/superradcompany/microsandbox/pull/1150.

src/runenv/mod.rs
```diff
@@ -109,14 +109,10 @@
 //! # Ok(()) }
 //! ```
```
khj809, line 120: This doc is totally outdated, there's no `RunEnv` and `RunEnvHandle` anymore.

grf53:
> @grf53 I've just fixed it, but still takes longer, and it'll be fixed by [superradcompany/microsandbox#1150](https://github.com/superradcompany/microsandbox/pull/1150).

I also applied a fix for this at the same time. As I see, there is no reason to reject your fix, so let me follow that.

src/runenv/mod.rs
grf53: Just renewed the docstrings.

khj809:
Just fixed warnings (broken links) on `cargo doc` without `sandbox` feature.

grf53:
> Just fixed warnings (broken links) on `cargo doc` without `sandbox` feature.

Thanks for checking this.
(Good to know I can check docstring completeness through that...)

src/runenv/sandbox.rs
```diff
@@ -84,385 +81,485 @@ impl VolumeMount {
...
 
-    /// Stops the VM only; the VM definition is removed when the `Sandbox` is dropped.
-    async fn stop(&mut self) {
-        Self::stop_vm(&self.name).await;
+    async fn stop(&mut self) -> anyhow::Result<()> {
+        if let Some(console) = self.console.take() {
+            console.inner.stop_and_wait().await?;
+        }
+        Ok(())
     }
```
khj809, line 529: `stop()` calls `self.console.take()` **before** `stop_and_wait().await`. If `stop_and_wait` returns `Err` (e.g. a transient RPC/agentd error) the VM is still alive but the console handle is already gone, so:

- `is_running()` returns `false` (a lie — the VM is actually running)
- a later `start()` sees `console.is_none()` and calls `microsandbox::Sandbox::start(&self.name)` again → if the VM is still up this fails with `SandboxStillRunning`. There is no recovery path, so this `Sandbox` is stuck **permanently unusable** until it is dropped.

`stop_and_wait(&self)` borrows rather than consumes the console, so we can clear it **only after success**, preserving the handle and `is_running()` state on error:

```suggestion
    async fn stop(&mut self) -> anyhow::Result<()> {
        if let Some(console) = self.console.as_ref() {
            console.inner.stop_and_wait().await?;
        }
        self.console = None;
        Ok(())
    }
```

src/runenv/sandbox.rs
```diff
@@ -84,385 +81,485 @@ impl VolumeMount {
...
+                if let Ok(rt) = tokio::runtime::Builder::new_current_thread()
+                    .enable_all()
+                    .build()
+                    && let Err(e) = rt.block_on(console.inner.stop_and_wait())
+                {
+                    log::warn!("Drop stop_and_wait `{name}`: {e}");
+                }
+                let _ = tx.send(());
+            });
+            let _ = rx.recv();
```
khj809: `rx.recv()` blocks with no timeout. If the worker thread's `stop_and_wait()` never returns (e.g. a wedged agentd/libkrun), it never reaches `tx.send(())`, so `rx.recv()` blocks forever — and when `Drop` runs on a tokio worker or the process-exit path, **shutdown itself hangs**.

The old implementation on `develop` deliberately did not wait (`let _ = rx;`) for exactly this reason, and left `recv_timeout(30s)` as a candidate in the comments. Since the `run_msb_cli(["remove", "-f", ...])` call right below force-removes the VM anyway, the graceful-stop wait should be bounded:

```suggestion
            let _ = rx.recv_timeout(std::time::Duration::from_secs(5));
```

src/runenv/sandbox.rs
grf53: @jhlee525
While this code is out of the scope of my works(conflict resolve & sync), the proposal seems reasonable.
I could apply this, if you want.

src/runenv/sandbox.rs
grf53: @jhlee525 ditto for this.

src/runenv/sandbox.rs
jhlee525: Ok it looks reasonable. I'll apply it.

src/runenv/sandbox.rs
khj809: This seems not resolved yet.

grf53 (approved):

src/runenv/sandbox.rs
jhlee525: Sorry The comment and the patch goes out for somewhat reason.

I applied it.

khj809 (approved):

PR #426, opened 2026-06-05
fix(lang_model/gemini): default missing content.role to Assistant

### LINKED ISSUE #425: Gemini adapter: "Role not specified" when API response omits content.role
## Summary

The Gemini adapter (`src/lang_model/impl/api/gemini.rs`) fails to unmarshal responses whose `candidates[0].content` does not include a `role` field. The downstream `MessageDelta::finish()` then bails with `"Role not specified"`.

In practice, Gemini occasionally returns a response where `content.parts` is populated but `content.role` is absent — typically near `finishReason: STOP` on multi-turn tool-use conversations. This causes an `Agent::run` stream error mid-conversation that aborts the agent loop with otherwise-valid partial output thrown away.

## Reproduction

Observed on `gemini-2.5-flash` running an iterative ReAct agent against FinanceBench (50 questions, ~10 tool calls per turn). ~4% of QAs hit this error after 8-12 successful tool calls.

Example error trace:

```
stream error: Role not specified
```

The failed turn's tool-call sequence shows normal corpus-tool usage right up to the failure; no `tool_choice` override was set on the request, so the adapter is processing a default `auto` response.

Sample response that triggers the bug (sanitised):

```json
{
  "candidates": [
    {
      "content": { "parts": [ { "text": "..." } ] },
      "finishReason": "STOP"
    }
  ]
}
```

Note the missing `"role": "model"` inside `content`.

## Current behavior

`src/lang_model/impl/api/gemini.rs` (around line 402):

```rust
// Parse role
if let Some(r) = content.pointer("/role") {
    let s = r.as_str().ok_or_else(|| ...)?;
    let v = match s { ... "model" => Role::Assistant, ... };
    rv.role = Some(v);
}
// no else branch — rv.role stays None
```

Then `MessageDelta::finish` (`src/message/message_delta.rs:223`):

```rust
let Some(role) = role else {
    bail!("Role not specified")
};
```

## Suggested fix

For Gemini specifically, a missing `content.role` in a candidate response should default to `Role::Assistant` (i.e. `"model"` in Gemini terminology), since the candidate content is by Gemini API spec always the model's output.

```rust
if let Some(r) = content.pointer("/role") {
    // existing parsing
    rv.role = Some(v);
} else {
    // Gemini occasionally omits role on assistant-side candidates.
    // The candidate.content of a generateContent response is always
    // model output by spec, so default to Assistant.
    rv.role = Some(Role::Assistant);
}
```

This is a one-line behavioural change in `parse_candidate_content` and does not affect other providers.

## Impact

- Affects long-running agent loops on `gemini-2.5-flash` (likely all 2.x/3.x flash variants).
- Currently surfaces as an opaque stream error and discards partial conversation state.
- Workarounds at the call site are limited because the failure is in the deserialization path.

## Environment

- ailoy: develop
- Provider: Google `generativelanguage.googleapis.com/v1beta`
- Models observed: `gemini-2.5-flash`
- Adapter: `LangModelAPISchema::Gemini`

nuri-yoo:
## Summary

Closes #425.

The Gemini adapter occasionally received a candidate whose `content.parts` is populated but `content.role` is omitted (observed on `gemini-2.5-flash` near `finishReason: STOP` in multi-turn tool-use). `parse_candidate_content` had no `else` branch, leaving `MessageDelta.role` as `None`, so `MessageDelta::finish()` then bailed with `"Role not specified"` and aborted the agent loop mid-conversation, discarding otherwise-valid partial output.

`candidate.content` of a `generateContent` response is by spec always model output, so when `content.role` is absent we now default to `Role::Assistant` instead of leaving the field unset.

## Changes

- `src/lang_model/impl/api/gemini.rs`: add an `else` branch in `parse_candidate_content` to set `rv.role = Some(Role::Assistant)` when `content.role` is missing.
- Regression test `test_unmarshal_missing_content_role_defaults_to_assistant`: feeds the exact sanitized sample from the issue, asserts `delta.role == Some(Assistant)` and that `finish()` returns `Ok` rather than bailing.

## Test plan

- [x] `cargo test --lib --no-default-features 'lang_model::r#impl::api::gemini::tests::'` — 8 passed, 2 ignored (API-key gated).
- [x] Same with `-- --include-ignored` and `GEMINI_API_KEY` set — 10 passed, including the live `test_run_max_tokens` and `test_run_response_format_json_schema`.

### DIFF
--- a/src/lang_model/impl/api/gemini.rs
+++ b/src/lang_model/impl/api/gemini.rs
@@ -398,7 +398,7 @@ fn parse_candidate_content(candidate: &Value) -> anyhow::Result<MessageDelta> {
         .pointer("/content")
         .ok_or_else(|| anyhow::anyhow!("Missing content in candidate"))?;
 
-    // Parse role
+    // Gemini occasionally omits `content.role`; candidates are always model output.
     if let Some(r) = content.pointer("/role") {
         let s = r
             .as_str()
@@ -412,6 +412,8 @@ fn parse_candidate_content(candidate: &Value) -> anyhow::Result<MessageDelta> {
             other => bail!("Unknown role: {other}"),
         };
         rv.role = Some(v);
+    } else {
+        rv.role = Some(Role::Assistant);
     }
 
     // Parse parts
@@ -486,7 +488,7 @@ mod tests {
     use crate::{
         datatype::{Bytes, Value},
         lang_model::{LangModel, LangModelAPISchema, LangModelOptions, LangModelProviderElem},
-        message::{FinishReason, Message, Part, Role, TokenUsage},
+        message::{Delta, FinishReason, Message, Part, Role, TokenUsage},
         tool::{ToolDesc, ToolDescBuilder},
     };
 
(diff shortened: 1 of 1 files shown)


khj809 (approved):
Thanks for the fix.

PR #428, opened 2026-06-11
fix(web_fetch): switch to wreq Firefox emulation to bypass TLS bot detection

khj809:
## Summary

- Replace `reqwest::Client` with `wreq::Client` using `Emulation::Firefox135` in `web_fetch`
- Add `stream` feature to `wreq` dependency to enable `bytes_stream()` support
- Update network integration test URL to AccuWeather and fix the corresponding body assertion

## Background

The `reqwest` crate's `rustls` backend produces a distinctive TLS ClientHello fingerprint that Akamai and similar WAFs detect and block at the handshake level — before any HTTP-layer headers are even inspected. This caused `web_fetch` to fail with a transport-level error (`HTTP/2 INTERNAL_ERROR` or `ECONNRESET`) on sites protected by Akamai, even when using a browser-like User-Agent string.

The `web_search` Bing engine already solved the same problem using `wreq` with `Emulation::Firefox135`, which replicates the Firefox TLS handshake (cipher suites, extensions, GREASE values) to pass fingerprint-based bot detection. This PR applies the same fix to `web_fetch`.

## Changes

| File | Change |
|---|---|
| `Cargo.toml` | Add `features = ["stream"]` to `wreq` for `bytes_stream()` |
| `src/tool/impl/builtins/web_fetch.rs` | Swap `reqwest::Client` → `wreq::Client` with Firefox emulation; remove dead `USER_AGENT` constant |
| `src/tool/impl/builtins/web_fetch.rs` (tests) | Update network test URL to AccuWeather (분당구) and fix body assertion |

## Test plan

- [x] All 14 unit tests in `web_fetch` pass (`cargo test -p ailoy web_fetch`)
- [x] Network integration test `network_single_fetch_returns_200_and_body` passes against AccuWeather with `--include-ignored`
- [x] Network integration test `network_single_fetch_html_returns_raw_markup` still passes

### DIFF
--- a/src/tool/impl/builtins/web_fetch.rs
+++ b/src/tool/impl/builtins/web_fetch.rs
@@ -31,6 +31,7 @@ impl WebFetchState {
     fn new() -> Self {
         let client = Client::builder()
             .emulation(Emulation::Firefox135)
+            .redirect(wreq::redirect::Policy::limited(10))
             .timeout(Duration::from_secs(REQUEST_TIMEOUT_SECS))
             .build()
             .expect("wreq::Client builder cannot fail with these settings");
--- a/src/tool/impl/builtins/web_fetch.rs
+++ b/src/tool/impl/builtins/web_fetch.rs
@@ -4,8 +4,9 @@ use std::time::{Duration, Instant};
 
 use html_to_markdown_rs::{ConversionOptions, OutputFormat};
 use parking_lot::Mutex;
-use reqwest::Client;
 use url::Url;
+use wreq::Client;
+use wreq_util::Emulation;
 
 use crate::{
     datatype::Value,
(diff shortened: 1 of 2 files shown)


src/tool/impl/builtins/web_fetch.rs
```diff
@@ -31,10 +30,10 @@ struct WebFetchState {
 impl WebFetchState {
     fn new() -> Self {
         let client = Client::builder()
-            .user_agent(USER_AGENT)
+            .emulation(Emulation::Firefox135)
```
nuri-yoo, line 33: `wreq`'s `ClientBuilder` defaults to `redirect::Policy::none()` — it doesn't follow redirects at all (`wreq` v5.3.0 `src/client/http.rs:216`). `reqwest` defaults to following up to 10. So this swap silently regresses any URL that starts with a 301/302: `web_fetch("https://google.com")` will return the redirect placeholder body instead of the final `www.google.com` page.

The AccuWeather test URL happens to have no redirects (`curl --max-redirs 10 -w '%{num_redirects}'` = 0), so the integration test doesn't catch this.

To preserve parity with reqwest:

```suggestion
            .emulation(Emulation::Firefox135)
            .redirect(wreq::redirect::Policy::limited(10))
```

khj809: Verified against wreq v5.3.0 source (`src/client/http.rs:216`): `redirect_policy: redirect::Policy::none()` is indeed the hardcoded default in `ClientBuilder`. Fixed in b3d81ff by adding `.redirect(wreq::redirect::Policy::limited(10))` to match reqwest's default behavior.

nuri-yoo (approved):
LGTM!

PR #429, opened 2026-06-11
chore(deps): bump microsandbox to v0.5.6

nuri-yoo:
## Summary

Bumps the microsandbox dependency from `v0.5.2` to `v0.5.6` and adopts the SDK-surface changes shipped along the way. PR #418 took `v0.5.0` → `v0.5.2`; this picks up `v0.5.3`–`v0.5.6`.

Two breaking sandbox-SDK changes land between `v0.5.2` and `v0.5.6`:

- superradcompany/microsandbox#909 — `refactor(sdk)!: use builder flag for detached creation`. The `SandboxBuilder::create_detached()` shortcut is removed; both `create_registered` and `create_from_snapshot` switch to `.detached(true).create()`.
- superradcompany/microsandbox#938 — `feat(sdk): unify sandbox lifecycle APIs`. `stop_and_wait` and `remove_persisted` are gone from the connected `Sandbox`, and `SandboxHandle::stop` is now completion-aware (waits for the VM to reach Stopped before returning). The `Sandbox::new` / fork stop path and the force-stop branch in `start_and_connect` all move to `inner.stop().await`. `Sandbox::remove_persisted` simplifies to a `handle.stop()` (if running) + `handle.remove()` pair — the previous `connect()` + `remove_persisted()` two-step is no longer needed.

Two upstream features are picked up at the same time:

- superradcompany/microsandbox#801 (race-free per-sandbox named volumes, originally referenced as pending in PR #418) + superradcompany/microsandbox#950 (idempotent named volume mounts). The auto-injected `/tmp` volume previously needed a separate `Volume::create` followed by a `VolumeMount::Named` push, with a manual `Volume::remove` salvage path if the sandbox create failed. It now flows through a single `m.named_with(name, |n| n.ensure_exists().label(...))` call, so the volume row is inserted in the same DB transaction as the sandbox row and the salvage path disappears. `VolumeMount::Named` gains `create_if_missing: bool` and `labels: Vec<(String, String)>`, exposing the new mount mode to ailoy users.
- superradcompany/microsandbox#830 — `feat(sandbox): enforce sandbox name byte limit`. ailoy used to derive its own sandbox-name cap from `SUN_PATH_MAX` against the per-sandbox agent-socket path. superradcompany/microsandbox#820 already moved that socket to a fixed-length hash under `~/.microsandbox/run/agent/` back in `v0.5.2`, so the OS-path derivation was obsolete. The local `validate_sandbox_name` is removed and validation delegates to upstream `microsandbox::validate_sandbox_name`. The syntactic cap rises from ~100 OS-dependent chars to 128 UTF-8 bytes; see *Practical name length* below for the runtime cap on `v0.5.6` specifically.

The other `v0.5.x` deltas — msb-metrics sidecar (#845/#856), legacy-mount config acceptance (#859), per-sandbox labels and label-selectors (#864), guest-runtime metrics (#939), CLI mount-kind flags (#927), oci image archive/save (#862), explicit named-volume modes from the CLI (#950), DNS/runtime/network hardening — are transparent at this layer.

## Dep form

The `microsandbox = { version = "0.5.2", optional = true }` line moves to `"0.5.6"`. Every transitive `microsandbox-*` member follows in `Cargo.lock`.

## Practical name length on v0.5.6

The new test `test_name_at_128_bytes_passes_validation` only covers the syntactic validator. End-to-end spawn for a 128-byte sandbox name fails on `v0.5.6` with `sandbox process exited (exit status: 0) before agent relay became available`, because the upstream spawn path uses the sandbox name as the guest `MSB_HOSTNAME` fallback and Linux `sethostname(2)` caps at `__NEW_UTS_LEN = 64` bytes. The boot-error trail is `agentd: init failed: init error: sethostname(<name>): EINVAL: Invalid argument`.

That is purely an upstream issue and is being addressed in superradcompany/microsandbox#971 (`fix(sandbox): keep guest hostname within UTS limit when sandbox name exceeds it`), which derives a 64-byte hostname from the sandbox name when no explicit `hostname()` is set. Until that PR lands and we move past `v0.5.6`, ailoy callers should keep sandbox names ≤ 64 UTF-8 bytes or set an explicit hostname via the microsandbox API. The validator-level 128-byte ceiling is intentional: it tracks what upstream claims to support, and tightens automatically as soon as upstream closes the gap.

## Verification

`~/.microsandbox/{bin,lib}` were wiped so the `v0.5.6` prebuilt was cold-pathed through `microsandbox::setup::install` on the first `Sandbox::new`. The full sandbox test module was then run.

```
cargo test --release --all-features runenv::sandbox::tests -- --test-threads=1
...
test result: ok. 39 passed; 0 failed; 0 ignored; 0 measured; 299 filtered out; finished in 23.55s
```

The 39 tests are the 38 pre-existing ones plus a new `test_name_at_128_bytes_passes_validation` that exercises the upstream 128-byte boundary. `test_named_volume_persists_across_sandboxes` was extended to drive both an `existing`-mode mount and an `ensure_exists` reuse mount against the same volume in a single test, so the new `create_if_missing` path has live coverage.

`cargo check --all-features` and `cargo check --all-features --tests` also pass.

## Cross-reference

- superradcompany/microsandbox v0.5.3 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.3
- superradcompany/microsandbox v0.5.4 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.4
- superradcompany/microsandbox v0.5.5 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.5
- superradcompany/microsandbox v0.5.6 release: https://github.com/superradcompany/microsandbox/releases/tag/v0.5.6
- superradcompany/microsandbox#971 — pending upstream fix for the hostname-cap gap discussed above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,7 +51,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "axum",
- "base64 0.22.1",
+ "base64",
  "dedent",
  "derive_builder",
  "dirs",
@@ -81,7 +81,7 @@ dependencies = [
  "serde_json",
  "shellexpand",
  "strum 0.27.2",
- "strum_macros",
+ "strum_macros 0.27.2",
  "tar",
  "tempfile",
  "test-with",
@@ -215,9 +215,9 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.6.2"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
+checksum = "b7f43a50ac4fdca5df8e885c21b835997f0a1cdee65494a6847694a98652d9d8"
 dependencies = [
  "asn1-rs-derive",
  "asn1-rs-impl",
@@ -225,15 +225,15 @@ dependencies = [
  "nom",
  "num-traits",
  "rusticata-macros",
- "thiserror 1.0.69",
+ "thiserror 2.0.18",
  "time",
 ]
 
 [[package]]
 name = "asn1-rs-derive"
-version = "0.5.1"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "965c2d33e53cb6b267e148a4cb0760bc01f4904c1cd4bb4002a085bb016d1490"
+checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -456,6 +456,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a054912289d18629dc78375ba2c3726a3afe3ff71b4edba9dedfca0e3446d1fc"
 dependencies = [
  "aws-lc-sys",
+ "untrusted 0.7.1",
  "zeroize",
 ]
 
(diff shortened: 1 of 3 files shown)


khj809 (approved):
Thanks for many contributions on msb upstream!
Looks good, and verified all test cases pass well.

grf53 (approved):

PR #430, opened 2026-06-12
feat(sandbox): add `Sandbox::exists()`

khj809:
## Summary

Adds `Sandbox::exists(name)` — a lightweight, read-only probe that returns `true` if a persisted sandbox with the given name is already registered, and `false` otherwise (including on any error, e.g. when the microsandbox runtime is not installed).

## Changes

- `src/runenv/sandbox.rs`: new `pub async fn exists(name: &str) -> bool` on `Sandbox`, implemented as `MsbSandbox::get(name).await.is_ok()`

## Tests

Two new tests under `runenv::sandbox::tests` (requires `--features sandbox`):

- [x] **`test_exists_returns_false_for_unknown_name`** — a never-registered random name returns `false`
- [x] **`test_exists_lifecycle`** — verifies `true` immediately after sandbox creation, then `false` after `remove_persisted`

Run with:
```
cargo test -p ailoy --lib --features sandbox -- runenv::sandbox::tests::test_exists
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,7 +51,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "axum",
- "base64 0.22.1",
+ "base64",
  "dedent",
  "derive_builder",
  "dirs",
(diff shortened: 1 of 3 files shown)


nuri-yoo (approved):
Tested and all passes. Looks good to me.

PR #431, opened 2026-06-12
fix(lang_model): do not retry permanent quota 429s

nuri-yoo:
## Summary

Stops retrying 429 responses that represent permanent quota or credit exhaustion.

Transient rate limits still go through the existing retry/backoff path. Permanent quota failures now return the provider error immediately, so callers do not spend the full retry budget on an error that cannot recover within the request.

## Problem

`LangModel::run` treated every HTTP 429 as a retryable rate limit. That is correct for temporary throttling, but not for quota or credit exhaustion. OpenAI reports permanent quota exhaustion as a 429 with `error.type` / `error.code = "insufficient_quota"`, which meant ailoy would retry the same non-recoverable failure before surfacing it.

Gemini uses `RESOURCE_EXHAUSTED` for both cases. The transient form includes `google.rpc.RetryInfo`; quota exhaustion does not. Anthropic billing exhaustion is outside this path because it is reported as 402 rather than 429.

## Changes

- `src/lang_model/rt.rs`: read the 429 response body before deciding whether to retry.
- Add `is_permanent_quota_error(schema, body)` to classify provider-specific permanent quota failures:
  - OpenAI / ChatCompletion: `insufficient_quota` in `error.type` or `error.code`.
  - Gemini: `RESOURCE_EXHAUSTED` without a `google.rpc.RetryInfo` detail.
  - Anthropic: always retryable in the 429 path.
- Keep unparseable 429 bodies retryable, preserving the previous behavior when the provider response is not structured JSON.
- Add regression coverage for both the permanent-quota short-circuit and the existing transient-429 retry path.

## Test plan

- [x] `cargo check --manifest-path /Users/nryoo/work/ailoy-429/Cargo.toml`
  - `Finished dev profile [unoptimized + debuginfo]`
- [x] `cargo test --manifest-path /Users/nryoo/work/ailoy-429/Cargo.toml --lib permanent`
  - `test lang_model::rt::tests::test_is_permanent_quota_error ... ok`
  - `test lang_model::rt::tests::test_run_does_not_retry_on_permanent_429 ... ok`
  - `test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 294 filtered out; finished in 0.03s`
- [x] `cargo test --manifest-path /Users/nryoo/work/ailoy-429/Cargo.toml --lib test_run_retries_on_429`
  - `test lang_model::rt::tests::test_run_retries_on_429 ... ok`
  - `test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 295 filtered out; finished in 0.04s`

### DIFF
--- a/src/lang_model/rt.rs
+++ b/src/lang_model/rt.rs
@@ -4,8 +4,8 @@ use url::Url;
 use super::{LangModelAPISchema, LangModelProviderElem};
 use crate::{
     datatype::Value,
-    lang_model::{LangModelOptions, r#impl::api, r#impl::api::QuotaClassifier as _},
-    message::{Delta as _, Marshaled, Message, MessageOutput, Unmarshal as _},
+    lang_model::{LangModelOptions, r#impl::api},
+    message::{Delta as _, Marshaled, Message, MessageOutput},
     tool::ToolDesc,
 };
 
@@ -157,6 +157,7 @@ impl LangModel {
                 let body: serde_json::Value = body.clone().into();
 
                 // Send request with retry on 429 (rate limit)
+                let provider = api::provider_api(schema);
                 let client = reqwest::Client::new();
                 const MAX_RETRIES: u32 = 3;
                 let (status, response_text) = {
@@ -181,21 +182,7 @@ impl LangModel {
                                 .min(MAX_WAIT_SECS);
                             let text = response.text().await?;
                             // Permanent quota/credit exhaustion never recovers; don't retry.
-                            let permanent = match schema {
-                                LangModelAPISchema::Anthropic => {
-                                    api::AnthropicUnmarshal.is_permanent_quota_error(&text)
-                                }
-                                LangModelAPISchema::ChatCompletion => {
-                                    api::ChatCompletionUnmarshal.is_permanent_quota_error(&text)
-                                }
-                                LangModelAPISchema::Gemini => {
-                                    api::GeminiUnmarshal.is_permanent_quota_error(&text)
-                                }
-                                LangModelAPISchema::OpenAI => {
-                                    api::OpenAIUnmarshal.is_permanent_quota_error(&text)
-                                }
-                            };
-                            if permanent {
+                            if provider.is_permanent_quota_error(&text) {
                                 log::warn!("Quota exhausted (429), not retrying: {text}");
                                 last_status = Some(s);
                                 last_text = Some(text);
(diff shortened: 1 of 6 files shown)


src/lang_model/rt.rs
```diff
@@ -226,6 +233,34 @@ impl LangModel {
...
+/// Whether a 429 body is permanent quota exhaustion (don't retry) vs a transient rate limit.
+fn is_permanent_quota_error(schema: &LangModelAPISchema, body: &str) -> bool {
+    let Ok(json) = serde_json::from_str::<serde_json::Value>(body) else {
+        return false;
+    };
+    let error = &json["error"];
+    match schema {
+        LangModelAPISchema::OpenAI | LangModelAPISchema::ChatCompletion => {
+            error["type"] == "insufficient_quota" || error["code"] == "insufficient_quota"
+        }
```
khj809: Providers sharing the same ChatCompletion schema may have different error code (or different field name), but they cannot be distinguished in schema level.
I think the better way is to add a trait method such as `fn is_permanent_quota_error(&self, body: &str) -> bool` on Unmarshal struct of each provider, so `rt.rs` does not need to have a centralized function like this.

src/lang_model/rt.rs
```diff
@@ -446,4 +481,75 @@ mod tests {
...
+        let gemini_rate = r#"{"error":{"status":"RESOURCE_EXHAUSTED","details":[{"@type":"type.googleapis.com/google.rpc.RetryInfo","retryDelay":"34s"}]}}"#;
+        assert!(is_permanent_quota_error(&Gemini, gemini_quota));
+        assert!(!is_permanent_quota_error(&Gemini, gemini_rate));
+
+        // Anthropic 429 is always transient; billing exhaustion is a 402.
+        assert!(!is_permanent_quota_error(&Anthropic, r#"{"type":"error","error":{"type":"rate_limit_error"}}"#));
+
+        // Unparseable body is treated as transient (don't suppress retries).
+        assert!(!is_permanent_quota_error(&OpenAI, "not json"));
+    }
```
khj809: Related to the above comment, we can separate the test cases into each provider if `is_permanent_quota_error` is decentralized.

src/lang_model/rt.rs
nuri-yoo: This issue was discovered while testing openai, so other provider issue has been accidentally ignored.
Your point is reasonable so applied this (with test split) in latest commit [6f77718](https://github.com/brekkylab/ailoy/pull/431/commits/6f777185e70e2d305b70a97ef93de5e331650863)

src/lang_model/rt.rs
```diff
@@ -181,7 +181,21 @@ impl LangModel {
...
+                                }
+                                LangModelAPISchema::ChatCompletion => {
+                                    api::ChatCompletionUnmarshal.is_permanent_quota_error(&text)
+                                }
+                                LangModelAPISchema::Gemini => {
+                                    api::GeminiUnmarshal.is_permanent_quota_error(&text)
+                                }
+                                LangModelAPISchema::OpenAI => {
+                                    api::OpenAIUnmarshal.is_permanent_quota_error(&text)
+                                }
```
khj809: How about separating the mapping from schema to unmarshal, so it can be reused on the below `.unmarshal()` part?

nuri-yoo: Addressed in [0cb0b98](https://github.com/brekkylab/ailoy/pull/431/commits/0cb0b982b1d9aaf7373626a4eaf47e3f2f18fd83)

khj809 (approved):
LGTM!

PR #432, opened 2026-06-22
feat(vfs): external provider mounts (S3/Notion/GDrive) via host FUSE + sandbox forwarder

khj809:
## Summary

Adds `src/vfs`: a unified virtual filesystem over external providers (**S3, Notion, Google Drive**), mounted so the agent's existing `shell`/`read`/`write`/`grep` tools operate on provider paths transparently — **no custom agent tools**. Gated behind a new `vfs` feature; the default build is unaffected.

## How it works

A single `Vfs` owns the provider `Resource`s, routes virtual paths to them, and keeps credentials host-side. It is exposed through two FUSE frontends behind one builder surface (`AgentBuilder::vfs(VfsConfig)`):

- **Non-sandbox** — a host FUSE mount (`fuser` 0.17); the agent's host shell sees `/…/s3` directly.
- **Sandbox** — a token-authed host forward server + an in-guest forwarder that mounts `/mnt/vfs` and delegates every FUSE op back to the host over `host.microsandbox.internal`.

The agent holds one opaque `AgentVfs` that tears everything down on `Drop`. Non-filesystem-shaped domain writes (e.g. Notion page-create) are issued by writing JSON to a `/<mount>/.cmd/<op>` control path, routed to `Resource::command`.

**In-guest forwarder** — a static, dependency-free Rust binary (`crates/ailoy-vfs-forwarder`) that mounts `/dev/fuse` directly as root, needing no python/fuse3/apt (clean image mounts on first run, ~6.5 s vs ~30 s). Not committed: `build.rs` cross-compiles it to `<arch>-unknown-linux-musl` and embeds it via `include_bytes!` (bundled lld links it — no external toolchain).

**VM-restart self-healing (agent-k use case)** — a persisted sandbox is stopped while idle and its forwarder dies with it, so the mount is re-established on every agent attach: `ensure_mounted` reacquires the VM, functionally probes the mount through the *current* host server, and re-bootstraps the forwarder if the probe fails. All provider/forwarder I/O is bounded (fail-fast EIO, never a wedged mount), and sandbox acquire/reconnect is retried to ride over microsandbox's SQLite-state hang.

## Providers

- **S3** (`object_store`): read/write/readdir/stat/unlink + dir ops; multi-mount / multi-credential isolation (per-mount accessor, optional `key_prefix`, S3-compatible `endpoint`). `readdir` is name-sorted and drops the zero-byte dir-marker object.
- **Notion** (`reqwest`): hierarchical page tree; `page.json` emits a normalized schema from a depth-capped block tree + markdown renderer. Domain writes: `page-create`, `block-append`, `comment-add`.
- **Google Drive** (`reqwest` + OAuth refresh): mirrors the Drive folder hierarchy; Docs/Sheets/Slides surface as `.gdoc/.gsheet/.gslide.json` returning native Google API JSON (`documents.get` etc.), other files download as-is. Listings + exports cached. Domain write: `docs-append`.

## Verification

Live, credential-driven integration tests under `tests/vfs/*.rs` (`#[ignore]`, one `vfs` binary split by subsystem). `all_vfs()` mounts only providers whose env vars are present; each test creates and tears down its own fixture.

- **Providers** — direct adapter read/write, multi-mount isolation, domain `.cmd` writes, stat metadata + range/EOF handling.
- **Host FUSE** — agent-driven S3 round-trip (shell tool only); write/append/truncate/rename/rm via `std::fs`; interactive inspector.
- **Sandbox** — re-mount across VM restart (single + N concurrent agents), forwarder process-death recovery, concurrent-reader stress, `.cmd` through the forwarder.
- **Forwarder** — static binary end-to-end, write/unlink, large chunked reads.

```sh
set -a; . .env; set +a
cargo test --features "vfs sandbox" --test vfs -- --ignored --nocapture
```

## Notes

- `cargo build`, `--features vfs`, and `--features "vfs sandbox"` all compile clean; `cargo +nightly fmt` + `clippy` clean.
- Sandbox-only frontend code is gated under `src/vfs/sandbox/`; the host frontend uses `fuser` 0.17.


src/runenv/sandbox.rs
```diff
@@ -609,24 +689,66 @@ fn check_home_conflict(requested: &Path, current: Option<&std::ffi::OsStr>) -> a
...
+            .memory(config.memory_mib)
+            .pull_policy(PullPolicy::IfMissing)
+            .detached(true);
+        for (k, v) in &config.env {
+            builder = builder.env(k.as_str(), v.as_str());
+        }
+        if config.disable_network {
+            builder = builder.disable_network();
+        } else if config.allow_host_egress {
+            builder = builder.network(|n| n.policy(NetworkPolicy::allow_all()));
```
nuri-yoo: `allow_host_egress` opens all egress, not just the host. The field name and doc ("allow the guest to reach host services / the `host.microsandbox.internal` alias") promise a narrow grant, but `NetworkPolicy::allow_all()` lets the guest reach the entire internet/LAN. The forwarder only needs `host.microsandbox.internal`, so enabling VFS-in-sandbox silently gives the LLM-driven shell unrestricted network egress.

A host-only policy is available in the same 0.6.1 release (and is covered by `microsandbox-network`'s own `group_host_allow_overrides_private_deny_when_ordered_first` test: host allowed, other private destinations denied):

```rust
NetworkPolicy {
    default_egress: Action::Deny,
    default_ingress: Action::Allow,
    rules: vec![Rule::allow_egress(Destination::Group(DestinationGroup::Host))],
}
```

The `microsandbox` facade only re-exports `NetworkPolicy`, so this needs a direct `microsandbox-network = "0.6.1"` dep for `Action`/`Destination`/`DestinationGroup`/`Rule`. If that's not worth it, at least rename or redocument the field to say it opens full egress.

Same `allow_all()` is also in `create_from_snapshot` (line 771).

khj809: Thanks for the catch. `allow_host_egress` was indeed mapping to `NetworkPolicy::allow_all()` (full internet/LAN/loopback/metadata), so enabling VFS silently gave the guest shell unrestricted egress.

Rather than just narrow that one flag, I replaced the two mutually-exclusive bools (`disable_network` / `allow_host_egress`) with a single `SandboxNetwork` enum:
- `HostOnly`: host (`host.microsandbox.internal`) only, no internet/LAN. Least-privilege for VFS.
- `Public` (default): host + public internet, but not private LAN / loopback / link-local / cloud-metadata (169.254.169.254) / multicast.
- `Full`: everything (grant deliberately).

Each maps to an explicit `NetworkPolicy` (`default_egress: Deny` + `allow_egress(Group::Host)` [+ `Group::Public`]); `Full` is `allow_all()`. Since every variant allows host egress, VFS-in-sandbox works in all modes — no silent full-internet grant, and no build-time validation needed. I pulled in `microsandbox-network = "0.6.1"` for `Action`/`Rule`/`Destination`/`DestinationGroup` (the `microsandbox` facade only re-exports `NetworkPolicy`).

Verified live: `HostOnly` blocks a public connect (1.1.1.1:443) while the forwarder round-trip + agent-k re-mount flow work under both `HostOnly` and the default `Public`. (Note: `:53` is a bad egress probe — microsandbox intercepts DNS regardless of policy).

nuri-yoo (approved):
LGTM

grf53:
It looks good, as I saw. while I have a little doubt about the fact that it includes specific implementation methods for each provider.

Since Ailoy is a library designed to make agent implementations easy, it seems the thought is leading to the conclusion that "it feels awkward to include Notion-related implementations in Ailoy" because I could answer to the question "Is connecting Notion an essential part of agent implementation?" is "No."

I believe it would be more natural to provide an interface for 'vfs' and then handle the specific implementation methods for each provider on the 'user side' of Ailoy. Would this be feasible? I haven't looked into it deeply, so I can't be sure if it's possible, but it would be much better if it were.

khj809:
> It looks good, as I saw. while I have a little doubt about the fact that it includes specific implementation methods for each provider.
> 
> Since Ailoy is a library designed to make agent implementations easy, it seems the thought is leading to the conclusion that "it feels awkward to include Notion-related implementations in Ailoy" because I could answer to the question "Is connecting Notion an essential part of agent implementation?" is "No."
> 
> I believe it would be more natural to provide an interface for 'vfs' and then handle the specific implementation methods for each provider on the 'user side' of Ailoy. Would this be feasible? I haven't looked into it deeply, so I can't be sure if it's possible, but it would be much better if it were.

I agree with the fact that "connecting Notion is not an essential part of agent implementation", and this applies not only the VFS providers but also to other features already included in Ailoy (such as sandbox). Rust offers the option of "feature-gating" for such situations — those who do not need Notion can simply choose not to use that feature (this requires making the current `vfs` feature more fine-grained though).

It seems we have different views on "Ailoy is a library designed to make agent implementations easy". While there may be users who do not need Notion connectivity, there are certainly others who do (without looking too far, we ourselves, who are developing agent-k, are those users). I believe that providing many out-of-the-box features for such users is the direction that truly makes Ailoy as "a library designed to make agent implementations easy".

Of course, I agree with providing an abstract interface that allows users to customize the VFS provider according to their preferences. However, I don't believe that providing only that interface makes Ailoy "a library designed to make agent implementations easy".

jhlee525:
I concern that this PR hardcodes design decisions(for agent-k) into Ailoy.

- Whether to use FUSE at all.
- How FUSE should be mounted on the host.
- How FUSE should be mounted in the guest. (For example, the TCP server approach is ultimately a design choice made specifically for the MSB and agent-k use cases.)
- How providers such as Notion should be exposed as a filesystem.

My concern is that these design decisions are being made based on agent-k's specific use cases, which risks turning this into a feature primarily for agent-k rather than a general-purpose capability for Ailoy.

It can be one of (good) approaches to implement filesystem binding, unfortunately not reusable features for all agent development...

I'm thinking of splitting "interface" of putting external source features from implementation of "fuse". But i have not find proper solution for this. Please let me find an idea for this: Generalizing "external source" feature for our runenv system.

khj809:
Closing in favor of https://github.com/brekkylab/agent-k/pull/204

PR #434, opened 2026-06-29
feat(lang_model, agent): end-to-end streaming (SSE)

selloriwoo:
## Summary

Adds token-level streaming **alongside** the existing blocking path — purely additive, no breaking changes. Both modes are available at both layers:

|        | blocking (existing)     | streaming (new)              |
|--------|-------------------------|------------------------------|
| LangModel | `run` → `MessageOutput` | `run_stream` → `Stream<MessageDeltaOutput>` |
| Agent     | `run` → `Stream<MessageOutput>` | `run_stream` → `Stream<AgentEvent>` |

- `LangModel::run_stream` — requests an SSE response and frames events out of the byte stream into incremental `MessageDeltaOutput`s.
- `Agent::run_stream` — drives the same agentic loop but streams each model delta as `AgentEvent::Delta` for live rendering, while accumulating the turn into `AgentEvent::Message` (with tool results) to drive history/tool logic.

Implemented for **all four providers**: Anthropic, ChatCompletion (OpenAI chat/completions + Grok/DeepSeek/Kimi), OpenAI Responses, and Gemini. Token usage is reported while streaming for all of them.

## Backward compatibility

**The existing `run` methods are untouched and remain the default.** `run_stream` is added next to them, not in place of them:

- `LangModel::run` and `Agent::run` keep their exact signatures and behavior, so all current callers (sub-agents, tests, downstream crates) work as-is.
- `run` and `run_stream` share the same request building and 429/quota retry (`marshal_request` + `send_with_retry`, extracted in this PR); they differ only in how the response body is consumed (whole vs. SSE-framed).
- Callers opt into streaming explicitly by choosing `run_stream`; nothing forces a migration.

## How it works

- New `StreamUnmarshal` trait (supertrait of `ProviderApi`): parses a single SSE event into an incremental `MessageDeltaOutput`, distinct from the whole-response `Unmarshal`. Each provider implements it for its own wire shape:
  - Anthropic: `content_block_delta` etc.; usage merged across `message_start` / `message_delta`.
  - ChatCompletion: `choices[].delta` fragments + `[DONE]` sentinel; requests `stream_options.include_usage` and reads usage from the final empty-`choices` chunk.
  - OpenAI Responses: `output_text.delta` (text), `output_item.done` (function call), `reasoning_summary_text.delta` (thinking); `response.completed` reuses the full-response parser for finish_reason + usage.
  - Gemini: each `?alt=sse` chunk is a partial `GenerateContentResponse`, reusing `parse_candidate_content` per chunk.
- Marshals reflect a new `LangModelRequest.stream` flag — body `"stream": true` (Gemini switches to the `:streamGenerateContent?alt=sse` endpoint) plus an `Accept: text/event-stream` header.
- The SSE loop drains to end-of-body rather than stopping at the first `finish_reason`: ChatCompletion sends its usage in a final chunk *after* the finish_reason chunk, so stopping early would drop it. The body ends right after the terminal event, so this still exits promptly.
- `MessageDeltaOutput::accumulate` merges `TokenUsage` field-wise (providers split input/output token counts across events).
- **Anthropic stream errors are surfaced, not swallowed.** A mid-stream `error` event (e.g. `overloaded_error`, rate limit) fails with the server's own error type + message instead of a generic "unknown event type". Unrecognized (future) event types are skipped rather than aborting an otherwise-complete stream.

## Testing

- Per-provider SSE unit tests (text / tool-call / thinking / usage-merge) — no network.
- Anthropic SSE error handling: an `error` event surfaces the server's type + message; an unknown event type is ignored — no network.
- Gated end-to-end `run_stream` tests against each real API, asserting streamed deltas accumulate into the final message **and** token usage is populated.
- `Agent::run_stream` end-to-end test: streamed deltas accumulate to the final assistant message.
- Existing `run`-based tests still pass unchanged.
- Verified end-to-end through agent-k (live token streaming in the UI).

## Notes

- Streaming is incremental for the data you consume as it arrives (assistant text, thinking). Tool-call **arguments** are only usable once complete — the agent can't invoke a tool on partial JSON — so they're effectively delivered whole: Anthropic / ChatCompletion stream `arguments` fragments that accumulate back into one call, while OpenAI Responses (`output_item.done`) and Gemini hand over the finished call directly. Same end result either way.
- **Known limitation (tool-call merge):** streamed tool-call fragments are merged by arrival order (last open call), but the documented OpenAI-compatible contract is to concatenate arguments by the `tool_calls[].index` field, which is currently ignored. This holds in practice for all current providers (OpenAI, Anthropic, DeepSeek, Kimi, Grok stream calls sequentially); a backend that interleaves parallel calls by `index` would misroute fragments. Documented in `MessageDelta` with the fix path (carry `index` through `PartDelta::Function`) if it ever surfaces.


src/lang_model/impl/api/gemini.rs
```diff
@@ -340,15 +352,85 @@ impl super::QuotaClassifier for GeminiUnmarshal {
...
+        // it. Treat absence (or a refusal) as an empty delta.
+        let delta = match (&finish_reason, candidate.pointer("/content")) {
+            (Some(FinishReason::Refusal { .. }), _) | (_, None) => MessageDelta::default(),
+            _ => parse_candidate_content(&candidate)?,
+        };
+
+        // Gemini reports "STOP" even for tool calls; adjust when present.
+        if !delta.tool_calls.is_empty() && matches!(finish_reason, Some(FinishReason::Stop {})) {
+            finish_reason = Some(FinishReason::ToolCall {});
+        }
```
khj809: **[#1 / HIGH] Gemini can split `functionCall` and `finishReason` across SSE chunks → tool call silently lost**

This STOP→ToolCall adjustment only fires when a single chunk carries *both* a `functionCall` part and `finishReason`. Live testing passed because **Flash co-locates them**, but that is incidental — Gemini **2.5 Pro / 3.x / thinking** models stream the `functionCall` in one chunk and a content-less terminal `finishReason: STOP` chunk later (refs: [livekit/agents#4307](https://github.com/livekit/agents/issues/4307), [langchainjs#8454](https://github.com/langchain-ai/langchainjs/issues/8454), [Google AI forum](https://discuss.ai.google.dev/t/finishreason-stop-but-parts-is-missing-inside-candidate/99331)). When that happens, the functionCall chunk has `finish_reason = None` (adjustment skipped) and the STOP chunk has empty `tool_calls` (adjustment skipped); `MessageDeltaOutput::accumulate` then keeps the tool_calls but leaves `finish_reason = Stop` (`other.or(self)`), so the agent loop treats it as a normal stop and **never executes the tool**.

The fix belongs at the right altitude — on the fully-accumulated message, not per chunk. Since `MessageDeltaOutput::finish()` isn't touched by this PR I can't attach a one-click suggestion, but the change is small and provider-agnostic (no-op for OpenAI/Anthropic, which already report ToolCall). In `src/message/message_delta.rs`, `impl Delta for MessageDeltaOutput::finish()`:

```rust
fn finish(self) -> anyhow::Result<Self::Item> {
    let message = self.delta.finish()?;
    let mut finish_reason = self
        .finish_reason
        .ok_or_else(|| anyhow::anyhow!("finish_reason not specified"))?;
    // A turn that produced tool calls is a tool-call turn, even if the provider
    // reported a plain Stop. Gemini can deliver the functionCall part and the
    // terminal `finishReason: STOP` in separate SSE chunks (2.5 Pro / 3.x), so
    // promote on the accumulated message rather than per chunk.
    if matches!(finish_reason, FinishReason::Stop {})
        && message.tool_calls.as_ref().is_some_and(|tc| !tc.is_empty())
    {
        finish_reason = FinishReason::ToolCall {};
    }
    Ok(MessageOutput { message, finish_reason, usage: self.usage, depth: None, source_agent: None })
}
```

With that in place, this per-chunk block (lines 422-425) becomes redundant and can be deleted. Also consider treating `MALFORMED_FUNCTION_CALL` (Gemini 3 emits it in its own content-less terminal chunk) as a non-success finish rather than a clean stop.

src/agent/rt.rs
```diff
@@ -477,6 +490,84 @@ impl Agent {
...
+                        &self.tool_descs,
+                        &self.model_options,
+                    );
+                    while let Some(item) = delta_stream.next().await {
+                        let delta = item?;
+                        acc = acc.accumulate(delta.clone())?;
+                        yield AgentEvent::Delta(delta);
+                    }
+                }
+                let mut output = acc.finish()?;
```
khj809: **[#2 / HIGH] A stream that ends without a terminal `finish_reason` aborts the whole turn and poisons history**

`MessageDeltaOutput::finish()` hard-errors with `"finish_reason not specified"` when no terminal event arrived (dropped connection, provider quirk, or — combined with the Gemini EOF-flush path — a missing final chunk). `acc.finish()?` then propagates the error out of `run_stream`. Two consequences:

1. The turn fails even though usable text/tool deltas were already streamed and yielded to the UI.
2. **History is left corrupted:** the user `query` was pushed at the top of the turn, but the assistant message is only pushed *after* `acc.finish()?`. On the error path history ends with a dangling `User` turn; since the agent is reused, the next `run_stream` pushes a second `User` → two consecutive user messages, which most providers reject. One transient stream drop poisons the entire subsequent conversation.

Minimal guard for the common case (stream produced content but no terminal finish_reason):

```suggestion
                // If the stream ended without a terminal finish_reason (dropped
                // connection, provider quirk), treat end-of-stream as a Stop so a
                // partial-but-usable turn survives instead of aborting the loop
                // with "finish_reason not specified".
                if acc.finish_reason.is_none() && acc.delta.role.is_some() {
                    acc.finish_reason = Some(FinishReason::Stop {});
                }
                let mut output = acc.finish()?;
```

Follow-up (not in this suggestion): for the zero-chunk case (`role` never set), `finish()` still bails — there, roll back the just-pushed `query` (`self.state.history.pop()`) on the error path so the next turn isn't poisoned. The non-streaming `run` shares the same dangling-query shape and deserves the same treatment.

src/lang_model/impl/api/chat_completion.rs
```diff
@@ -266,6 +279,102 @@ pub struct ChatCompletionUnmarshal;
...
+            .pointer("/finish_reason")
+            .filter(|v| !v.is_null())
+            .map(Self::parse_finish_reason);
+
+        let mut delta = MessageDelta::new();
+
+        // role: present only on the first chunk.
+        if let Some(role) = choice.pointer("/delta/role").and_then(|v| v.as_str()) {
+            delta = delta.with_role(role.parse::<Role>().unwrap_or(Role::Assistant));
+        }
```
khj809: **[#3 / MEDIUM] Streaming path lacks the `Role::Assistant` default the non-streaming path has → role-less stream makes `finish()` bail**

Here role is set *only* when `/delta/role` is present (the first chunk). OpenAI itself always sends it, but several OpenAI-compatible backends this PR targets (DeepSeek / Kimi / local servers — see the comment in `message_delta.rs:160-164`) sometimes omit `role` on the first delta. Then no chunk ever sets a role and `MessageDelta::finish()` bails with `"Role not specified"` — even though the non-streaming `unmarshal` (chat_completion.rs:484-489) already defaults a missing role to `Assistant`. The two paths should agree:

```suggestion
        // role: usually present only on the first chunk. Default to Assistant
        // when a chunk omits it (some OpenAI-compatible backends do), mirroring
        // the non-streaming `unmarshal` default — otherwise a role-less stream
        // accumulates to a message with no role and `finish()` bails with
        // "Role not specified".
        let role = choice
            .pointer("/delta/role")
            .and_then(|v| v.as_str())
            .map(|r| r.parse::<Role>().unwrap_or(Role::Assistant))
            .unwrap_or(Role::Assistant);
        delta = delta.with_role(role);
```

src/lang_model/impl/api/gemini.rs
```diff
@@ -340,15 +352,85 @@ impl super::QuotaClassifier for GeminiUnmarshal {
...
+            .pointer("/finishReason")
+            .and_then(|v| v.as_str())
+            .map(Self::parse_finish_reason);
+
+        // Content is present on text/tool chunks; a finish-only chunk may omit
+        // it. Treat absence (or a refusal) as an empty delta.
+        let delta = match (&finish_reason, candidate.pointer("/content")) {
+            (Some(FinishReason::Refusal { .. }), _) | (_, None) => MessageDelta::default(),
+            _ => parse_candidate_content(&candidate)?,
+        };
```
khj809, line 427: **[#3 / MEDIUM] Same role-default gap on the Gemini path**

A finish-only or refusal chunk (e.g. a top-level `SAFETY` block: `{"candidates":[{"finishReason":"SAFETY"}], ...}`) produces `MessageDelta::default()` with `role = None`. If such a chunk is the *only* / first delta of the stream, the accumulated message has no role and `finish()` bails with `"Role not specified"` instead of yielding a usable refusal/stop. Candidates are always model output, so default the role here:

```suggestion
        let delta = match (&finish_reason, candidate.pointer("/content")) {
            // Finish-only or refusal chunk may omit content; still set the role
            // so a stream whose only/first chunk is finish-only doesn't
            // accumulate to a role-less message that makes `finish()` bail.
            (Some(FinishReason::Refusal { .. }), _) | (_, None) => {
                MessageDelta::new().with_role(Role::Assistant)
            }
            _ => parse_candidate_content(&candidate)?,
        };
```

src/lang_model/impl/api/openai.rs
```diff
@@ -269,6 +279,99 @@ impl super::QuotaClassifier for OpenAIUnmarshal {
...
+                    usage: parsed.usage,
+                }));
+            }
+            "error" => {
+                let msg = val
+                    .pointer("/message")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or("unknown error");
+                anyhow::bail!("OpenAI Responses stream error: {msg}");
+            }
```
khj809: **[#4 / MEDIUM] OpenAI Responses: `response.failed` is swallowed, and the `error` event loses its message**

Two problems in this terminal-event handling:

1. **`response.failed` finishes as a non-error with no `finish_reason`.** It is routed through the full `unmarshal`, whose status match maps `"failed"` to the `_ => None` arm — so `unmarshal_event` returns `finish_reason: None` and no error. The failure reason in `/response/error` is never read. If no other event set a finish_reason, the stream then bails downstream with the generic `"finish_reason not specified"` instead of the real cause.
2. **The `error` event reads the wrong field.** Live-testing an `insufficient_quota` failure surfaced only `"OpenAI Responses stream error: unknown error"` — `/message` was absent, so the real reason was lost. Reading `/error/message` as a fallback recovers it.

```suggestion
            // Terminal success/partial: reuse the full-response parser for
            // finish_reason + usage; streamed content is dropped (already sent).
            "response.completed" | "response.incomplete" => {
                let Some(resp) = val.pointer("/response") else {
                    return Ok(None);
                };
                let parsed = OpenAIUnmarshal.unmarshal(resp.clone())?;
                return Ok(Some(MessageDeltaOutput {
                    delta: MessageDelta::new(),
                    finish_reason: parsed.finish_reason,
                    usage: parsed.usage,
                }));
            }
            // A failed response carries its reason in `response.error`; surface
            // it instead of silently finishing with no finish_reason.
            "response.failed" => {
                let msg = val
                    .pointer("/response/error/message")
                    .and_then(|v| v.as_str())
                    .unwrap_or("response failed");
                anyhow::bail!("OpenAI Responses failed: {msg}");
            }
            "error" => {
                let msg = val
                    .pointer("/message")
                    .or_else(|| val.pointer("/error/message"))
                    .and_then(|v| v.as_str())
                    .unwrap_or("unknown error");
                anyhow::bail!("OpenAI Responses stream error: {msg}");
            }
```

khj809 (commented):
Adversarial review of the SSE streaming changes. Live-verified happy-path streaming (text + tool calls) works for Anthropic, Gemini, OpenAI Responses, and ChatCompletion (OpenAI + xAI). The inline comments below are non-happy-path correctness issues, each with a suggested fix:

- **#1 (HIGH)** Gemini split functionCall/finishReason chunks -> tool call lost (confirmed real for 2.5 Pro / 3.x).
- **#2 (HIGH)** stream with no terminal finish_reason aborts the turn and poisons history.
- **#3 (MED)** streaming lacks the Assistant role default the non-streaming path has (ChatCompletion + Gemini).
- **#4 (MED)** OpenAI Responses `response.failed` swallowed + `error` event message lost (observed live).

src/lang_model/impl/api/gemini.rs
selloriwoo: Fixed in [5adf588](https://github.com/brekkylab/ailoy/pull/434/commits/5adf588708cfc123a046ecbce805233cabbbede8)

src/agent/rt.rs
selloriwoo: Fixed in [aea9ef5](https://github.com/brekkylab/ailoy/pull/434/commits/aea9ef5863e7ef7979fc56e9e24a3fbc48bcbbd5) and [007faa6](https://github.com/brekkylab/ailoy/pull/434/commits/007faa6cd40f978154f6dc153ab8384e1f8ab5b4)

src/lang_model/impl/api/openai.rs
```diff
@@ -269,6 +279,111 @@ impl super::QuotaClassifier for OpenAIUnmarshal {
...
+                delta = delta
+                    .with_role(Role::Assistant)
+                    .with_tool_calls([PartDelta::Function {
+                        id,
+                        function: PartDeltaFunction::WithStringArgs { name, arguments },
+                    }]);
+            }
+            // Terminal: reuse the full-response parser on the embedded `response`
+            // for finish_reason + usage. Its content is dropped — already
+            // streamed by the delta events above.
```
khj809: Deprecated comments.

```suggestion
```

khj809 (approved):
LGTM with minor comment.

src/message/message_delta.rs
```diff
@@ -307,9 +322,20 @@ impl Delta for MessageDeltaOutput {
 
     fn finish(self) -> anyhow::Result<Self::Item> {
         let message = self.delta.finish()?;
```
nuri-yoo, line 413: This panics on a no-argument tool call over streaming.

`self.delta.finish()` calls `PartDelta::finish()` (part_delta.rs:283), which does `to_parsed_function().unwrap()`. For a `WithStringArgs` function that returns `None` whenever the accumulated `arguments` isn't valid JSON — and an empty string `""` isn't valid JSON, so it panics.

This is reachable on a **normal, complete** call, not just a truncated one. I ran a no-arg tool (`get_current_time`) against the live Anthropic API and the stream is:

```
content_block_start … "input":{}
content_block_delta … "partial_json":""     ← empty
message_delta … "stop_reason":"tool_use"
```

so the accumulated arguments stay `""` and `finish()` panics at part_delta.rs:283. I confirmed it end-to-end: driving `Agent::run_stream` with this tool over the real API panics (it unwinds straight through — the `match acc.finish()` arm doesn't catch a panic — crashing the turn and leaving the dangling user query in history). As a control I patched `PartDelta::finish()` and re-ran the same live agent loop: it goes green (tool call → tool executes → final answer), and reverting the patch reproduces the panic.

I also reproduced it via mid-stream truncation (`max_tokens` hit during tool args) on ChatCompletion and Anthropic. The blocking path is safe because it serializes the whole `input`/`arguments` object; only the streamed fragment-accumulation produces `""`.

The fix I verified has two parts in `PartDelta::finish()`: (a) treat empty/blank args as `{}` so a no-arg call succeeds, and (b) return an `Err` instead of `unwrap()` on genuinely malformed JSON so the agent's existing `match acc.finish()` handles it and rolls back. I checked that replacing the `unwrap()` with only an error isn't enough — every no-arg Anthropic tool call then fails cleanly instead of working; the empty-args → `{}` recovery is what actually restores it.

src/message/message_delta.rs
```diff
@@ -155,7 +155,22 @@ impl Delta for MessageDelta {
...
+        //
+        // KNOWN LIMITATION: this relies on arrival order, but the documented
+        // contract for OpenAI-compatible streaming is to concatenate arguments
+        // by the `tool_calls[].index` field — which is currently ignored here.
+        // A backend is spec-permitted to interleave parallel calls by `index`
+        // (or to omit the id on a new call's first fragment); either would
+        // misroute argument fragments into the wrong call. If that surfaces,
+        // carry `index` through `PartDelta::Function` and match on it here
+        // instead of comparing against `last()`.
         for part_incoming in other.tool_calls {
```
nuri-yoo, line 178: The known limitation above undersells the failure mode. If a backend interleaves parallel calls by `index`, the arrival-order merge doesn't just "misroute fragments into the wrong call" — one call loses all its arguments and the other accumulates concatenated garbage, which is then invalid JSON and **panics** in `finish()` (same part_delta.rs:283 path). I fed an interleaved synthetic stream through the accumulator and got `fa` ending with `arguments:""` and `fb` with `arguments:"{"x":{"y":1}2}"`, which panics.

Not reachable with any current provider though — I checked OpenAI's real wire and it sends parallel calls sequentially (index 0 fully, then index 1), so this stays latent. Worth noting it's a crash, not a silent misroute; the part-(b) fix from the other comment (error instead of `unwrap()`) downgrades it from a panic to a clean error, and full correctness still needs carrying `index` through `PartDelta::Function` as the comment says.

src/message/message_delta.rs
selloriwoo: That was an important point. Thank you!
[3849d6c](https://github.com/brekkylab/ailoy/pull/434/commits/3849d6c7e5c23690368e93e22fbbbc95cae93039)

src/message/message_delta.rs
selloriwoo: I've updated this comment to match the real failure mode.
[dc3d7bc](https://github.com/brekkylab/ailoy/pull/434/commits/dc3d7bc19e400a4d0d4d850e55bec80bfa95aa8b)

nuri-yoo (approved):
LGTM

src/lang_model/impl/api/mod.rs
```diff
@@ -22,24 +22,45 @@ pub trait QuotaClassifier {
...
+/// the final response for every provider (e.g. Anthropic `content_block_delta`
+/// events vs. a complete `content` array; ChatCompletion `choices[].delta` vs.
+/// `choices[].message`), so streaming needs its own parser.
+///
+/// `data` is the raw `data:` field of one SSE event. Control events that carry
+/// no delta (e.g. OpenAI's `[DONE]` sentinel, Anthropic `ping`) return
+/// `Ok(None)`. The default implementation reports that streaming is unsupported,
+/// so a provider opts in by overriding this method.
+pub trait StreamUnmarshal {
+    fn unmarshal_event(&self, _data: &str) -> anyhow::Result<Option<MessageDeltaOutput>> {
```
grf53: I found that `self` is immutable here for `StreamUnmarshal::unmarshal_event()`, while it is mutable(`&mut self`) in `Marshal::marshal()` and `Unmarshal::unmarshal()`. This is not critical, but it could be good to be consistent.

It appears that, contrary to the initial intention of making it mutable, `self` did not actually need to be mutable during implementation. Unless there is a compelling reason to maintain `mutable`, it would be better to standardize it to `immutable`.

Since this section was written by @jhlee525, I need your opinion.

src/agent/rt.rs
```diff
@@ -14,6 +14,19 @@ use crate::{
...
 
+/// An item emitted by [`Agent::run_stream`].
+///
+/// `Delta` carries an incremental model update (text / thinking / tool-argument
+/// fragments) for live rendering; `Message` is a completed turn — an assistant
+/// message or a tool result — equivalent to what [`Agent::run`] yields, and the
+/// unit consumers use to drive history/tool logic. A consumer that only wants
+/// the blocking behaviour can ignore `Delta` and read `Message`.
+#[derive(Clone, Debug)]
+pub enum AgentEvent {
```
jhlee525: Do we need this enum?

How about just returning `MessageDeltaOutput` for agent's `stream` function?

selloriwoo: `Deltas` alone can't drive the loop. Tool results come out of execute_tool_calls as complete MessageOutputs, so there's no delta form for them.
The loop also commits the finalized turn message, which is the accumulated delta after finish() with parsed parts plus `depth` and `source_agent` that the `delta type` doesn't have.
The stream surfaces that directly as Message so consumers don't have to re-accumulate. `run_stream` just adds the Delta channel on top of what `Agent::run` already returns.

jhlee525: Ok but returning delta is the most important feature in this PR.

* run -> Returning `MessageOutput`
* run_stream -> Returning `MessageOutputDelta`

Changing tool to return `MessageOutputDelta`, instead of `MessageOutput` can be a solution

src/lang_model/impl/api/mod.rs
jhlee525: 1. I agree with that point. `Marshal` and `Unmarshal` logic is based on `&mut self`, so removing mutability in this trait could force unnecessary `Arc` or `Mutex` wrapping.
2. In here, `&mut self` used since that a single SSE stream is an API that requires aggregation. We intentionally allowed mutation so implementations can update their internal state while performing that aggregation.
3. In addition, do we really need this trait? I noticed that unmarshaling "a whole response" and "streaming response" is not consolidated.

grf53: As discussed in person, we do not have a separate `StreamUnmarshal` trait but utilize the existing `Unmarshal` and mutability.

selloriwoo: dropped the separate StreamUnmarshal trait and folded unmarshal_event
into Unmarshal, keeping &mut self for consistency with unmarshal/marshal (and to
leave room for per-stream aggregation state without Arc/Mutex). ProviderApi now
just re-exposes both parsers for dynamic dispatch, since Unmarshal: Default isn't
object-safe.

[99c9632](https://github.com/brekkylab/ailoy/pull/434/commits/99c96328a8c54f4d9602bf9df190498f3e0be304)

selloriwoo:
## Follow-up since `2732302`

Quick summary of what changed and why, across the three commits.

**Trait cleanup (`99c9632`).** Per the thread, I dropped the separate `StreamUnmarshal` trait and just folded `unmarshal_event` into `Unmarshal`, keeping `&mut self` so it stays consistent with `unmarshal`/`marshal` (and leaves room for per-stream state later without wrapping in `Arc`/`Mutex`). `ProviderApi` is still there only because `Unmarshal: Default` can't be made into a trait object — it's a thin re-export for the dynamic dispatch, nothing more.

Same commit also swaps the `AgentEvent` enum for a single `MessageDeltaOutput` stream out of `run_stream`: you get the model's deltas for live rendering, and each tool result comes through as one complete delta. If you only care about finished messages, `into_messages` folds the stream back into the same `MessageOutput`s that blocking `run` gives you.

**The main fix (`cdccbc4`).** The tricky part was: what happens when a provider ends the stream without a `finish_reason`? Before, everyone downstream had to guess at where a message ended. So instead I close that gap at the source — `run_stream` now always ends a message with a `finish_reason`, synthesizing a `Stop` delta at EOF if the provider didn't send one. Now the rule is just "a message ends when you see a `finish_reason`," everywhere.

Because of that, `run_stream` no longer needs its own patch-up logic, and `into_messages` goes back to keying purely off `finish_reason`. (This replaces `5c2133a`, which tried to handle it further downstream — you can ignore that commit, it's superseded.)

A couple of things I was careful about, with tests to keep them from regressing:
- the synthesized `Stop` goes *after* the EOF flush, so we don't double up when the real finish shows up there (e.g. Gemini's finish+usage chunk);
- if the stream errors mid-way, we bail without synthesizing anything — otherwise a failed turn would look like it completed cleanly and get committed.

One caveat worth flagging: a clean finish and a mid-stream truncation both end up as `Stop` right now — we can't tell them apart without the provider's signal. If we want to, a dedicated `FinishReason` variant could distinguish them later.

**Manual check (multi-turn with a tool call, live against OpenAI):**
```
#1  Assistant  finish=-         tc=1   → assistant makes the tool call (args streamed)
#2  -          finish=ToolCall         → that turn closes
#3  Tool       finish=Stop             → tool result comes back as one delta
#4–13 Assistant finish=-               → the answer streams in token by token
#14 -          finish=Stop             → clean finish (OpenAI's own, nothing synthesized)

folding it back with into_messages → 3 messages:
  Assistant (ToolCall, 1 call) → Tool (Stop) → Assistant (Stop, "The temperature in Seoul is 25°C.")
```
Tokens stream live, the tool round-trips, the follow-up streams, and re-aggregating gives exactly what blocking `run` would. Tests green (299 passed), clippy clean.

grf53 (approved):

PR #435, opened 2026-06-30
feat(vfs): add Gmail provider

khj809:
## Summary

Adds **Gmail** as a VFS provider. Stacked on #432 (introduces `src/vfs`); base is `feat/vfs-provider-mounts`.

Agents use the existing `shell`/`read`/`write` tools on a Gmail path tree — no custom tools.

## Layout
```
<label>/<yyyy-mm-dd>/<subject>__<message-id>.gmail.json   # the email (processed JSON)
<label>/<yyyy-mm-dd>/<subject>__<message-id>/<filename>   # attachments (only if any)
```
- `cat <…>.gmail.json` → processed email (`from`/`to`/`cc`/`subject`/`date`/`body_text`/`snippet`/`labels`/`attachments`); attachment bytes from the sibling dir.
- A date dir narrows the Gmail query server-side (`after:/before:`); message-id is encoded in the filename (no shared index).
- `rm <…>.gmail.json` → Trash. Writes via control path: `.cmd/send`, `.cmd/reply`, `.cmd/reply-all`, `.cmd/forward`.

## How it works
- `accessor/gmail.rs` — Google OAuth (token refresh + 401 retry, mirroring gdrive) over the Gmail API.
- `resource/gmail.rs` — readdir/read/stat/unlink/command. Cheap stat (label cache; message files use a direct_io sentinel size; attachments report exact size). Dependency-free date math.
- Listings are capped at the newest 50, so `Resource::listings_complete()` (gmail = `false`) disables the cache's negative-ENOENT shortcut, and visited date dirs fold into the label listing incrementally (so older dates stay reachable and tab-completion offers them).

## Tests
- **Unit** (no creds): sanitize, message-id parse, date math, `process_message` shape, MIME encoding.
- **Live** (`tests/vfs/providers.rs`): labels/dates/read, send, reply/reply-all/forward, capped-listing negative-cache regression, incremental date folding.

Verified end-to-end with a `gmail.modify`-scoped token through the host FUSE mount: read, send, reply, reply-all, forward, older-date navigation, incremental folding.

## Note
Requires a refresh token whose scope includes Gmail (e.g. `https://www.googleapis.com/auth/gmail.modify`) — enabling the Gmail API alone is not enough.

### DIFF
--- a/src/vfs/accessor/gmail.rs
+++ b/src/vfs/accessor/gmail.rs
@@ -1,0 +1,221 @@
+use std::time::{Duration, Instant};
+
+use base64::Engine;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use tokio::sync::Mutex;
+
+const OAUTH_TOKEN_URL: &str = "https://oauth2.googleapis.com/token";
+const GMAIL_API_BASE: &str = "https://gmail.googleapis.com/gmail/v1";
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct GmailConfig {
+    pub client_id: String,
+    pub client_secret: String,
+    pub refresh_token: String,
+}
+
+/// Holds Google OAuth credentials (one refresh token) and a cached access
+/// token. Mirrors [`GDriveAccessor`](super::GDriveAccessor)'s OAuth flow; both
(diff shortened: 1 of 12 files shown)


khj809:
Will be applied on top of https://github.com/brekkylab/agent-k/pull/204

PR #436, opened 2026-07-14
Apply `SandboxNetwork` policies

grf53:
## Description
Add configurable guest network egress to microsandbox-backed sandboxes, so a VFS-in-sandbox forwarder can always reach the host while the guest shell's outward reach stays controllable.
Callers pick a `SandboxNetwork` posture on the builder; it is applied at creation and re-applied on archive restore.

## SandboxNetwork
A single egress policy per sandbox. The sandbox host (`host.microsandbox.internal`) is reachable in **every** variant — they differ only in outside reach:

- **HostOnly** — host only; public internet, LAN, loopback, link-local, and
  metadata denied. Least-privilege posture for VFS-in-sandbox.
- **Public** (default) — host + public internet; private LAN, loopback,
  link-local, cloud-metadata, and multicast denied.
- **Full** — unrestricted egress/ingress.

Each variant maps to a microsandbox `NetworkPolicy` built via `microsandbox_network::policy`.

## Changes
- Add `SandboxNetwork { HostOnly, Public, Full }` (`Default` = `Public`) and its mapping to a microsandbox `NetworkPolicy`.
- `SandboxBuilder::network(SandboxNetwork)` selects the posture; the builder default applies `Public`.
- `Sandbox::try_from_archive` restores with the default posture;
  `try_from_archive_with_network(path, network)` restores with an explicit one (an archive is only a filesystem snapshot and doesn't carry the policy).
- Add the `microsandbox-network` dependency under the `sandbox` feature.

## Tests
- Integration tests on a booted VM: `HostOnly` blocks a raw public-IP egress probe (`nc 1.1.1.1:443`) while `Public` allows it.
- Existing sandbox lifecycle tests (stop, clean drop, archive/restore, fork) pass.

## Notes
- Network policies are applied through the microsandbox 0.6.x runtime's config-fd launch path; an older runtime binary cannot apply them. (existing `~/.microsandbox` and binary in it or cargo build cache can cause conflict. if you have a problem, try to clean it before)

## Related PR
- #432: Referenced many parts as the functional base necessary to implement the mount function here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,7 @@ dependencies = [
  "jsonschema",
  "log",
  "microsandbox",
+ "microsandbox-network",
  "ordered-float 5.3.0",
  "parking_lot",
  "png",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ path = "src/lib.rs"
 
 [features]
 default = []
-sandbox = ["dep:microsandbox"]
+sandbox = ["dep:microsandbox", "dep:microsandbox-network"]
 
 [dependencies]
 anyhow = "1"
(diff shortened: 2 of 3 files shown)


khj809 (approved):

grf53:
Let me merge this quickly.

PR #437, opened 2026-07-20
chore(deps): refresh Cargo.lock to clear RustSec advisories in six crates

nuri-yoo:
## Summary

Lockfile-only refresh (`cargo update`, no `Cargo.toml` change) that clears every RustSec/OSV advisory resolvable within the current version requirements. Six crates in `Cargo.lock` carried open advisories whose fixes are SemVer-compatible with what `Cargo.toml` already allows:

| crate | locked | now | advisory |
|---|---|---|---|
| `anyhow` | 1.0.102 | 1.0.104 | RUSTSEC-2026-0190 — unsoundness in `Error::downcast_mut()` |
| `tar` | 0.4.45 | 0.4.46 | GHSA-3pv8-6f4r-ffg2 — PAX header desynchronization (direct dep) |
| `astral-tokio-tar` | 0.6.0 | 0.6.4 | RUSTSEC-2026-0112 / -0113 / -0145 — PAX header desync ×2 + `unpack_in` symlink chmod (via `microsandbox`, `sandbox` feature only) |
| `quinn-proto` | 0.11.14 | 0.11.16 | RUSTSEC-2026-0185 — remote memory exhaustion via unbounded out-of-order stream reassembly, CVSS 7.5 |
| `rustls-webpki` | 0.103.10 | 0.103.13 | RUSTSEC-2026-0098 / -0099 — two certificate name-constraint bypasses; RUSTSEC-2026-0104 — reachable panic in CRL parsing |
| `rand` | 0.8.5 / 0.9.2 | 0.8.7 / 0.9.5 | RUSTSEC-2026-0097 — unsound with a custom logger using `rand::rng()` |

The refresh also drops `proc-macro-error2` (RUSTSEC-2026-0173, unmaintained) from the graph as a side effect of `getset` moving to 0.1.7.

An OSV scan of the resulting lockfile shows the only remaining hits are the ones a lockfile refresh cannot reach, tracked in #422: `rmcp` 0.11.0 (fix is in 1.4.0, direct pin — see the unused-dependency cleanup PR), `hickory-proto` 0.26.0-alpha.1 (held by `microsandbox-network`; upstream fix in superradcompany/microsandbox#1186), `lru` 0.13.0 (held by `wreq`, which is already at its latest release), `rsa` (no upstream fix exists), and the unmaintained-crate notices for `bincode`, `fxhash`, `rustls-pemfile`.

Dependabot in this repo only watches npm, so cargo lockfile drift has no automated refresh; this is the manual pass.

[...]

Issue #422, opened 2026-05-29
[deps] stale dependency audit on develop


nuri-yoo:
## Summary

Audit of pinned dependencies on `develop`. Originally taken at HEAD `311233f` (2026-05-29); refreshed at HEAD `e3f9091` (2026-07-20). Reports only crates whose `Cargo.toml` pin actually blocks `cargo update` from picking up the latest. Excludes major-caret pins like `tokio = "1.0"` or `uuid = "1.18.0"` that are SemVer-compatible with current latest and resolve through the lockfile.

Changes since the original audit: no pin from the first pass has been bumped — every entry below is still stale, and several gaps widened (`infer` 0.19→0.22, `jsonschema` 0.46→0.48, `rmcp` latest crossed two major lines to 2.2.0). `getrandom` joined the stale list (0.4 released while the pin says 0.3), and a RustSec/OSV scan of `Cargo.lock` is added at the bottom.

Three buckets plus four apparently unused dependencies.

## Follow-up PRs (2026-07-20)

- #437 — lockfile-only `cargo update`; clears the six `cargo update`-resolvable advisory crates below (plus `proc-macro-error2` via the `getset` bump).
- #438 — Bucket 1 + Bucket 2 pin bumps (`infer`, `fancy-regex`, `strum`/`strum_macros`, `scraper`, `jsonschema`); the `selectors` bump also drops unmaintained `fxhash` from the graph.
- #439 — drops the four unused dependencies. Removing `rmcp` exposed that ailoy's own `tokio` dep was missing the `process` feature (`runenv/local.rs` uses `tokio::process::Command`; it compiled only through `rmcp`'s `transport-child-process` feature unification) — the PR declares it explicitly. Also clears the `rmcp` RUSTSEC-2026-0189 audit hit.
- superradcompany/microsandbox#1186 — upstream hickory 0.26.1 bump; reaches ailoy once a microsandbox release containing it is picked up.
- Bucket 3 (`schemars` 0.8 → 1.x) has no PR yet — it stays blocked on the agent-k `schemars`/`aide` lockstep described below.

## Bucket 1

Safe to bump. Public API used in ailoy is unchanged between pinned and latest. All call sites re-verified at `e3f9091`.

| crate | pinned | latest | API used in ailoy | notes |
|---|---|---|---|---|
[...]

PR #438, opened 2026-07-20
chore(deps): bump infer, fancy-regex, strum, scraper, jsonschema pins

nuri-yoo:
## Summary

Bumps the five `Cargo.toml` pins from #422's Bucket 1/2 — the ones whose pinned line blocks `cargo update` but whose API surface used by ailoy is unchanged between pinned and latest. Every call site was re-verified at the current `develop` before bumping:

| crate | pin | API used in ailoy | notes |
|---|---|---|---|
| `infer` | 0.16 → 0.22 | `infer::get(bytes)` + `kind.mime_type()` at `src/tool/impl/builtins/read.rs:21` | 0.17–0.22 only widen file-format coverage, so unknown-type bytes may now resolve to a more specific MIME |
| `fancy-regex` | 0.16 → 0.18 | `Regex::new` at `src/lang_model/impl/api/gemini.rs:78`, `src/tool/impl/builtins/glob.rs:228`, `:324` | new syntax additions only; default behavior of existing patterns unchanged |
| `strum` / `strum_macros` | 0.27 → 0.28 | `Display` / `EnumString` derives at `src/message/message.rs:15` | 0.28's `TryFrom`/`FromStr` infallibility change only applies with a default variant, which ailoy's enums don't have |
| `scraper` | 0.24 → 0.27 | `Html::parse_document`, `Selector::parse`, `ElementRef` across the 10 `web_search` engine parsers | internal `html5ever` 0.35 → 0.37 / `selectors` 0.31 → 0.38 bumps; public API unchanged |
| `jsonschema` | 0.29 → 0.48 | `jsonschema::validator_for` at `src/lang_model/options.rs:70`, error via `Display` | `validator_for` is unchanged; the 0.29 → 0.48 breaking changes (`Validator::validate` return shape, `ValidationOptions` builder, private `ValidationError` fields) sit outside this call site. Biggest semantic jump of the five — conformance fixes could change verdicts on edge-case schemas, which is why the `response_format` suites below matter |

Side effect worth having: the `selectors` bump drops `fxhash` (RUSTSEC-2025-0057, unmaintained) from the dependency graph entirely.

Not included: `schemars` 0.8 → 1.x (Bucket 3) stays behind — it rewrites three manual `JsonSchema` impls, changes the emitted draft from Draft 7 to 2020-12 on the provider wire, and has to land in lockstep with agent-k's `schemars`/`aide` bumps, so it needs its own coordinated PR.

The branch is rebased on top of #437's lockfile refresh, so the lockfile delta here stays requirement-driven only: the bumped crates and their transitive churn (the old `html5ever` 0.35 stack out, `jsonschema-regex` and friends in), nothing else.

Refs #422.

## Verification

All results below are from the rebased branch (on top of #437's merged lockfile refresh). `cargo check --all-targets` and `cargo check --all-features --all-targets` pass. The full suite was then run with live API keys (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, plus `XAI_API_KEY` / `DEEPSEEK_API_KEY` / `KIMI_API_KEY` for the provider registry), so all `test_with::env`-gated live tests executed — including the three `test_run_response_format_json_schema` tests (anthropic / chat_completion / gemini), which take a schema through the bumped jsonschema 0.48 `validator_for` and then through a real provider structured-output round trip. The strum-derived `Role` serialization rides every one of those live request/response cycles:

```
cargo test --all-features -- --skip runenv::sandbox
test result: ok. 318 passed; 0 failed; 13 ignored; 0 measured; 10 filtered out; finished in 17.80s
```

Real-VM sandbox suite:

```
cargo test --all-features runenv::sandbox -- --test-threads=1
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 331 filtered out; finished in 11.00s
```

The `#[ignore = "requires network"]` set is the risk surface of the scraper bump — all nine web_search engines parse live result pages through scraper 0.27 / html5ever 0.37 / selectors 0.38, plus both web_fetch tests, the aggregation test, and the sandbox + Anthropic `skill` benchmark:

```
cargo test --all-features -- --ignored --test-threads=1
test tool::…::web_search::engines::{bing,brave,duckduckgo,google,mojeek,naver,startpage,yahoo,yandex}::tests::test_search_returns_results ... ok  (9/9)
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 328 filtered out; finished in 39.72s
```

Focused offline suites over the remaining call sites: `cargo test tool::` (glob → fancy-regex, read → infer, web_search parsers) — 172 passed; `cargo test response_format` — 22 passed; `cargo test message` (strum `Display`/`EnumString` round-trips) — 9 passed; all 0 failed.

(The `--ignored` sweep also force-runs three ` ```ignore `-fenced pseudo-code doctests in `src/tool/func.rs` that contain `…` placeholders and are non-compiling by design; they fail identically on `develop` and are not part of any CI target.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,7 +57,7 @@ dependencies = [
  "dirs",
  "dotenvy",
  "encoding_rs",
- "fancy-regex 0.16.2",
+ "fancy-regex",
  "flate2",
  "futures",
  "getrandom 0.3.4",
@@ -81,8 +81,8 @@ dependencies = [
  "serde_bytes",
  "serde_json",
  "shellexpand",
- "strum 0.27.2",
- "strum_macros 0.27.2",
+ "strum 0.28.0",
+ "strum_macros 0.28.0",
  "tar",
  "tempfile",
  "test-with",
@@ -822,13 +822,13 @@ dependencies = [
 
 [[package]]
 name = "cfb"
-version = "0.7.3"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d38f2da7a0a2c4ccf0065be06397cc26a81f4e528be095826eee9d4adbb8c60f"
+checksum = "a347dcabdae9c31b0825fd6a8bed285ec9c2acb89c47827126d52fa4f59cece3"
 dependencies = [
- "byteorder",
  "fnv",
  "uuid",
+ "web-time",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


khj809 (approved):
👍

PR #439, opened 2026-07-20
chore(deps): drop unused dependencies (rmcp, which, zip, getrandom)

nuri-yoo:
## Summary

Removes the four `Cargo.toml` dependencies that no file in the repository references, identified in #422:

- `rmcp = "0.11.0"` — beyond being unused, 0.11.0 carries RUSTSEC-2026-0189 (DNS rebinding in the Streamable HTTP server transport, fixed in 1.4.0), so it shows up in every `cargo audit` run; and the crate is now two major lines ahead (2.2.0), so a hypothetical future MCP integration would not start from this pin anyway.
- `which = "7"` — latest is 8.x, zero call sites.
- `zip = "2"` — latest is 8.x, zero call sites.
- `getrandom = "0.3"` — declared with no features, so it was inert on every axis: no code references it, it enables nothing for the wasm backend cfg in `.cargo/config.toml` (that flag governs transitive `getrandom` instances and requires the `wasm_js` feature, which nothing enables), and it unified no versions (the lockfile carries transitive 0.2.x/0.3.x/0.4.x copies before and after this change). It arrived together with `.cargo/config.toml` in #401; the config file is left untouched.

Reproducer for "unused" (all four return zero matches repo-wide, not just `src/`):

```bash
git grep -rnE 'rmcp' -- ':!Cargo.toml' ':!Cargo.lock'
git grep -rnE 'which::|use which|extern crate which' -- '*.rs'
git grep -rnE 'zip::|ZipArchive|ZipWriter|use zip' -- '*.rs'
git grep -rnE 'getrandom' -- ':!Cargo.toml' ':!Cargo.lock' ':!.cargo/config.toml'
```

If any of these is a stub for planned work (`rmcp` for a native MCP client being the obvious candidate), happy to close this — though re-adding at the then-current major when the work actually starts costs one line and avoids carrying the audit hit in the meantime.

## One real finding: ailoy under-declared its tokio features

Removing `rmcp` broke the build — `src/runenv/local.rs:54` uses `tokio::process::Command`, but ailoy's own tokio dependency is `default-features = false` without the `process` feature. It only ever compiled because `rmcp`'s `transport-child-process` feature enabled `tokio/process` transitively, and Cargo feature unification leaked it to ailoy's build. This PR adds `process` to ailoy's tokio features, so the crate now declares what it actually uses instead of borrowing it from an unused dependency.

The branch is rebased on top of #437 and #438; against the current `develop` the lockfile drops 31 packages (`rmcp`, `which`, `zip` and their now-unreachable transitives — including the duplicate `schemars` 1.x line that only `rmcp` pulled in; ~425 lock lines).

Refs #422.

## Verification

All results below are from the rebased branch (on top of #437 + #438). `cargo check --all-targets` and `cargo check --all-features --all-targets` pass. The full suite was then run with live API keys (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, plus `XAI_API_KEY` / `DEEPSEEK_API_KEY` / `KIMI_API_KEY` for the provider registry), so every `test_with::env`-gated live test executed on the pruned dependency graph:

```
cargo test --all-features -- --skip runenv::sandbox
test result: ok. 318 passed; 0 failed; 13 ignored; 0 measured; 10 filtered out; finished in 14.58s
```

The sandbox suite matters doubly here — it spawns real microsandbox VMs and drives the `runenv` process paths whose `tokio/process` feature this PR moves from rmcp's feature unification onto ailoy's own manifest:

```
cargo test --all-features runenv::sandbox -- --test-threads=1
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 331 filtered out; finished in 11.94s
```

The `#[ignore = "requires network"]` set (nine web_search engine scrapes, both web_fetch tests, aggregation, and the sandbox + Anthropic `skill` benchmark): 12 of 13 pass; the `skill` benchmark hit a transient error in the batch run and passes in isolation. The one persistent failure is the mojeek scrape, and it is environmental, not from this PR — root cause confirmed by fetching `https://www.mojeek.com/search?q=ailoy` directly: mojeek is currently serving this IP an anti-bot CAPTCHA page (HTTP 200, `<title>Captcha`, altcha challenge widget) instead of results, after the whole engine suite ran repeatedly from here today. The parser correctly extracts zero results from that page, the same test passed three times earlier today, and the identical test fails with the identical `Expected at least one result` assert on the current `develop` HEAD (37fe8c8) run at the same time:

```
cargo test --all-features -- --ignored --test-threads=1
test result: FAILED. 11 passed; 2 failed  (batch)
cargo test --all-features test_benchmark_python_snippet_skill -- --ignored
test result: ok. 1 passed; 0 failed          (isolated re-run)
cargo test --all-features mojeek -- --ignored      # on develop 37fe8c8, same minute
test result: FAILED. 0 passed; 1 failed      (same assert — pre-existing, environmental)
```

(The `--ignored` sweep also force-runs three ` ```ignore `-fenced pseudo-code doctests in `src/tool/func.rs` that contain `…` placeholders and are non-compiling by design; they fail identically on `develop` and are not part of any CI target.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,7 +60,6 @@ dependencies = [
  "fancy-regex",
  "flate2",
  "futures",
- "getrandom 0.3.4",
  "hex",
  "html-to-markdown-rs",
  "image",
@@ -73,9 +72,8 @@ dependencies = [
  "ordered-float 5.3.0",
  "parking_lot",
  "png",
- "reqwest 0.13.4",
- "rmcp",
- "schemars 0.8.22",
+ "reqwest",
+ "schemars",
  "scraper",
  "serde",
  "serde_bytes",
@@ -92,10 +90,8 @@ dependencies = [
  "url",
  "urlencoding",
  "uuid",
- "which 7.0.3",
  "wreq",
  "wreq-util",
- "zip",
 ]
 
 [[package]]
@@ -195,15 +191,6 @@ name = "anyhow"
 version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
-
-[[package]]
-name = "arbitrary"
-version = "1.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
-dependencies = [
- "derive_arbitrary",
-]
 
 [[package]]
 name = "arc-swap"
@@ -1064,12 +1051,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "constant_time_eq"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c74b8349d32d297c9134b8c88677813a227df8f779daa29bfc29c183fe3dca6"
-
-[[package]]
 name = "convert_case"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


khj809 (approved):
MCP integration was not migrated from main for now, but it should also be supported eventually. The different point is that we don't need to consider wasm environments anymore, which makes the mcp integration extremely easy. Let me do this in the near future.

nuri-yoo:
> MCP integration was not migrated from main for now, but it should also be supported eventually. The different point is that we don't need to consider wasm environments anymore, which makes the mcp integration extremely easy. Let me do this in the near future.

Thanks for laying out the direction ahead. Looking forward to when this lands.

PR #440, opened 2026-07-27
chore(deps): bump microsandbox to v0.6.7

nuri-yoo:
Bumps the `microsandbox` and `microsandbox-network` pins from 0.6.1 to 0.6.7 and adapts `runenv/sandbox.rs` to the API changes that landed in between. The lockfile was already resolving to 0.6.6, so the manifest pins were stale; this moves both the pins and the lock to 0.6.7.

## Notable upstream changes between 0.6.1 and 0.6.7

- **Snapshot API contract finalized** (superradcompany/microsandbox#1118): `Snapshot::export`/`Snapshot::import` became `Snapshot::save`/`Snapshot::load`, and the builder signature changed. This is what breaks our `Sandbox::archive` / `try_from_archive` path (adaptation below).
- **Sparse snapshot export** (superradcompany/microsandbox#1150): files with holes are stored as GNU sparse tar entries on export, so the `.tar.zst` archives our `Sandbox::archive` produces no longer materialize the empty regions of the upper layer. That PR reports a 4 GiB mostly-empty upper going from ~6.5 s to ~0.02 s to bundle, with the archive shrinking from ~2 MB to ~7 KB. superradcompany/microsandbox#1166 and superradcompany/microsandbox#1200 follow up with an owned sparse codec and a bidirectional v0.6.6 format migration, so archives written by 0.6.6 still load.
- **Typed sandbox spec** (superradcompany/microsandbox#1144): `spec.network.policy` changed from `Option<serde_json::Value>` to a typed wire-format `NetworkPolicy` (adaptation below).
- **Network fixes and features**: IPv6 DNS edge cases (superradcompany/microsandbox#1083), scoped upstream TLS verification (superradcompany/microsandbox#1073), fragmented UDP / PMTU relay traffic (superradcompany/microsandbox#1086), composable network profiles (superradcompany/microsandbox#1198).
- **Runtime and lifecycle**: live modify/resize for running sandboxes (superradcompany/microsandbox#1099), exec kills the whole process group instead of just the direct child (superradcompany/microsandbox#1104), faster `msb load`/`pull` via an early cache gate and zlib-rs decode (superradcompany/microsandbox#1075).
- **Dependency hygiene**: hickory bumped to 0.26.1, migrating the discontinued `hickory-client` 0.26.0-alpha.1 to `hickory-net` (superradcompany/microsandbox#1186); this clears the hickory RustSec advisories from our lockfile as well.

## API adaptations in this PR

**Snapshot archive API rename** (`export`/`import` → `save`/`load`): `Snapshot::export(name_or_path, out, ExportOpts)` is now `Snapshot::save(name_or_path, out, SaveOpts)` and `Snapshot::import(archive_path, dest)` is now `Snapshot::load(archive_path, dest)`. The signatures and option fields are otherwise identical, so `Sandbox::archive` / `try_from_archive` keep their exact behavior.

**Snapshot builder argument order**: `Snapshot::builder(source_sandbox).name(name)` became `Snapshot::builder(name).from_sandbox(source_sandbox)`. `Sandbox::archive` passes the sandbox's own name for both, as before.

**Typed network policy in the spec**: `spec.network.policy` now stores `Option<microsandbox_types::NetworkPolicy>`, a wire-format type distinct from the engine-side `microsandbox_network::policy::NetworkPolicy` that `SandboxNetwork` builds (and that the sandbox builder's `.network(|n| n.policy(...))` closure still takes). The old `network_policy_value` serde helper is replaced by `wire_network_policy`, which converts engine → wire through the same serde round-trip microsandbox itself uses between its engine config and spec types. Since the facade re-exports neither the wire `NetworkPolicy` nor the `microsandbox-types` crate, `microsandbox-types` is added as an optional dependency of the `sandbox` feature, mirroring the existing `microsandbox-network` direct-dependency rationale documented in `Cargo.toml`.

## Snapshot integrity hashing cost and dev-profile mitigation

v0.6.7 also made the snapshot integrity digest mandatory (superradcompany/microsandbox#1200): create, save, and load each compute or verify a `sparse-sha256-v1` digest over the upper layer's full logical size, with holes hashed as synthesized zeros. The cost is O(logical size) CPU regardless of how much data the sandbox actually wrote — about 1.7 s per pass for the default 4 GiB upper on an M4 Pro in release builds. Filed upstream as superradcompany/microsandbox#1220: the save-time re-verification is redundant in the immediate create-then-save flow, and the docs still describe the hash as opt-in.

Under the dev profile this dominated test time: unoptimized sha2 hashes at ~130 MB/s, so `test_archive_and_restore` went from ~11 s on 0.6.1 to ~99 s on 0.6.7 (three ~32 s passes). This PR adds `[profile.dev.package.{sha2,digest,block-buffer}] opt-level = 3` to Cargo.toml, which restores hardware-speed hashing and brings the test back to ~5-7 s. Profile overrides only take effect when this workspace is built directly, so release builds and crates depending on ailoy are unaffected.

## Verification

- `cargo build --features sandbox`: passes.
- `cargo clippy` and `cargo clippy --features sandbox`: no new warnings (the two existing lib warnings in `agent/rt.rs` and `lang_model/impl/api/mod.rs` are untouched by this PR).
- `cargo test --features sandbox --lib` against the installed msb 0.6.7 runtime: effectively all tests pass (324/324 counting a flaky rerun; 17 ignored). All sandbox e2e tests pass, including the paths this PR touches (`test_archive_and_restore`, `test_fork`, `test_host_only_blocks_public_egress`, `test_public_allows_public_egress`, `test_exec`, `test_stop`). Known flake: `agent::rt::tests::test_delegate_to_subagent` (a live-LLM test gated on `OPENAI_API_KEY`, no overlap with this change) failed once in the full run and passed on a standalone rerun.
- `cargo test --features sandbox runenv::sandbox::tests::test_archive_and_restore` with the dev-profile override: 5.48 s (was ~99 s on 0.6.7 without it; ~11 s on 0.6.1).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,7 @@ dependencies = [
  "log",
  "microsandbox",
  "microsandbox-network",
+ "microsandbox-types",
  "ordered-float 5.3.0",
  "parking_lot",
  "png",
@@ -1586,24 +1587,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "66b7e2430c6dff6a955451e2cfc438f09cea1965a9d6f87f7e3b90decc014099"
 
 [[package]]
-name = "endian-type"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c34f04666d835ff5d62e058c3995147c06f42fe86ff053337632bca83e42702d"
-
-[[package]]
-name = "enum-as-inner"
-version = "0.6.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1e6a265c649f3f5979b601d26f1d05ada116434c87741c9493cb56218f76cbc"
-dependencies = [
- "heck 0.5.0",
- "proc-macro2",
- "quote",
- "syn 2.0.119",
-]
-
-[[package]]
 name = "enumflags2"
 version = "0.7.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


grf53 (approved):
It is a shallow observation, but in my view, it is sufficiently acceptable.

khj809 (approved):
Thanks!

nuri-yoo:
@grf53 There’s a minor TOML change in [cf6d61f](https://github.com/brekkylab/ailoy/pull/440/commits/cf6d61f069b3b2554245807e696e9e56a18dc85b). Could you please take a look?

grf53 (approved):

PR #441, opened 2026-07-28
chore(deps): migrate schemars and refresh stable dependencies

nuri-yoo:
## Summary

This PR performs the Ailoy side of the dependency modernization tracked by #422:

- migrates `schemars` from 0.8.22 to 0.9.0 and ports the three manual `JsonSchema` implementations;
- preserves the existing `Value`, `Bytes`, and `ResponseFormat` schema payloads with regression coverage;
- updates the remaining compatible direct dependencies, including `base64` 0.23, `jsonschema` 0.49, and `test-with` 0.17;
- resolves the existing Rust 1.97 Clippy warning baseline;
- migrates the actively used web client from `wreq` 5.3 / `wreq-util` 2.2 to exact v6 release-candidate pins;
- removes the advisory-affected `lru` 0.13 path from the resolved dependency graph.

The wreq migration is the only prerelease exception. After publishing wreq 5.3.0, upstream has continued publishing the active client line as successive v6.0 release candidates rather than cutting another stable 5.x or 6.0 release. Current development and the `lru` fix are available on that RC line. Both RC dependencies are pinned exactly so the lock graph cannot move to a later candidate without review.

Refs #422

## Dependency audit

| Dependency | Before | After | Migration notes | Verification |
| --- | --- | --- | --- | --- |
| `schemars` | 0.8.22 | 0.9.0 | Ports manual schemas to `Schema`, `SchemaGenerator`, and `json_schema!`; changes the URL feature from `url` to `url2`. | Schema contract tests and full test suite |
| `base64` | 0.22 | 0.23 | Compatible stable API refresh. | Check, Clippy, and tests |
| `jsonschema` | 0.48 | 0.49 | Compatible stable validator refresh. | Schema and response-format tests |
| `test-with` | 0.16 | 0.17 | Updates the test environment helper. | Full test suite |
| `wreq` | 5.3.0 | `=6.0.0-rc.29` | Adopts `Response::uri()` and explicitly preserves charset, OS-level proxy detection (`system-proxy`), plus gzip, Brotli, deflate, zstd, and streaming support. | Check, warnings-denied Clippy, `cargo tree -e features`, existing web fetch/search tests |
| `wreq-util` | 2.2.6 | `=3.0.0-rc.14` | Enables `emulation-compression`; v3 no longer enables the old compression behavior implicitly. | Full test suite and dependency-tree inspection |
| `lru` through wreq | 0.13.0 | 0.18.1 | Removes the active `IterMut` advisory path. The remaining `lru` instance is shared by wreq and `html-to-markdown-rs`. | `cargo tree -i lru`; RustSec scan |
| compatible transitive graph | older patches | current compatible patches | Refreshes html-to-markdown, Tokio, serde_json, tar, quinn, and the wreq transport stack. | Lockfile resolution, check, Clippy, tests |

No dependency declaration is removed.

### [`3b5d943`](https://github.com/brekkylab/ailoy/commit/3b5d9436d443bf5ba51b1c05a5f45bb73ab92aac) — chore(deps): migrate schemars from 0.8 to 0.9

- ports the manual `JsonSchema` implementations away from public 0.8 schema internals;
- preserves the serialized schema contracts:
  - `Value`: `{}`;
  - `Bytes`: `{"type":"string","format":"base64"}`;
  - `ResponseFormat`: `{"type":"object","required":["type"]}`;
- updates the Schemars URL feature to the 0.9-compatible `url2` name.

### [`f1a042a`](https://github.com/brekkylab/ailoy/commit/f1a042a54f32faebc2e0d417095e8f5bb9c2c5ff) — chore: resolve clippy warnings on Rust 1.97

- resolves the 42-warning lib-test baseline reported by current stable Clippy;
- applies mechanical idiom updates without removing behavior or tests;
- makes the warnings-denied gate pass.

### [`c2163c4`](https://github.com/brekkylab/ailoy/commit/c2163c4b27228cdbed3c0bd65c84a403462be3e1) — chore(deps): refresh remaining stable dependencies

- updates `base64`, `jsonschema`, and `test-with`;
- refreshes compatible transitive patches;
- retains Schemars 0.9 because the current stable Aide line used by agent-k still depends on it.

### [`6aca191`](https://github.com/brekkylab/ailoy/commit/6aca191dbf8f4a862ff9415102af564ad4ea5070) — chore(deps): migrate wreq stack to v6 RC

- pins `wreq =6.0.0-rc.29` and `wreq-util =3.0.0-rc.14` exactly;
- replaces the removed `Response::url()` accessor with `Response::uri()`;
- explicitly enables the compression and charset features previously supplied implicitly by wreq-util 2.x;
- enables `emulation-compression` so Firefox emulation continues to advertise a matching content-encoding profile;
- replaces the active `lru` 0.13 dependency with 0.18.1.

This commit changes no test source.

### [`6ed0447`](https://github.com/brekkylab/ailoy/commit/6ed04478b56784ea8e8ae39ad2c1b2ff8fa8c108) — chore: declare rust-version 1.95.0

Addresses review feedback. `test-with` 0.17 raises its own `rust-version` to 1.95.0, up from 1.88.0 in 0.16.x, and is the only crate in the resolved graph above 1.91. Building the test targets on an older toolchain failed with no declared minimum to point at, so cargo surfaced it as a compile error inside the dependency. Declares the requirement on `[workspace.package]` and inherits it into `[package]` via `rust-version.workspace = true`.

### [`53039e7`](https://github.com/brekkylab/ailoy/commit/53039e7905c2e43aba85af04433379df2407140c) — fix(deps): restore OS-level proxy detection under wreq 6

Addresses review feedback, and fixes a behavior regression introduced by `6aca191`. wreq 5 read the macOS network configuration through `macos-system-configuration`, which its `default` set enabled, and read the Windows registry unconditionally. wreq 6 gates both behind a single `system-proxy` feature absent from `default`, so the v6 pin silently dropped OS-level proxy detection on both platforms.

The path stays live regardless: `ClientBuilder` still defaults `auto_sys_proxy` to true and pushes a system proxy matcher. `Matcher::from_system` continues to read `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`, so only the OS-configured case was affected. Since this crate never sets a proxy explicitly, `web_fetch` and `web_search` would have failed with plain connection errors for anyone whose proxy is configured only through macOS System Settings or the Windows registry.

Enables `system-proxy`, and records above the line why each wreq feature is listed by name so a future pin bump can be audited against the default set. The lockfile gains only two dependency edges on `wreq` itself; no new package or version resolves, as `system-configuration` 0.7.0 and `windows-registry` 0.6.1 were already in the graph.

## Compatibility decisions

| Item | Selected | Newer candidate | Decision |
| --- | --- | --- | --- |
| Schemars | 0.9.0 | 1.x | Wait for a stable Aide release supporting Schemars 1; forcing it now would split or break the shared schema graph. |
| wreq | `=6.0.0-rc.29` | unreleased 6.0 stable | Since 5.3.0, upstream releases have continued on the v6.0 RC line; stable 5.3 retains the active `lru` advisory, whose fix is present in the maintained RC line. |
| wreq-util | `=3.0.0-rc.14` | unreleased 3.0 stable | Kept on the matching RC line and pinned exactly. |
| microsandbox | 0.6.7 | current stable | Retains upstream `bincode` and `rustls-pemfile` warnings until their owning projects migrate. |

No alpha, beta, nightly, or canary dependency is introduced.

## Security audit

The wreq migration removes the active `lru 0.13` unsoundness warning.

| Finding at PR head | Reachability | Disposition |
| --- | --- | --- |
| `rsa 0.9.10` timing-side-channel advisory | Lockfile-only; absent from the all-features inverse dependency tree | No patched stable release is available |
| `bincode 2.0.1` unmaintained | Active through microsandbox / msb-imago | Requires a versioned QCOW2 metadata migration upstream |
| `rustls-pemfile 2.2.0` unmaintained | Active through microsandbox image/network TLS handling | Requires an upstream microsandbox migration |

Current RustSec result:

- 1 vulnerability finding: `rsa`, lockfile-only;
- 2 warnings: `bincode` and `rustls-pemfile`, both unmaintained;
- no remaining `lru` advisory.

## Verification

- `cargo check --all-features --all-targets`
- `cargo clippy --all-features --all-targets -- -D warnings`
- `cargo test --all-features`
  - lib tests: 327 passed, 0 failed, 17 ignored
  - integration target: 11 passed, 0 failed, 4 ignored
  - doctests: 0 passed, 0 failed, 3 ignored
  - whole suite: 338 passed, 0 failed, 24 ignored
- RustSec scan: 1 vulnerability, 2 unmaintained warnings
- `cargo tree -i lru`: only `lru 0.18.1`
- `cargo tree --all-features -e features`: `wreq feature "system-proxy"` present
- `cargo metadata --no-deps`: `ailoy` reports `rust_version = 1.95.0`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,13 @@ unicode-segmentation = "1.12.0"
 url = { version = "2.5", features = ["serde"] }
 urlencoding = "2.1"
 uuid = { version = "1.18.0", features = ["v4"] }
-wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "zstd"] }
+# wreq 6 trimmed its default feature set down to webpki-roots and tokio-rt, so
+# everything else this crate relies on has to be requested by name. charset and
+# system-proxy were in wreq 5's defaults; the decompression features and stream
+# were already explicit; emulation-compression restores the content-encoding
+# profile that wreq-util 2.x applied on its own. Re-check this list whenever the
+# pin moves, since a feature leaving default is a silent behavior change.
+wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "system-proxy", "zstd"] }
 wreq-util = { version = "=3.0.0-rc.14", features = ["emulation-compression"] }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ sandbox = ["dep:microsandbox", "dep:microsandbox-network", "dep:microsandbox-typ
 anyhow = "1"
 async-stream = "0.3"
 async-trait = "0.1"
-base64 = "0.22"
+base64 = "0.23"
 dedent = "0.1.1"
 derive_builder = "0.20.2"
 fancy-regex = "0.18"
@@ -54,7 +54,7 @@ parking_lot = "0.12.4"
 png = "0.18"
 reqwest = { version = "0.13", features = ["form", "json", "stream"] }
 scraper = "0.27"
-jsonschema = { version = "0.48", default-features = false }
+jsonschema = { version = "0.49", default-features = false }
 schemars = { version = "0.9", features = ["preserve_order", "url2"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
@@ -79,7 +79,7 @@ wreq-util = "2.2.6"
 [dev-dependencies]
 axum = "0.8"
 dotenvy = "0.15"
-test-with = { version = "0.16.1", default-features = false, features = ["executable"] }
+test-with = { version = "0.17", default-features = false, features = ["executable"] }
 tokio = { version = "1.0", features = ["net"] }
 
 # microsandbox 0.6.7 computes a sparse-sha256 integrity digest over the upper
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,8 +73,8 @@ unicode-segmentation = "1.12.0"
 url = { version = "2.5", features = ["serde"] }
 urlencoding = "2.1"
 uuid = { version = "1.18.0", features = ["v4"] }
-wreq = { version = "5.3.0", features = ["stream"] }
-wreq-util = "2.2.6"
+wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "zstd"] }
+wreq-util = { version = "=3.0.0-rc.14", features = ["emulation-compression"] }
 
 [dev-dependencies]
 axum = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ exclude = [
 [workspace.package]
 version = "0.3.0"
 edition = "2024"
+rust-version = "1.95.0"
 license = "Apache-2.0"
 repository = "https://github.com/brekkylab/ailoy"
 
@@ -16,6 +17,7 @@ repository = "https://github.com/brekkylab/ailoy"
 name = "ailoy"
 version.workspace = true
 edition.workspace = true
+rust-version.workspace = true
 license.workspace = true
 
 [lib]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ png = "0.18"
 reqwest = { version = "0.13", features = ["form", "json", "stream"] }
 scraper = "0.27"
 jsonschema = { version = "0.48", default-features = false }
-schemars = { version = "0.8", features = ["preserve_order", "url"] }
+schemars = { version = "0.9", features = ["preserve_order", "url2"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
 serde_json = { version = "1.0", features = ["preserve_order"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,25 +182,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "antidote"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34fde25430d87a9388dadbe6e34d7f72a462c8b43ac8d309b42b0a8505d7e2a5"
-
-[[package]]
 name = "anyhow"
 version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
-
-[[package]]
-name = "arc-swap"
-version = "1.9.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c049c0be4daef0b145cb3555416b3b8ef5b7888a38aea1a3a155801fe7b0810b"
-dependencies = [
- "rustversion",
-]
 
 [[package]]
 name = "asn1-rs"
@@ -649,35 +634,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "boring-sys2"
-version = "4.15.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab6c876a958d17057d9d3a31075bb9609237413a1fe665432782c31ef596eb6b"
-dependencies = [
- "autocfg",
- "bindgen",
- "cmake",
- "fs_extra",
- "fslock",
-]
-
-[[package]]
-name = "boring2"
-version = "4.15.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e4470e96bd94533c2f88c08be95a8e6d2d37a3b497a773b0a46273a376978f00"
-dependencies = [
- "bitflags 2.13.1",
- "boring-sys2",
- "brotli",
- "flate2",
- "foreign-types",
- "libc",
- "openssl-macros",
- "zstd",
-]
-
-[[package]]
 name = "borrow-or-share"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -702,6 +658,31 @@ checksum = "3a32acac15fe1967bc3986b2a6347dffc965602354ea6f450ad07e8bfd253583"
 dependencies = [
  "alloc-no-stdlib",
  "alloc-stdlib",
+]
+
+[[package]]
+name = "btls"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c5e60b8c8d282c86360cab651ded04ab0335a7b5390c8d34145cbeab8cacf5f"
+dependencies = [
+ "bitflags 2.13.1",
+ "btls-sys",
+ "foreign-types",
+ "libc",
+ "openssl-macros",
+]
+
+[[package]]
+name = "btls-sys"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b1b8638a2e1c38a5ae4efa90ae57e643baec35a30d03fc5b399b893adc4954b"
+dependencies = [
+ "bindgen",
+ "cmake",
+ "fs_extra",
+ "fslock",
 ]
 
 [[package]]
@@ -859,7 +840,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-link 0.2.1",
+ "windows-link",
 ]
 
 [[package]]
@@ -2208,7 +2189,7 @@ dependencies = [
  "bitflags 2.13.1",
  "html-escape",
  "html5ever",
- "lru 0.18.1",
+ "lru",
  "memchr",
  "once_cell",
  "phf 0.14.0",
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -73,13 +73,13 @@ unicode-segmentation = "1.12.0"
...
-wreq = { version = "5.3.0", features = ["stream"] }
-wreq-util = "2.2.6"
+wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "zstd"] }
+wreq-util = { version = "=3.0.0-rc.14", features = ["emulation-compression"] }
 
 [dev-dependencies]
 axum = "0.8"
 dotenvy = "0.15"
-test-with = { version = "0.16.1", default-features = false, features = ["executable"] }
+test-with = { version = "0.17", default-features = false, features = ["executable"] }
```
khj809, line 90: Accidentally found that the latest `test-with` is available on rustc 1.95.0+ (https://github.com/yanganto/test-with/blob/main/Cargo.toml#L15). I was using rustc 1.92.0 and failed to compile it. If it's really necessary to update this, we should also declare `rust-version = "1.95.0"` same as what `test-with` did.

Cargo.toml
```diff
@@ -73,13 +73,13 @@ unicode-segmentation = "1.12.0"
 url = { version = "2.5", features = ["serde"] }
 urlencoding = "2.1"
 uuid = { version = "1.18.0", features = ["v4"] }
-wreq = { version = "5.3.0", features = ["stream"] }
-wreq-util = "2.2.6"
+wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "zstd"] }
```
khj809: **`system-proxy` is missing — this silently drops macOS/Windows system proxy detection.**

The commit message says this "preserve[s] the previous compression and charset behavior explicitly", and it does — but the *proxy* behavior didn't come along, because v6 moved that behind a feature gate too.

| | wreq 5.3.0 | wreq 6.0.0-rc.29 |
|---|---|---|
| macOS | `macos-system-configuration` — **in `default`** ✅ | `system-proxy` — **not in `default`** ❌ |
| Windows | `#[cfg(target_os = "windows")]` — **unconditional** ✅ | `system-proxy` — **not in `default`** ❌ |

`wreq-6.0.0-rc.29/src/proxy.rs`:

```rust
#[cfg(all(target_os = "macos", feature = "system-proxy"))]
#[cfg(all(windows,             feature = "system-proxy"))]
```

And this path is live by default — `client.rs:483`:

```rust
if config.auto_sys_proxy {
    config.proxies.push(ProxyMatcher::system());
}
```

`Matcher::from_system()` still calls `from_env()` unconditionally, so `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` keep working. What's lost is the OS-level configuration: macOS System Settings → Network → Proxies, and the Windows registry (`windows-registry`, which v5 read unconditionally).

**Impact:** `ailoy` never configures a proxy explicitly, so it relies entirely on this auto-detection. Anyone behind a corporate proxy configured only through the OS UI — no `*_PROXY` env vars — will see `web_fetch` and `web_search` fail with connection errors, with nothing pointing at the proxy as the cause.

Worth noting that `Cargo.lock` gives no signal here: the lockfile is feature-agnostic and lists optional dependencies regardless of whether the gate is on, so `system-configuration` and `windows-registry` appear in the graph either way. `cargo tree -e features` is what actually shows it:

```
$ cargo tree --all-features -e features | grep -oE 'wreq feature "[a-z-]+"' | sort -u
wreq feature "brotli"
wreq feature "charset"
wreq feature "default"
wreq feature "deflate"
wreq feature "gzip"
wreq feature "stream"
wreq feature "webpki-roots"
wreq feature "zstd"          # ← no system-proxy
```

Adding the feature is enough to restore v5 parity (verified — the feature shows up in the tree afterwards, and it produces zero `Cargo.lock` churn for the reason above):

```suggestion
wreq = { version = "=6.0.0-rc.29", features = ["brotli", "charset", "deflate", "gzip", "stream", "system-proxy", "zstd"] }
```

Also — since the whole point of this hunk is "v6 dropped things from `default` that v5 had", a short comment above the line recording *why* each of these is listed explicitly would make the next `default` change much easier to audit.

nuri-yoo: Verified against both crate sources — the regression is real and the diagnosis is exact. Fixed in 53039e7.

`wreq` 5.3.0 `[features]`:

```toml
default = ["webpki-roots", "charset", "macos-system-configuration"]
macos-system-configuration = ["dep:system-configuration"]
```

`wreq` 6.0.0-rc.29:

```toml
default = ["webpki-roots", "tokio-rt"]
system-proxy = ["dep:system-configuration", "dep:windows-registry"]
```

So macOS lost detection because `macos-system-configuration` was in v5's `default` and its v6 successor is not. Windows lost it for the separate reason you identified — v5's registry read was gated on the target alone, with no feature in the way:

```
wreq-5.3.0/src/proxy.rs:32:  #[cfg(target_os = "windows")]
wreq-5.3.0/src/proxy.rs:1213: #[cfg(target_os = "windows")]
```

against v6 requiring the feature on both:

```
wreq-6.0.0-rc.29/src/proxy.rs:1: #[cfg(all(target_os = "macos", feature = "system-proxy"))]
wreq-6.0.0-rc.29/src/proxy.rs:5: #[cfg(all(windows, feature = "system-proxy"))]
```

And the path is live by default, as you noted — `client.rs:284` sets `auto_sys_proxy: true`, `client.rs:482` acts on it. `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` kept working throughout, so the lost case is exactly the OS-configured one.

Applied your suggestion, plus the comment you asked for above the line — recording which features came from v5's `default` versus which were already explicit, so the next pin bump can be diffed against the default set rather than re-derived:

```toml
# wreq 6 trimmed its default feature set down to webpki-roots and tokio-rt, so
# everything else this crate relies on has to be requested by name. charset and
# system-proxy were in wreq 5's defaults; the decompression features and stream
# were already explicit; emulation-compression restores the content-encoding
# profile that wreq-util 2.x applied on its own. Re-check this list whenever the
# pin moves, since a feature leaving default is a silent behavior change.
```

`cargo tree --all-features -e features` now resolves it:

```
wreq feature "stream"
wreq feature "system-proxy"
wreq feature "zstd"
```

One small correction to the lockfile note: it is near-zero churn rather than zero. No new package or version resolves, since `system-configuration` 0.7.0 and `windows-registry` 0.6.1 were already in the graph as you said — but `wreq`'s own dependency list gains the two edges:

```diff
@@ wreq dependencies @@
+ "system-configuration",
+ "windows-registry",
```

Verification on this branch: `cargo check --all-features --all-targets` and `cargo clippy --all-features --all-targets -- -D warnings` both clean, and the macOS `system-proxy` code compiled in rather than being gated out. `cargo test --all-features` green at 338 passed / 0 failed / 24 ignored across three test binaries — lib 327/17, integration target 11/4, doctests 0/3.

One thing worth flagging rather than leaving implied: the first `cargo test` run in this session had 1 failure out of 344. It did not recur in five subsequent full-suite runs and I lost the test name before capturing it, so I can only report it as unreproduced, not as diagnosed.

I also updated the PR body's test breakdown while here. Its totals were correct — the "doctests: 11 passed, 7 ignored" line was summing the integration target and the doctests under one label, so it now lists the three binaries separately.

Cargo.toml
nuri-yoo: Confirmed, and thanks for catching it from the compile failure rather than the diff — nothing in the hunk shows it.

`test-with` 0.17.0 declares `rust-version = "1.95.0"`, up from `1.88.0` in both 0.16.1 and 0.16.3. It is also the only crate in the resolved graph above 1.91:

```
$ cargo metadata --all-features --format-version 1 \
    | jq -r '.packages[]|select(.rust_version)|"\(.rust_version) \(.name) \(.version)"' \
    | sort -Vr | head -4
1.95.0 test-with-derive 0.17.0
1.95.0 test-with 0.17.0
1.95.0 ailoy 0.3.0          # this PR's declaration, added below
1.91 smoltcp 0.13.1
```

Declared it on the workspace in 6ed0447, inherited into `[package]` via `rust-version.workspace = true`:

```toml
[workspace.package]
rust-version = "1.95.0"
```

`cargo metadata --no-deps` now reports `rust_version = 1.95.0` for `ailoy`, so cargo reports a toolchain mismatch against the manifest instead of failing inside a dependency's expansion.

On whether the bump was necessary: strictly, no — `test-with` is only used as `#[test_with::env(...)]`, and 0.17 adds nothing this crate consumes. Reverting to 0.16.3 would have held the graph at 1.91 and kept your 1.92.0 building. I kept 0.17 and declared the MSRV instead, so this does still require a toolchain bump on your side to run the tests. Happy to switch to the revert if you'd rather not move the floor for a dev-dependency that buys us nothing — it's a one-line change either way.

khj809: I'm also fine with upgrading the toolchain. Let's bump it.

nuri-yoo: Sounds good — keeping `test-with` 0.17 and the declared `rust-version = "1.95.0"` as they are in 6ed0447, then. No further change needed for this.

For what it's worth on the CI side: there is no `rust-toolchain.toml` in the repo, and every workflow uses `actions-rust-lang/setup-rust-toolchain@v1` with no `toolchain:` input, so CI resolves to stable (currently 1.97.1) and is unaffected by the floor. The declaration only changes what cargo reports locally when someone builds on an older toolchain — it now names the requirement instead of failing inside `test-with`'s expansion.

khj809 (approved):
LGTM

PR #442, opened 2026-07-29
fix(agent): seed the system message into a resumed history

grf53:
## Description

The instruction + skills table was synthesised only when `state.history` was empty, so a host that rebuilds the `Agent` each turn from a spec plus a persisted conversation — which stores only the user/assistant turns — ran with no system message from the second turn onward. Seed it whenever the history does not lead with a `Role::System` message instead, inserted at the front.

Persisting the system message is not the alternative: its text is derived from the spec, so a stored copy goes stale whenever the instruction or skill set changes.

## Behaviour change

```rust
AgentBuilder::new(model)
    .instruction("...")                 // spec carries an instruction
    .history([user_msg, assistant_msg]) // non-empty, no system message
    .build()
```

- Before: seeding was skipped, and the agent ran with no system message.
- After: a system message is inserted at index 0.

That is the only case affected. A host doing this deliberately — inlining its instruction into the first user message, or keeping `spec.instruction` as display metadata it never sends — now gets a system message it did not ask for. Hosts that round-trip `get_history()` in full are unaffected, since the returned history includes the seeded system message and the condition never fires. A history leading with its own system message is still taken as-is, and its text still wins over `spec.instruction`.

## Changes

- `src/agent/rt.rs` — `Agent::try_with_provider_and_state`: condition `state.history.is_empty()` → `!state.history.iter().any(|m| m.role == Role::System)`, and `push` → `insert(0, …)`. Front, because that is where every schema expects a system message — Anthropic, Gemini and OpenAI extract the first one, and ChatCompletion sends them in place.
- The predicate is `any`, one step wider than the documented leading-position contract: a stray system message past index 0 still suppresses seeding, where a `history[0]` check would seed a second one and let the first-match lookup drop the caller's.
- `src/agent/builder.rs` — two regression tests.
- Doc comments on the call sites updated to the new contract.

Sub-agents are unaffected — `subagent.rs` builds each invocation with a fresh `AgentState::new()`.

## Tests

- Instruction seeded into a history lacking a system message — `[User, Assistant]` yields length 3, system message at index 0, original turns preserved.
- Existing system message neither replaced nor duplicated — `[System, User]` stays at length 2 and keeps its own text despite a differing `spec.instruction`.
- Suite: `cargo test --lib agent::` — 39 passed, 0 failed, 0 ignored (7 of them need API keys from `.env`).

### DIFF
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -136,8 +136,8 @@ impl AgentBuilder {
     }
 
     /// Seed the agent's [`AgentState::history`] (e.g. for resuming a prior session).
-    /// When non-empty, this overrides the system message that the spec's instruction
-    /// would otherwise produce.
+    /// A system message here overrides the one the spec's instruction would produce;
+    /// otherwise the instruction is still seeded, at the front of this history.
     pub fn history(mut self, history: impl IntoIterator<Item = Message>) -> Self {
         self.history = history.into_iter().collect();
         self
@@ -495,5 +495,47 @@ mod tests {
             .build()
             .unwrap();
         assert!(agent.get_context_manager().is_some());
+    }
+
+    fn msg(role: Role, text: &str) -> Message {
+        Message::new(role).with_contents([crate::message::Part::text(text)])
+    }
+
+    #[tokio::test]
+    async fn test_instruction_seeded_into_history_without_system_message() {
+        ensure_dummy_provider();
+        let agent = AgentBuilder::new(TEST_MODEL)
+            .agent_provider(TEST_PROVIDER_NAME)
+            .instruction("You are a test agent.")
+            .history([msg(Role::User, "hello"), msg(Role::Assistant, "hi")])
+            .build()
+            .unwrap();
+
+        let history = agent.get_history();
+        assert_eq!(history.len(), 3);
+        assert_eq!(
+            system_text(&agent).as_deref(),
+            Some("You are a test agent.")
+        );
+        assert_eq!(history[1].role, Role::User);
+        assert_eq!(history[2].role, Role::Assistant);
+    }
+
+    #[tokio::test]
+    async fn test_existing_system_message_is_not_replaced() {
+        ensure_dummy_provider();
+        let agent = AgentBuilder::new(TEST_MODEL)
+            .agent_provider(TEST_PROVIDER_NAME)
+            .instruction("spec instruction")
+            .history([
+                msg(Role::System, "stored instruction"),
+                msg(Role::User, "hello"),
+            ])
+            .build()
+            .unwrap();
+
+        let history = agent.get_history();
+        assert_eq!(history.len(), 2);
+        assert_eq!(system_text(&agent).as_deref(), Some("stored instruction"));
     }
 }
--- a/src/agent/builder.rs
+++ b/src/agent/builder.rs
@@ -136,8 +136,8 @@ impl AgentBuilder {
     }
 
     /// Seed the agent's [`AgentState::history`] (e.g. for resuming a prior session).
-    /// A system message here overrides the one the spec's instruction would produce;
-    /// otherwise the instruction is still seeded, at the front of this history.
+    /// A leading system message here overrides the one the spec's instruction would
+    /// produce; otherwise the instruction is still seeded, at the front of this history.
     pub fn history(mut self, history: impl IntoIterator<Item = Message>) -> Self {
         self.history = history.into_iter().collect();
         self
--- a/src/agent/rt.rs
+++ b/src/agent/rt.rs
@@ -106,10 +106,10 @@ impl Agent {
     /// skill paths are **not** rewritten, so sub-agent skills are portable
     /// across parents.
     ///
-    /// If `state.history` is empty, a system message synthesised from
-    /// `spec.instruction` and the declared skills table is seeded as the first
-    /// entry.  When `state.history` is non-empty (e.g. resuming a prior
-    /// session) it is taken as-is and no system message is synthesised.
+    /// Unless `state.history` already carries a [`Role::System`] message, one
+    /// synthesised from `spec.instruction` and the declared skills table is
+    /// inserted at the front; a history that carries one is taken as-is, so the
+    /// caller's own system message wins.
     pub fn try_with_provider_and_state(
         spec: AgentSpec,
         provider: impl AsRef<str>,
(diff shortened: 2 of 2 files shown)


src/agent/rt.rs
```diff
@@ -167,8 +167,9 @@ impl Agent {
         }
 
         // Build the system message: instruction + (optionally) skills table.
-        // Only seeded when the caller didn't supply a pre-existing history.
-        if state.history.is_empty() {
+        // `any`, not `history[0]`: a system message anywhere means the caller
+        // manages it, and the providers resolve theirs by first match.
```
nuri-yoo: Comment wording only. The code is right for every schema; the stated reason misses one.

`resolve theirs by first match` holds for three of the four API schemas. `anthropic.rs:177`, `gemini.rs:240` and `openai.rs:173` each do `.find(|m| m.role == Role::System)` and filter every `System` out of the message array. `ChatCompletionMarshal` does neither: `chat_completion.rs:61` marshals each message inline with `"role": item.role.to_string()`, and the request marshal at `chat_completion.rs:185` calls it over the whole slice with no extraction or dedup. grok, deepseek, kimi and any OpenAI-compatible endpoint reached through `chat_completion(url)` therefore receive every `System` message, in place.

Front-insertion is still correct there, since leading position is the convention that schema expects, so the code needs no change. If you want the comment to hold up against `chat_completion.rs`, one option is to name the split: Anthropic/Gemini/OpenAI extract the first system message, ChatCompletion sends them inline where leading position is the convention. Worth it?

src/agent/rt.rs
```diff
@@ -178,9 +179,12 @@ impl Agent {
                 (None, None) => None,
             };
             if let Some(text) = system_text {
-                state
-                    .history
-                    .push(Message::new(Role::System).with_contents([Part::text(text)]));
+                // Front: the providers read the first system message, and
+                // `ContextManager` only pins `history[0]` against eviction.
```
nuri-yoo: Comment wording again.

`only pins history[0] against eviction` reads as a guarantee in force today, but nothing is evicted yet. `truncate_history` replaces the contents of `Role::Tool` messages with a placeholder and never removes a message, so a `System` message is safe at any index right now and `start_idx` has no effect on it. `context.rs` leaves group-level dropping to a future iteration, so the invariant is real as intent; only the present tense is off.

src/agent/builder.rs
```diff
@@ -496,4 +496,46 @@ mod tests {
...
+            .history([
+                msg(Role::System, "stored instruction"),
+                msg(Role::User, "hello"),
+            ])
+            .build()
+            .unwrap();
+
+        let history = agent.get_history();
+        assert_eq!(history.len(), 2);
+        assert_eq!(system_text(&agent).as_deref(), Some("stored instruction"));
```
nuri-yoo, line 539: The doc comment this PR adds to `try_with_provider_and_state` says "Unless `state.history` already carries a `Role::System` message", and `carries` means anywhere. Both new tests put it at index 0, where a `history[0]` check and an `any()` check agree, so the anywhere half of the contract ships untested.

It does behave as documented. With `instruction("spec instruction")` and `history([User("hello"), System("mid-conversation nudge")])` nothing is seeded, the history stays at length 2, and the providers send `system = "mid-conversation nudge"` with `spec.instruction` dropped. One test covers it, and it passes on de88ccb as written:

```rust
#[tokio::test]
async fn test_system_message_not_at_front_still_suppresses_seeding() {
    ensure_dummy_provider();
    let agent = AgentBuilder::new(TEST_MODEL)
        .agent_provider(TEST_PROVIDER_NAME)
        .instruction("spec instruction")
        .history([
            msg(Role::User, "hello"),
            msg(Role::System, "mid-conversation nudge"),
        ])
        .build()
        .unwrap();

    let history = agent.get_history();
    assert_eq!(history.len(), 2);
    assert_eq!(history[0].role, Role::User);
    let system = history
        .iter()
        .find(|m| m.role == Role::System)
        .and_then(|m| m.contents.iter().find_map(|p| p.as_text()));
    assert_eq!(system, Some("mid-conversation nudge"));
}
```

`find` rather than `system_text()`: the helper reads `history.first()` and returns `None` for this shape.

One measurement on how far the two new tests reach: swapping the condition to `!state.history.first().is_some_and(|m| m.role == Role::System)` leaves both of them green and fails only the test above, with `left: 3, right: 2`.

nuri-yoo (commented):
The fix is correct and front-insertion is the right call. What I checked, so it is reproducible rather than asserted. The three inline notes are comment wording plus one test suggestion; none is a behaviour problem.

## What I checked

* `context.rs:53` confirms the pinning claim: `let start_idx = if history[0].role == Role::System { 1 } else { 0 }`.
* `anthropic.rs:177`, `gemini.rs:240` and `openai.rs:173` each extract with `.find(|m| m.role == Role::System)` and filter every `System` out of the message array, so front-insertion is what gets the seeded message into the request.
* Front-insertion leaves the index-based thinking-inclusion logic intact, the interaction I expected to break. `anthropic.rs::marshal_messages` and `openai.rs::marshal_messages` compare `enumerate()`'s `i` against a `last_user_index` from `rposition` over that same full slice, so inserting at 0 shifts both by 1 and `i > last_user_index` is invariant. `gemini.rs` has no index logic.
* `subagent.rs:106` adds nothing to `AgentState::new()` before the `try_with_provider_and_state` call, so sub-agents are unaffected as stated.
* `any()` is the safer predicate. Under a `history[0]` check, a caller whose system message sits at a non-zero index gets a second one at the front, and since those three schemas take the first match and drop the rest, the caller's own message disappears from the request.
* One case affected is accurate: for an empty history `push` and `insert(0, …)` are identical, and a history already carrying a `System` message is skipped under both conditions.

## Test check

Reverting only the `rt.rs` condition and `insert`, `test_instruction_seeded_into_history_without_system_message` fails with `left: 2, right: 3`, so it is real regression coverage. `test_existing_system_message_is_not_replaced` passes under both conditions, so it guards rather than covers; that is what the `builder.rs` note is about.

On de88ccb, `cargo test --lib agent::` gives 39 passed / 0 failed / 0 ignored, and `--features sandbox` the same 39. `cargo fmt --check` shows no diff in either changed file, and `cargo clippy --lib --all-targets` no warning inside the changed regions.

src/agent/rt.rs
grf53: I don't think it's necessary to write out the logic in detail. I replaced it with an explanation of the problem.

src/agent/rt.rs
grf53: revised that.

src/agent/builder.rs
grf53: As discussed, we assume that system messages are not included in the middle of the history and intend to adopt the principle of managing them as instructions (without removing the processing for the leading system message in the history).

So your suggested test won't be added here, and I adjusted some comments to be align with that.

nuri-yoo: Agreed. LGTM

nuri-yoo (approved):

grf53:
I postponed the merge to allow more people to become aware of and approve of this content, and now that enough time has passed, I will proceed.

PR #443, opened 2026-07-30
fix(sandbox): gate host-side web_fetch egress and scope guest host reach per port

nuri-yoo:
Hardening for the boundary between the host process and the sandbox guest, from an internal review of the network posture. Two independent parts, one commit each.

## `web_fetch` egress is gated to public addresses

`web_fetch` is registered as a pure tool, so it receives no console and the request runs in the host process rather than inside the sandbox VM. The guest network policy therefore never applies to it, and the only validation was URL length, scheme, and the presence of a host string. That left the tool able to reach addresses the sandbox exists to keep an agent away from, and the tool is callable both by a model and, indirectly, by a document a model reads.

A new `net_guard` module decides whether a destination is on the public internet, and the `web_fetch` client is wired through it in three places:

- **IP-literal hosts** are checked against the IANA special-purpose ranges before a connection is opened. The check runs on the parsed host rather than the raw input, so `url`'s normalization of the decimal, octal, hex, and shorthand spellings of an address is what makes those cases safe. There is a test asserting exactly that, since `std`'s `IpAddr` parser rejects all of those forms on its own.
- **Named hosts** go through a custom DNS resolver that drops every non-public answer. Filtering inside the resolver rather than checking the addresses up front is the point: the connector reaches only addresses the filter returned, so there is no gap between a check and the connect for a second answer to land in. A split answer keeps its public addresses and loses the rest.
- **Redirects** re-check scheme and host on every hop. A public start URL can redirect inward, and a `Location` holding an IP literal never reaches the resolver at all, so the per-hop check is what covers it. The hop limit behavior is unchanged.

Addresses that carry an IPv4 address inside an IPv6 one are judged by the address they carry, because that is where the connection lands: `::ffff:127.0.0.1` and the NAT64 form both reach loopback. Teredo and 6to4 prefixes are refused for the same reason. RFC 8215's local-use NAT64 range, `64:ff9b:1::/48`, is refused wholesale rather than unwrapped, because the prefix may be any length up to `/96` there, so the embedded address does not sit at a fixed offset.

Blocked targets are logged at `warn`, which also gives a signal when something is probing for a way around the gate.

All three paths report a refusal the same way, which took a fix of its own after @selloriwoo caught the resolver path not doing so. An IP literal refused before the connection returned `blocked: 127.0.0.1 is not a public address`, but a name refused inside the resolver returned `request failed: ... client error (Connect)` — the block held, the local server in the test recorded no hit, but the refusal sat three levels down a source chain whose outer `Display` stops one level in, so `download`'s `format!("request failed: {e}")` never reached it. That is worth fixing rather than documenting, because the two messages carry opposite instructions: a connect failure reads as a target that happens to be down and invites a retry, a refusal reads as terminal, and the path that will never succeed no matter how many times it is tried was the one reporting as retryable. The guard now raises a `Blocked` error type and `blocked_reason` walks the chain for it — a type rather than a message prefix, since the refusal has to be recognized after the connector has wrapped it twice and matching on text would break the moment someone rewords it. The redirect path was not opaque, `wreq`'s `Display` does carry the hop error's message, but it arrived behind a `request failed: error following redirect for uri (...)` wrapper and now reads like the other two.

One case stays uncovered by design and is documented on the resolver: with an HTTP proxy configured, the connector resolves the proxy's host and the proxy resolves the target, so the target's addresses never pass through the filter. That configuration comes from the host environment rather than from anything a model can set, and the IP-literal check still applies. Worth being precise about how reachable that is rather than leaving it as a footnote: wreq enables system proxies by default (`auto_sys_proxy` is `true`), and the client here does not call `.no_proxy()`, so an `HTTP_PROXY` or `ALL_PROXY` in the environment is enough to put every named host on that path. Surfacing it — a warning when a proxy is configured, or making one opt-in — is a follow-up rather than part of this change.

## Guest host reach is scoped per port, and the default posture is the narrow one

Every `SandboxNetwork` variant built its host rule with `Rule::allow_egress(Destination::Group(DestinationGroup::Host))`. That constructor leaves the rule's port set empty, and an empty port set matches every port. The engine behaves exactly as documented here; the call site expressed more than was intended. The practical effect was that `HostOnly`, the posture meant to be least-privilege, granted the guest every TCP port on the host, so any host service listening on `0.0.0.0` was reachable from inside the VM. Two related defaults compounded it: the default posture was `Public`, so a caller that never mentioned the network got arbitrary outbound reach, and `HostOnly` and `Public` both set `default_ingress: Allow`, flipping microsandbox's own fail-closed default open.

This adds `NetworkPosture`, which pairs the outside-reach variant with the host TCP ports the guest may actually open:

- Host reach is gateway DNS plus the listed ports, nothing else. `Rule::allow_dns()` is what keeps name resolution working, including the lookup for `host.microsandbox.internal`; without it a deny-by-default policy refuses every query.
- The default is `HostOnly` with no host ports. Wider reach is opt-in.
- `default_ingress` is `Deny` again. It only governs connections inbound to a published guest port, and nothing here publishes one, so this is fail-closed for a capability that does not exist yet rather than a restriction on anything in use.
- `domain_suffixes` lets a session that has to install packages name its registries instead of taking `Public` and the whole internet with it. An unparseable suffix is dropped and logged, so a typo costs a failed install rather than an unintended grant. One limit is worth recording, since it belongs to microsandbox's evaluator rather than to this change: over TLS the match is a conjunction, the SNI has to match the suffix *and* the address has to have been resolved from a name that matches, so a shared address on its own admits nothing. Without SNI — plain HTTP, which is how `apt` fetches — the match is on the address alone against any name cached for it, so an allowlisted domain and an unrelated one behind the same CDN address are not told apart there.
- `Full` keeps `allow_all()` as the deliberate escape hatch.

The split between reach and host ports is deliberate: a posture on its own now says nothing about which host services are exposed, so widening outside reach cannot quietly widen host reach too.

### Call sites

`network()` and `try_from_archive_with_network()` take `impl Into<NetworkPosture>`, so a call site passing a bare `SandboxNetwork` keeps compiling and gets the narrower policy. A caller that needs a host service reachable from the guest (a forward server, for instance) asks for it by port:

```rust
SandboxBuilder::new()
    .network(SandboxNetwork::HostOnly.with_host_ports([forward_port]))
```

`try_from_archive_with_network` gains the same coverage, which matters because an archive is only a filesystem snapshot and carries no policy; the host ports would otherwise be lost across a save and restore.

The same `impl Into<NetworkPosture>` that keeps call sites compiling is what makes this a quiet change for anyone using the crate from outside: code that never mentioned the network, or that restores through `try_from_archive`, still builds and now gets `HostOnly` where it used to get `Public`. Nothing warns, because there is nothing to warn about at the type level. There is no CHANGELOG in this repo, so this is the note to carry into the release: a caller that did want outbound reach has to say so now, either `SandboxNetwork::Public` or a `with_domain_suffixes` list. `python_repl` is the first case of that and is documented in its own right.

The one caller inside this repo that the narrowing reaches is the skill benchmark test, which registers `python_repl` against a sandbox it never named a network for, so it now asks for the same suffixes `SETUP_CMD` documents. It passes either way today — twice over against a real VM and the real API — because the skill steers the model to `python -m timeit` in the shell and the model follows, but that is the model's choice rather than a property of the test, and the failure the other choice would hit is a quiet one: the run is `#[ignore]`d, needs a key, and a model that tried `python_repl` and lost it would likely fall back to the shell and still produce the expected table. Measuring the guest is what settled it: `python:3.12-slim` ships python3 but no uv, and under the new default the uv release and pypi.org both come back `URLError: [Errno 111] Connection refused`. The `runenv` module doc gains the same point for anyone starting from the sandbox example there, which showed a bare `build()` and did not mention the network at all.

`fork` builds from a snapshot the same way and had the same hole, so the policy is carried on `Sandbox` and re-applied there too. Without it a fork fell through to microsandbox's default, which is `Public` egress and `default_ingress: Allow` — both halves of this narrowing, undone for anyone who forks. It was quiet before now only because the old default was `Public` too, so a fork reproduced its parent by coincidence.

## Verification

```
cargo clippy --all-targets                      # clean
cargo clippy --all-targets --features sandbox   # clean
cargo test --features sandbox --lib
    test result: ok. 353 passed; 0 failed; 18 ignored
```

The 353 includes the sandbox tests that boot real VMs, so `HostOnly` narrowed to DNS plus explicit ports is confirmed to still produce a working guest network rather than a dead one, and `Public` still reaches the internet. One of the 18 ignored is new and is described below; it needs `--ignored` to run.

New coverage:

- `net_guard`: the range predicates for both families, the NAT64 unwrap in both directions and the local-use range that is refused instead, the resolver's filter on a split answer and on an all-internal one, and the redirect-target rule. Three ranges the predicate described as covered but let through are refused now — 100::/64, which discards what is sent to it, and 3fff::/20 and 5f00::/16, reserved for documentation and for SRv6 segment identifiers — each with the address just outside it in the allowed list, so a mask one bit too wide fails here rather than quietly refusing traffic that should have gone out. 2001:20::/28 and 2001:30::/28 needed no rule of their own, since both sit inside the 2001::/23 entry, and a case pinning that is in the refused list. `blocked_reason` gets two of its own: it finds a refusal under two wrappers that drop their source when formatted, which holds the contract even if `wreq` restructures its errors, and it returns `None` for an ordinary io error, since reporting a target that is genuinely down as a policy refusal would be the same confusion in reverse.
- `web_fetch`: a loopback target is refused with a real local server recording no hit, and a redirect into loopback fails with the first hop served and the second never requested. The redirect test asserts the first hop was served, because otherwise a failure to reach the server at all would look identical to a refused second hop. A third fetches a name that resolves to loopback, covering the resolver path end to end and asserting it reads like the literal case; the redirect test asserted only that some error occurred, which the old generic message satisfied, and now asserts the refusal too. Both of those fail without the recovery in `download`.
- `runenv::sandbox`: a posture that omits a field deserializes to the narrow default rather than failing, checked to fail without the `#[serde(default)]` that makes it so. The postures are also put through microsandbox's own evaluator for host ports, gateway DNS, unmatched ingress, and domain suffixes, so a rule wider than intended shows up as a failing assertion rather than as an open port nobody looked at. One further test boots a VM with two host listeners, one port granted and one not, and confirms the guest reaches exactly one — the port set has to be enforced by the runtime, not merely recorded in the spec. That one was checked against a control run granting both ports, which fails as it should.
- `runenv::sandbox`, forks: a fork of a `HostOnly` parent must not reach `1.1.1.1:443`, and a fork of a `Public` parent still must. The second is there because the first would also pass on a fork whose network was simply broken. Checked that the first fails without the re-apply, where the child does reach the internet.
- `python_repl`: the domain allowlist its setup needs is documented on `SETUP_CMD` and backed by a test that drives the real tool against a real VM under exactly that posture, installs a wheel end to end, and confirms a domain outside the list still fails to connect. Checked both ways: on bare `HostOnly` it fails at `apt`, so the allowlist is what carries it rather than an accidentally open policy, and the control asserts `URLError` rather than only a non-zero exit, since a broken probe script would also exit non-zero. `#[ignore]`d like the `web_search` tests because it installs from three services this repo does not control.

## Notes

`web_search` is also a pure tool, but its engine URL is fixed, so it is not routed through the guard in this change. The `net_guard` module is available if that changes.

Left for a separate change in `agent-k`, since the code lives there: the host-side forward server's bind address and token comparison, and the document parser's lack of a timeout or resource cap. This branch is the piece those depend on, because the forward server now has to name its port.

The consuming side is written against this API and is up for review as brekkylab/agent-k#221, so this is ready rather than draft. Merging it first is the intended order: #221 pins this branch's head and has to be repinned to the merged commit afterwards.

### DIFF
--- a/src/runenv/sandbox.rs
+++ b/src/runenv/sandbox.rs
@@ -154,46 +154,149 @@ fn wire_network_policy(policy: NetworkPolicy) -> microsandbox_types::NetworkPoli
     serde_json::from_value(value).expect("engine and wire NetworkPolicy share one JSON schema")
 }
 
-/// Guest network reachability. The sandbox host
-/// (`host.microsandbox.internal`, e.g. an ailoy VFS forward server) is
-/// reachable in **every** variant; they differ only in outside reach. There is
-/// no fully-offline variant. All map to a [`NetworkPolicy`] with
-/// `default_ingress: Allow` (published-port behavior).
+/// How far outside itself the guest can reach. There is no fully-offline
+/// variant; every variant permits DNS to the sandbox gateway so the guest can
+/// resolve names at all.
+///
+/// Reaching a service on the host is a separate grant, expressed per port
+/// through [`NetworkPosture::host_ports`]. That split is deliberate: a posture
+/// on its own says nothing about which host services are exposed, so widening
+/// the outside reach cannot silently widen host reach too.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum SandboxNetwork {
-    /// Host only — no public internet, LAN, loopback, or metadata.
-    /// Least-privilege for VFS-in-sandbox.
-    HostOnly,
-    /// Host + public internet, but not private LAN, loopback, link-local,
-    /// cloud-metadata, or multicast. The default.
+    /// Nothing but the host ports explicitly granted, plus gateway DNS. The
+    /// default, and the least-privilege posture for VFS-in-sandbox.
     #[default]
+    HostOnly,
+    /// The public internet, but not private LAN, loopback, link-local,
+    /// cloud-metadata, or multicast.
     Public,
-    /// Unrestricted egress/ingress — `Public` plus LAN, loopback, link-local,
-    /// cloud-metadata, and multicast. Grant deliberately (reopens SSRF).
+    /// Unrestricted egress *and* ingress — `Public` plus LAN, loopback,
+    /// link-local, cloud-metadata, multicast, and every host port regardless of
+    /// [`NetworkPosture::host_ports`]. Grant deliberately; it reopens SSRF.
     Full,
 }
 
 impl SandboxNetwork {
-    /// The microsandbox policy for this variant.
-    fn policy(self) -> NetworkPolicy {
-        use microsandbox_network::policy::{Action, Destination, DestinationGroup, Rule};
-        // `allow_egress(Host)` permits any port to the host — including :53, so
-        // the guest's `host.microsandbox.internal` DNS lookup works.
-        let host = || Rule::allow_egress(Destination::Group(DestinationGroup::Host));
-        let public = || Rule::allow_egress(Destination::Group(DestinationGroup::Public));
-        match self {
-            SandboxNetwork::HostOnly => NetworkPolicy {
-                default_egress: Action::Deny,
-                default_ingress: Action::Allow,
-                rules: vec![host()],
-            },
-            SandboxNetwork::Public => NetworkPolicy {
-                default_egress: Action::Deny,
-                default_ingress: Action::Allow,
-                rules: vec![public(), host()],
-            },
-            SandboxNetwork::Full => NetworkPolicy::allow_all(),
+    /// This posture plus egress to the listed host TCP ports.
+    pub fn with_host_ports(self, ports: impl IntoIterator<Item = u16>) -> NetworkPosture {
+        NetworkPosture::from(self).with_host_ports(ports)
+    }
+
+    /// This posture plus egress to the listed domain suffixes.
+    pub fn with_domain_suffixes(
+        self,
+        suffixes: impl IntoIterator<Item = impl Into<String>>,
+    ) -> NetworkPosture {
+        NetworkPosture::from(self).with_domain_suffixes(suffixes)
+    }
+}
+
+/// A guest network posture: how far out the guest can reach, and which host TCP
+/// ports it may open.
+///
+/// `host_ports` has to be named explicitly because the convenience constructor
+/// for a host rule (`Rule::allow_egress(Group(Host))`) leaves the port set
+/// empty, and an empty port set means every port. A guest that needs one host
+/// service would otherwise be handed the whole host: SSH, a database, a
+/// container daemon, anything bound to `0.0.0.0`.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct NetworkPosture {
+    /// Outside reach.
+    pub network: SandboxNetwork,
+    /// Host TCP ports the guest may connect to. Gateway DNS is always allowed
+    /// and does not need to be listed.
+    #[serde(default)]
+    pub host_ports: Vec<u16>,
+    /// Domain suffixes the guest may reach, matching the apex and any
+    /// subdomain. Lets a session that has to install packages name its
+    /// registries instead of taking [`SandboxNetwork::Public`] and the whole
+    /// internet with it.
+    ///
+    /// Matching works through the gateway resolver's hostname cache, so it
+    /// covers names the guest looked up there — which is every name it can
+    /// resolve under these postures.
+    #[serde(default)]
+    pub domain_suffixes: Vec<String>,
+}
+
+impl From<SandboxNetwork> for NetworkPosture {
+    fn from(network: SandboxNetwork) -> Self {
+        Self {
+            network,
+            host_ports: Vec::new(),
+            domain_suffixes: Vec::new(),
+        }
+    }
+}
+
+impl NetworkPosture {
+    /// Grant egress to these host TCP ports, replacing any already set.
+    pub fn with_host_ports(mut self, ports: impl IntoIterator<Item = u16>) -> Self {
+        self.host_ports = ports.into_iter().collect();
+        self
+    }
+
+    /// Grant egress to these domain suffixes, replacing any already set.
+    pub fn with_domain_suffixes(
+        mut self,
+        suffixes: impl IntoIterator<Item = impl Into<String>>,
+    ) -> Self {
+        self.domain_suffixes = suffixes.into_iter().map(Into::into).collect();
+        self
+    }
+
+    /// The microsandbox policy for this posture.
+    fn policy(&self) -> NetworkPolicy {
+        use std::str::FromStr as _;
+
+        use microsandbox_network::policy::{
+            Action, Destination, DestinationGroup, Direction, DomainName, PortRange, Protocol, Rule,
+        };
+
+        if matches!(self.network, SandboxNetwork::Full) {
+            return NetworkPolicy::allow_all();
+        }
+
+        // `Rule::allow_dns()` is narrow: UDP/TCP :53 to the gateway addresses
+        // only, not to arbitrary resolvers the guest might aim at. It has to
+        // come first, because under deny-by-default a policy without it refuses
+        // every DNS query, including the one for `host.microsandbox.internal`.
+        let mut rules = vec![Rule::allow_dns()];
+        if matches!(self.network, SandboxNetwork::Public) {
+            rules.push(Rule::allow_egress(Destination::Group(
+                DestinationGroup::Public,
+            )));
+        }
+        rules.extend(self.host_ports.iter().map(|&port| Rule {
+            direction: Direction::Egress,
+            destination: Destination::Group(DestinationGroup::Host),
+            protocols: vec![Protocol::Tcp],
+            ports: vec![PortRange::single(port)],
+            action: Action::Allow,
+        }));
+        rules.extend(self.domain_suffixes.iter().filter_map(|suffix| {
+            match DomainName::from_str(suffix) {
+                Ok(name) => Some(Rule::allow_egress(Destination::DomainSuffix(name))),
+                // Dropping the entry means the domain simply is not reachable,
+                // so a typo costs a failed install rather than an unintended
+                // grant. Logged loudly because the failure is otherwise opaque.
+                Err(e) => {
+                    log::warn!("sandbox network: ignoring invalid domain suffix '{suffix}': {e}");
+                    None
+                }
+            }
+        }));
+
+        NetworkPolicy {
+            default_egress: Action::Deny,
+            // microsandbox's own serde default. It only governs connections
+            // inbound to a published guest port, and nothing here publishes
+            // one, so this is fail-closed for a capability that does not exist
+            // yet rather than a restriction on anything in use.
+            default_ingress: Action::Deny,
+            rules,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/runenv/sandbox.rs
```diff
@@ -154,46 +154,149 @@ fn wire_network_policy(policy: NetworkPolicy) -> microsandbox_types::NetworkPoli
...
+///
+/// `host_ports` has to be named explicitly because the convenience constructor
+/// for a host rule (`Rule::allow_egress(Group(Host))`) leaves the port set
+/// empty, and an empty port set means every port. A guest that needs one host
+/// service would otherwise be handed the whole host: SSH, a database, a
+/// container daemon, anything bound to `0.0.0.0`.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct NetworkPosture {
+    /// Outside reach.
+    pub network: SandboxNetwork,
```
khj809: Suggest to put `#[serde(default)]` here too, in order to deserialize `{"host_ports":[8080]}` or `{}`.

src/runenv/sandbox.rs
```diff
@@ -154,46 +154,149 @@ fn wire_network_policy(policy: NetworkPolicy) -> microsandbox_types::NetworkPoli
...
 pub enum SandboxNetwork {
-    /// Host only — no public internet, LAN, loopback, or metadata.
-    /// Least-privilege for VFS-in-sandbox.
-    HostOnly,
-    /// Host + public internet, but not private LAN, loopback, link-local,
-    /// cloud-metadata, or multicast. The default.
+    /// Nothing but the host ports explicitly granted, plus gateway DNS. The
+    /// default, and the least-privilege posture for VFS-in-sandbox.
     #[default]
+    HostOnly,
```
khj809, line 178: Narrowing the default from `Public` to `HostOnly` would cause `python_repl` to be failed on downloading python runtime (uv). At least we should notify the minimum required network policy in `python_repl`'s comments.

src/runenv/sandbox.rs
nuri-yoo: Done in 809ff3f. `{}` and `{"host_ports":[8080]}` both failed with `missing field network`, and the other two fields already carried it.

Worth noting the direction: `SandboxNetwork` defaults to `HostOnly`, so an omitted field now resolves to the narrow posture instead of erroring. Covered by a test, which I checked fails without the attribute.

src/runenv/sandbox.rs
nuri-yoo: Confirmed. Under the new default, setup dies at its first step: `Unable to connect to archive.ubuntu.com:80`, then `E: Unable to locate package python3`.

Measured the minimum rather than guessing at it in a comment:

| posture | setup + `uv pip install` |
| --- | --- |
| `HostOnly` (new default) | fails, apt reaches no mirror |
| `Public` | works |
| `HostOnly` + the 5 suffixes below | works |

e58d451 documents the list on `SETUP_CMD` with a reason per entry: `ubuntu.com`, `github.com`, `githubusercontent.com` (where the release redirects, checked by dropping it), `pypi.org`, `pythonhosted.org`. Backed by a test running the real tool against a real VM under that posture, which also checks an unlisted domain still fails. 06cb509 marks it `#[ignore]` like the `web_search` tests, with the run command on the attribute.

src/tool/impl/builtins/net_guard.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+    match host {
+        Some(h) => check_host(h),
+        None => Err("blocked: redirect target has no host".to_string()),
+    }
+}
+
+/// Keep only the globally routable addresses in `addrs`. Returns `Err` when
+/// nothing survives, so the caller reports a blocked host rather than an
+/// opaque connection failure.
```
selloriwoo: This half of the claim doesn't hold at the caller. The `log::warn!` below fires
as described, so the operator signal is intact — but the caller does get the
opaque connection failure:

```
fetch_one(..., "http://localhost:<port>/admin", ...)
→ "error": "request failed: error sending request for uri (...): client error (Connect)"
```

(The block itself works — the local server records zero hits.)

The reason is that `Display` on the outer error stops one level in, while the
refusal sits three deep:

```
DISPLAY : error sending request for uri (...): client error (Connect)
SOURCE 1: client error (Connect)
SOURCE 2: dns error
SOURCE 3: blocked: 'localhost' resolves only to non-public addresses
```

so `download`'s `map_err(|e| format!("request failed: {e}"))` never sees it.

Worth fixing rather than just re-wording, I think, because it splits the two
guard paths: an IP literal is refused in `fetch_one` and reports
`"error": "blocked: ..."`, while a name refused by the resolver reports a
generic connect error. A model reads the first as terminal and the second as
transient, and retries — which is the opposite of what each one deserves.

The refusal is recoverable from the chain, and keying on the prefix the guard
already writes makes both paths report identically:

```rust
// net_guard.rs
/// Recover a guard refusal from a client error's source chain. The connector
/// buries a resolver error deep enough that `Display` on the outer error drops
/// it, so a name refused by [`PublicOnlyResolver`] would otherwise be
/// indistinguishable from the target being down.
pub fn blocked_reason(err: &(dyn std::error::Error + 'static)) -> Option<String> {
    let mut src = Some(err);
    while let Some(e) = src {
        let msg = e.to_string();
        if msg.starts_with("blocked: ") {
            return Some(msg);
        }
        src = e.source();
    }
    None
}

// web_fetch.rs, in `download`
.map_err(|e| net_guard::blocked_reason(&e).unwrap_or_else(|| format!("request failed: {e}")))?
```

Checked against this branch: the same `localhost` fetch then returns
`"error": "blocked: 'localhost' resolves only to non-public addresses"`, matching
the IP-literal path exactly.

If you'd rather keep the change out of scope, the alternative is narrowing the
doc — say the warn log is the signal and the caller sees a connect failure —
since as written it describes behaviour the caller doesn't get.

nuri-yoo: Confirmed, reproduced, and fixed in c8d2d1c rather than narrowing the doc, for the reason you gave: the two messages carry opposite instructions, and the path that will never succeed was the one reporting as retryable.

One deviation. I used a `Blocked` error type and a downcast instead of the `starts_with("blocked: ")` match; the chain walk is otherwise yours. A prefix couples the guard's wording to its control flow, so rewording a message would quietly turn the refusal back into a connect error. Checked the downcast survives the connector's boxing before committing to it. `check_host`, `check_redirect_target` and `filter_public` all return `Blocked` now.

A third path worth recording: the redirect refusal was not opaque, since wreq's `Display` does carry the hop error's message, but it arrived behind a `request failed: error following redirect for uri (...)` wrapper. It reports like the other two now. That also exposed a weak assertion of mine, since `fetch_one_refuses_redirect_into_loopback` checked only `error.is_some()`, which the old generic message satisfied.

Three tests: `blocked_reason` finding a refusal under two wrappers that drop their source when formatted, `blocked_reason` returning `None` for an ordinary io error, and a `fetch_one` against a name resolving to loopback. Both end-to-end assertions fail without the change. 353 passed, clippy clean on both feature sets, body updated.

selloriwoo (approved):

khj809 (approved):
LGTM

PR #446, opened 2026-08-24
[Example] Company analysis

grf53:
## Description

> Details in [`examples/company_analysis/README.md`](https://github.com/brekkylab/ailoy/pull/446/changes#diff-5fe93a5f7820daeba2317fa63c1630e43c469064f9c40e97eb686b9765a5aef6);
> the decisions and the measurements behind them are in `examples/company_analysis/docs/design.html`.
> What follows is the short version.

Adds `examples/company_analysis` — an agent that answers a question about a company
against two public registries and cites the path behind every claim.

The point of the example is *how* the registries are reached. GLEIF and SEC EDGAR are
each a `cortex::fs::FileSystem` mounted over FUSE, so the agent's only interface is the
built-in file tools — no bespoke search tool, no client library. A `read` is an API call
and a listing is what the registry says at that moment; nothing is copied or stored.

A path is a query:

```
gleif/search/entity.legalAddress.country/KR/entity.status/ACTIVE/pages/
    filter[entity.legalAddress.country]=KR&filter[entity.status]=ACTIVE
```

Three rules shaped the tree:

- **Files cost, directories don't.** A directory read asks for every entry's size, and a
  document's size is unknowable without fetching it, so every fetchable document is
  wrapped in a directory of its own. Listing them flat cost 742 requests in one run.
- **Narrowing is free.** The filter fields are known locally, so descending a query
  sends nothing; the first request happens when `pages` is opened. Pinned by an offline
  test that walks a four-deep query and asserts the request count is zero.
- **Where a listing cannot be complete, it says so.** `by-lei` has no index to build one
  from, and `by-cik` would be a million zero-padded numbers nobody can read a company out
  of. Answering empty would read as "nothing here", so each holds a `_README.md` saying
  how to address one and where an identifier comes from.

## Verifying

```sh
cargo test -p company_analysis                                    # offline, 37 tests
SEC_USER_AGENT='...' cargo test -p company_analysis -- --ignored --test-threads=1
```

> [!NOTE]
> `--test-threads=1` is a workaround, not a property of the example: two mounts whose
> lifetimes overlap deadlock when they unmount. **That is a cortex-side issue and a fix is
> [in progress](https://github.com/brekkylab/cortex/tree/fuse-teardown)**; the flag can come off once it lands.

A live run (`--preset cross-registry --company NVIDIA`) resolves NVIDIA in both
registries in 32 tool calls and 8 requests, with every request attributed to the call
that caused it.

### DIFF
--- a/examples/company_analysis/src/apifs.rs
+++ b/examples/company_analysis/src/apifs.rs
@@ -147,13 +147,31 @@ impl Fetcher {
 }
 
 /// Percent-decoding for a path segment, so a value containing `/` can be named.
+/// The byte two hex digits spell, or nothing if they are not both hex.
+///
+/// Not [`u8::from_str_radix`], which takes a sign: it reads `+1` as 1, so `%+1` would
+/// decode where it should stay as written.
+fn hex_byte(hi: u8, lo: u8) -> Option<u8> {
+    let digit = |b: u8| match b {
+        b'0'..=b'9' => Some(b - b'0'),
+        b'a'..=b'f' => Some(b - b'a' + 10),
+        b'A'..=b'F' => Some(b - b'A' + 10),
+        _ => None,
+    };
+    Some(digit(hi)? * 16 + digit(lo)?)
+}
+
 pub fn percent_decode(s: &str) -> String {
     let b = s.as_bytes();
     let mut out = Vec::with_capacity(b.len());
     let mut i = 0;
     while i < b.len() {
+        // Read the two digits as bytes. Slicing the `&str` here would panic on
+        // `%` followed by a multi-byte character — `%가` cuts a char boundary — and this
+        // runs inside an `extern "C"` callback with nothing to catch it, so the panic
+        // aborts and the mount outlives the process that made it.
         if b[i] == b'%' && i + 2 < b.len() {
-            if let Ok(v) = u8::from_str_radix(&s[i + 1..i + 3], 16) {
+            if let Some(v) = hex_byte(b[i + 1], b[i + 2]) {
                 out.push(v);
                 i += 3;
                 continue;
@@ -204,5 +222,21 @@ mod tests {
         for s in ["A/B", "plain", "a b", "100%"] {
             assert_eq!(percent_decode(&urlencode(s)), s);
         }
+    }
+
+    /// A `%` that is not an escape must survive whatever follows it.
+    ///
+    /// Path segments are values a caller wrote — a legal name, a query — so a stray `%`
+    /// is ordinary input. Reading the two digits as anything but bytes panics when the
+    /// next character is multi-byte, and a panic here runs in an `extern "C"` callback:
+    /// it aborts, and the mount is still there afterwards.
+    #[test]
+    fn a_stray_percent_is_left_alone() {
+        for s in ["%가", "%a한", "%🙂", "50%가치", "%", "%4", "%zz", "100%!", "%+1", "%+f"] {
+            assert_eq!(percent_decode(s), s, "{s}");
+        }
+        // Real escapes still decode, including one before a multi-byte character.
+        assert_eq!(percent_decode("%41가"), "A가");
+        assert_eq!(percent_decode("%ea%b0%80"), "가");
     }
 }
--- a/examples/company_analysis/src/apifs.rs
+++ b/examples/company_analysis/src/apifs.rs
@@ -1,0 +1,208 @@
+//! The part every API-backed [`FileSystem`](cortex::fs::FileSystem) needs and none of
+//! them differ on: one client, one cache, one count of what was spent.
+//!
+//! What differs between sources is the path grammar and which URL a path means. That
+//! stays in each store. Fetching does not, and four copies of a retry loop would be
+//! four places for the backoff to drift.
+
+use std::{
+    collections::{BTreeMap, HashMap},
+    io,
+    sync::{
+        Arc, Mutex,
+        atomic::{AtomicUsize, Ordering},
+    },
+    time::Duration,
+};
+
+/// Waits before the 1st and 2nd retry of a rate-limited or 5xx request.
+const BACKOFF: [Duration; 2] = [Duration::from_millis(400), Duration::from_millis(1600)];
+
+/// A client, the bodies it has fetched, and the number of requests that took.
+pub struct Fetcher {
+    http: reqwest::Client,
+    /// key → bytes. Also the size source for `stat`.
+    ///
+    /// Kept for the life of the store, which is one run. Two reasons, and neither is
+    /// speed. A `stat` and the reads after it have to agree on a size, and a report
+    /// that cites a path has to still mean the same bytes when someone follows the
+    /// citation — an index that moved under a 40-turn investigation would leave its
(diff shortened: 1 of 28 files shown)


grf53:
## The role of `_README.md`
(It had been decided a while ago, but the explanation was omitted.)

In cases where items are included that are not listed in the path, it includes methods to access each item even if they do not exist in the path, preventing `ls` from producing a simple empty result and mistakenly assuming there are no accessible items below.

This content may also be included in `CATALOG.md`, but it contains more specific details.

## Decision on the proposals raised during discussions
(FYI, existing examples serve not as a rationalization that 'this method is right because it exists,' but as a reason to add that it is better than other alternatives.)

### 1. Not listed is not unreachable

`ls by-lei` names only `_README.md` — millions of entities, no index to enumerate them
from — yet `ls by-lei/549300S4KLFTLO7GSQ80/` opens. The listing returns; it simply
leaves out what it cannot name. That is an ordinary shape:

| Precedent | |
| --- | --- |
| **autofs non-browsable maps** — `/net/<host>` | `ls` shows nothing; `cd /net/server1` mounts |
| **SMB hidden shares** — `\\server\C$` | omitted from enumeration, connects when named |
| **LDAP / SSSD** — `enumerate = false` | `getent passwd` lists nobody, `getent passwd <name>` answers |

The reasons are the ones here too: the keyspace is too large, enumerating is expensive,
or the right to enumerate is separate.

What such a listing cannot do is distinguish itself from an empty one. POSIX can, by
splitting a directory's permissions — read enumerates names, execute resolves one
through it — so with execute alone `ls` fails and "cannot enumerate" is a different
answer from "nothing here":

```sh
$ chmod 111 secret
$ ls secret                 # ls: secret: Permission denied
$ cat secret/known.txt      # hello
$ cat secret/nope.txt       # No such file or directory
```

`list` could refuse the same way — it returns a `Result`. It does not, because a refusal
stops the caller without telling them where to go. A listing that returns one file can
carry the pointer instead: **`_README.md` is there because a successful empty listing
reads as "nothing here"**, and it says how to address one and where an identifier comes
from.

### 2. The path is the whole address (not use `.config`)

The tree holds no "currently selected" state. A writable `.config` carrying `CIK=` would
dissolve §1's awkwardness — `ls edgar/` would be the registrant itself — and is not
there, for four reasons:

- **It needs a write path.** The stores implement three methods and leave the rest
  `EROFS`; what the file holds is a mode, not data.
- **Citations break.** `edgar/submissions/submissions.json` would mean "whatever was
  last written", so anyone following it gets a different company, silently.
- **Two entities cannot be open at once.** NVIDIA CORPORATION and NVIDIA INTERNATIONAL,
  INC. share an address and a jurisdiction, and were separated by Delaware file number
  (`2862596` against `6004882`). One id means comparing from memory.
- **A silently changed id is undetectable.** A wrong identifier answers with somebody
  else's company, not an error. In a path that survives into the report; in a file,
  nothing records when it changed.

### 3. What a listing costs is not ours alone (`FSKit` try made no effect)

Every fetchable document sits in a directory of its own — `facts/facts.json`, not
`facts.json` — because **files cost and directories do not.** A directory read is
followed by an attribute call per entry, and a size is unknowable without fetching, so a
document listed directly is fetched merely to be listed.

When that attribute call happens depends on the transport.

| | FUSE-T / NFS | FUSE-T / FSKit | kernel FUSE (Linux 6.12) |
| --- | --- | --- | --- |
| `find <entity> -type f` | 2 requests | 2 requests | file attribute calls **0** |

NFS READDIR carries no entry type, so `find` must `GETATTR` each name to know whether to
descend — which calls `stat`, which fetches. Kernel FUSE carries it in `getdents64`:
measured on a minimal filesystem of 500 directories with one file each, TTL at zero,
`find` named all 500 files and stat'd none. FSKit was expected to match, since its
enumeration packs an `itemType`; **it did not**, and whether the module drops the type,
enumerates with attributes, or the VFS distrusts it was not determined here.

So the wrapper is unnecessary for walk-safety on kernel FUSE — but anything that really
asks for a size fetches on either, because **needing a fetch to know a size is a
property of the API, not of the mount.** An example that costs tens of gigabytes on one
host and nothing on another is not safe, only lucky.

grf53:
Considering `by-lei` → `entities`, `by-cik` → `filers`

ljhh-0611:
> ## Decision on the proposals raised during discussions
> ### 1. Not listed is not unreachable
> `ls by-lei` names only `_README.md` — millions of entities, no index to enumerate them from — yet `ls by-lei/549300S4KLFTLO7GSQ80/` opens. The listing returns; it simply leaves out what it cannot name. That is an ordinary shape:
> 
> Precedent	
> **autofs non-browsable maps** — `/net/<host>`	`ls` shows nothing; `cd /net/server1` mounts
> **SMB hidden shares** — `\\server\C$`	omitted from enumeration, connects when named
> **LDAP / SSSD** — `enumerate = false`	`getent passwd` lists nobody, `getent passwd <name>` answers
> The reasons are the ones here too: the keyspace is too large, enumerating is expensive, or the right to enumerate is separate.
> 
> What such a listing cannot do is distinguish itself from an empty one. POSIX can, by splitting a directory's permissions — read enumerates names, execute resolves one through it — so with execute alone `ls` fails and "cannot enumerate" is a different answer from "nothing here":
> 
> ```shell
> $ chmod 111 secret
> $ ls secret                 # ls: secret: Permission denied
> $ cat secret/known.txt      # hello
> $ cat secret/nope.txt       # No such file or directory
> ```

Thank you for your research! If this approach is already known thing, then I agree with your decision- hopefully, AI agent will be not confused for using with this way

examples/company_analysis/src/apifs.rs
```diff
@@ -0,0 +1,208 @@
...
+/// Percent-decoding for a path segment, so a value containing `/` can be named.
+pub fn percent_decode(s: &str) -> String {
+    let b = s.as_bytes();
+    let mut out = Vec::with_capacity(b.len());
+    let mut i = 0;
+    while i < b.len() {
+        if b[i] == b'%' && i + 2 < b.len() {
+            if let Ok(v) = u8::from_str_radix(&s[i + 1..i + 3], 16) {
+                out.push(v);
+                i += 3;
```
ljhh-0611, line 176: `&s[i + 1..i + 3]` slices a `&str` by byte offset before the hex check runs, so `%가`, `%a한` and `%🙂` panic on the char boundary, and since the fuse-t callback is `extern "C"` with no `catch_unwind`, that aborts the process and skips `FuseTMount`'s `Drop`, leaving the mount behind.

Recommend: index `b[i + 1]`/`b[i + 2]` instead of slicing `s`.

grf53: Thanks.
it became like this:
where `hex_byte()`converts a hex expression(`a3`) of a byte into a decimal number(163).

```rust
if b[i] == b'%' && i + 2 < b.len() {
    if let Some(v) = hex_byte(b[i + 1], b[i + 2]) {
        out.push(v);
        i += 3;
        continue;
    }
}
```

ljhh-0611 (approved):
👍

PR #447, opened 2026-08-25
[Example] Headhunting, with own domain executable

ljhh-0611:
# A headhunting example, and an app that attaches its own domain executable

Adds `examples/headhunter`: an agent that reads one job posting, searches a pool of 600 candidates, picks the top k, and drafts a cold mail for each.

## Why this example

The examples we have use the tools ailoy ships. This one shows the other direction — **an app extending the agent's environment with commands of its own**, in the vocabulary of its domain rather than in SQL.

That turns out to cost nothing. cortex's delegation materializes a registered name as a symlink on `PATH`, so ailoy's existing `shell` tool calls it with no adapter in between. The app's whole job is implementing `Executable` and naming it.

Two things follow, and they are what makes this worth having as an example rather than a test:

**The agent's world is bounded and you can see the boundary.** It gets one narrow tree and one command. The database is a host path held by the executable, not a file in the tree, so there is no route to the pool that does not go through the command — and `trace.rs` reports any attempt to find one.

**Domain knowledge lives in the command, not in the instruction.** Tenure that overlaps, spellings that drift, the difference between a claim in a headline and evidence in a position description — the commands answer all of it, so the instruction is left with what a tool cannot say for you: how to judge. It is under half the length it was when the agent drove SQL directly.

## How it works

```
$ cargo run -p headhunter -- --jd jd/backend-rust.md --k 3 --out run-1
```

`main.rs` prepares a directory for this run, puts the posting and the schema into `in/`, and mounts that directory as the agent's whole filesystem. It starts the cortex console server, registers `headhunting` against the pool's path, and builds the agent with `system_tools()` — `shell` to call the command, `read` for the posting and the schema, `write` for the artifacts.

```rust
let console = Console::builder()
    .client(StdioClient::new(server)?)
    .mount(Tree(workspace))
    .executables(ExecutableSet::new().register(
        "headhunting", Headhunting::summary(), Headhunting::new(&db)))
    .build().await?;

let mut agent = AgentBuilder::new(&args.model)
    .console(console)
    .system_tools()
    .instruction(prompt::system(args.k))
    .build()?;
```

From there the agent works the posting on its own — one user message in, artifacts out.

```
headhunting search <conditions…>       people matching them, as a table, one line each
headhunting read <id…>                 the people you picked, in full
headhunting distribution <axis>        what the pool holds along that axis
headhunting query <sql>                read-only SQL, for what the others cannot ask
```

It gates on the posting's must-haves with `search`, checks the spelling distribution when the count looks short, ranks within that set with `--mentions`, then `read`s several people at once — some things are only visible with candidates side by side. It writes `00-shortlist.md` and one mail per pick.

The connection opens `SQLITE_OPEN_READ_ONLY`, so `query` cannot write. `main.rs` consumes the stream, collapses the tool traffic into one line of scale per step, records every pool call to `queries.log`, and fails the run if no tool was called at all — a model with no tools attached will imitate the commands in prose and exit cleanly.

## The data

600 candidates, committed as JSON. 65 of them are written by hand rather than generated: 21 carry one of 18 kinds of trap — a skill claimed only in a headline, a spelling the search term does not reach, one person holding two accounts, tenure that double-counts when summed — and 15 stand as controls, each differing from its trap on exactly one axis, on the passing side. A control is qualified by definition, so rejecting one means the pattern was matched instead of the fact.

**The SQLite file is not in the repository.** `python3 sql/load.py` builds it from the committed JSON, and it is the first thing to run. A 2.4MB binary does not diff and regenerates in a second, so what is kept is what it is built from — which is also why `candidates.json` and `ground_truth.json` account for 86% of this diff's line count.

The JSON is committed rather than generated at build time, because a trap only means something against a fixed pool. `data/ground_truth.json` says which of the 600 are planted and what each one tests — it is what a person checks a shortlist's rejections against, and the agent never sees it.

## The input

Four postings, in `jd/`. Each leans on a different part of the pool.

| posting | k | what it tests |
| --- | --- | --- |
| `backend-rust` | 3 | headline against practice, overlapping tenure, rank inversion |
| `backend-seoul-ko` | 3 | the cold mail's language, one name for two people, a Korean-only profile |
| `ml-platform-tokyo` | 12 | one person with two accounts, a certificate alone, a stale profile |
| `blockchain-solidity` | 5 | **picking nobody** — no one qualifies, and saying so is the pass |

`main.rs` copies the posting and the schema into the run's tree under `in/`, and mounts that directory as the agent's whole filesystem — so the posting it reads is inside the tree, not a path on the host.

## What it produced

`run_result/` holds a run of all four, committed so the example can be read without being run — the artifacts, every pool call untruncated, and what went by on screen. The `in/` directory is not kept: it is a copy of `jd/<posting>.md` and `sql/schema.sql` remade on every run.

| posting | k | picked | turns | pool calls | output |
| --- | --- | --- | --- | --- | --- |
| `backend-rust` | 3 | 3 | 42 | 11 | 31.7K |
| `backend-seoul-ko` | 3 | 3 | 31 | 9 | 30.5K |
| `ml-platform-tokyo` | 12 | **6** | 37 | 10 | 40.5K |
| `blockchain-solidity` | 5 | **0** | 30 | 10 | 8.4K |

The two bold figures are the point of those two postings. `ml-platform-tokyo` asks for 12 where fewer than 12 qualify — and among the near-misses is one person holding two accounts, under two spellings of one company name — so the run emitted 6 and said why it could not reach 12. `blockchain-solidity` asks for a skill nobody in the pool has; its shortlist opens with `## Picks` / `(none — see below)` and then lists the ~30 terms it swept for and the four adjacent profiles it read in full before concluding. Filling either number would have been the failure.

`backend-seoul-ko` picked two `ko` profiles and one `en`, and wrote each mail in that candidate's own language rather than the posting's.

**Reading the result is a person's job.** Whether a personalized sentence is grounded in the profile or reads as filler is not something to automate. What each run leaves behind to judge it by is the shortlist's rejections, `queries.log`, and the answer key.

## Running it

```sh
python3 sql/load.py            # writes data/headhunter.db from the committed JSON
cargo run -p headhunter -- --jd jd/backend-rust.md --k 3 --out run-1
```

Credentials come from the repository's `.env`. The default model is `bedrock/global.anthropic.claude-sonnet-5`; `--model anthropic/claude-sonnet-5` reads `ANTHROPIC_API_KEY` instead, which is what the committed run used.

---

https://claude.ai/code/session_01LYNAzB8pqsugaNuEfpeBQy


ljhh-0611:
Ran all four postings against this code. **All four pass with no automatic failures**, and the three command fixes did what they were meant to.

| posting | turns | commands | output | automatic failures |
| --- | --- | --- | --- | --- |
| `backend-rust` | 42 | 26 | 31.7K | none |
| `backend-seoul-ko` | 31 | 16 | 30.5K | none |
| `ml-platform-tokyo` | 37 | 20 | 40.5K | none |
| `blockchain-solidity` | 30 | 14 | 8.4K | none |

## Against the run these fixes came from

`query` is the emergency exit. Two postings had been living in it.

| | before | after |
| --- | --- | --- |
| `ml-platform-tokyo` | 153 turns, 106 commands, **16 `query`** | 37 turns, 20 commands, **0 `query`** |
| `blockchain-solidity` | 71 turns, 68 commands, **58 `query`** | 30 turns, 14 commands, **1 `query`** |

**`ml-platform-tokyo`.** Its must-have is Kubernetes ∧ MLOps ∧ PyTorch, and repeating `--skill` used to keep only the last flag — silently, so a wrong answer looked like a right one. The agent had built the intersection by hand with `comm -12` and verified it against raw SQL. `search` calls went 83 → 16 and the backdoor went to zero.

**`blockchain-solidity`.** Proving nobody qualifies means first knowing what kind of people the pool holds, and I had removed the `title` axis as unused. 58 raw queries were doing that work. Restoring `title` and `company` moved it back into the command: `distribution` 1 → 5, `query` 58 → 1, and the run costs a third of what it did.

**`backend-seoul-ko`** now picks `ko`, `ko`, `en` and writes each mail in its own candidate's language. The gate used to require a Korean profile — something the posting never asks for — which made the test passable by writing Korean to everyone and made one control structurally impossible.

One run died mid-way on `connection reset` from the API and was rerun; the numbers above are from the clean run.

grf53 (approved):

PR #448, opened 2026-09-04
Bedrock wire

grf53:
## Description
Add Amazon Bedrock as a language-model backend over the [Converse API](https://docs.aws.amazon.com/ko_kr/bedrock/latest/APIReference/API_runtime_Converse.html), so any [model family](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html) Bedrock serves is reachable through one `bedrock/*` registration.

## Why not only `Provider` but need `Schema`?
Based on verification using Anthropic (only), Bedrock’s schema compatibility API is not exactly the same as the target. Therefore, it was necessary to determine whether the Marshal/Unmarshal implementation was Bedrock or something else; this required structural changes to elements such as ProviderElem, and the processing for these changes had to be handled separately.

Since there are models that operate exclusively using the Converse API provided by Bedrock, a schema supporting the Converse API had to be added anyway. As this also enables support for models that previously used existing schemas, such as OpenAI and Anthropic, we decided to include a schema specifically for Bedrock here.

### Bedrock (Converse) wiring
- New `LangModelAPISchema::Bedrock` with `BedrockMarshal` / `BedrockUnmarshal` in `lang_model/impl/api/bedrock.rs`.
  - The provider `url` is the runtime base (`https://bedrock-runtime.<region>.amazonaws.com`); the model id and action (`converse` / `converse-stream`) are appended per request.
  - the API key is sent as a bearer token.
  - `EventStream` is used instead of `Sse`.
- `LangModelProvider::bedrock(region, key)` constructor; the default registry adds `bedrock/*` from `AWS_BEARER_TOKEN_BEDROCK` with the region from `AWS_REGION`, then `AWS_DEFAULT_REGION`, defaulting to `us-east-1`. Blank keys are no longer registered.

### Stream framing module
- `Framing` (`Sse` / `EventStream`) plus the SSE helpers move out of `rt.rs` into `lang_model/impl/framing/{mod,sse,eventstream}.rs`; `run_stream` picks the framing by schema and drains through it. Move-only for the SSE path.

## Tests
- Converse envelope: URL, bearer header, `converse-stream` action and `accept` header, no `model` body field (`envelope_targets_converse_with_bearer_token`)
- Message mapping: tool results merged into one `user` turn, empty text dropped, image format from MIME, thinking replayed only after the last user turn
- Response and stream unmarshal: tool call accumulation across `contentBlockDelta` events, `metadata` usage, reasoning text and signature
- Pre-send guards: `response_format` and URL images rejected (`validate_rejects_bedrock_only_gaps`, `marshal_request_bedrock_envelope_and_guard`)
- Event stream decoder: CRC reference vector, frames split across chunks, corrupt CRC rejected, `chunk` base64 unwrap, exception frame message
- Live Converse round trips (text, tool call, stream) gated on `AWS_BEARER_TOKEN_BEDROCK`
- Suite: `cargo test --lib lang_model` 89 passed

## Notes
- `LangModelAPISchema::Anthropic` cannot be pointed at a Bedrock endpoint. Bedrock's `InvokeModel` moves the model id into the path, the key into a bearer token and the API version into the body, so the Anthropic marshal would need to know it is on Bedrock; the Converse schema is the only supported way to reach Bedrock-hosted Claude models. An `InvokeModel`-based `BedrockAnthropic` schema exists on the local tag `bedrock-anthropic-wire` if Messages-API-only features turn out to be needed.
- The event stream decoder's `chunk` unwrap (the `InvokeModelWithResponseStream` shape) has no caller on the Converse path; it is unit-tested and kept for that follow-up.


src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+            url: runtime_url(region.as_ref()),
+            api_key: Some(api_key),
+        }
+    }
+}
+
+/// Runtime base for a region; the registered `url` for both Bedrock schemas.
+fn runtime_url(region: &str) -> Url {
+    Url::parse(&format!("https://bedrock-runtime.{region}.amazonaws.com")).unwrap()
+}
```
selloriwoo: `region` is interpolated into the host and the parse is unwrapped, which fails two ways: `AWS_REGION="us east 1"` panics with nothing naming the variable, and `a@evil.com` doesn't panic at all — it parses into host `evil.com.amazonaws.com`, bearer token included. Could we validate the region and return `Result`, the way `chat_completion` already does?

```suggestion
    ///
    /// Fails on a region that is not a syntactically valid region id, since the
    /// region reaches the wire as part of the host — see [`runtime_url`].
    pub fn bedrock(
        region: impl AsRef<str>,
        api_key: String,
    ) -> anyhow::Result<LangModelProviderElem> {
        Ok(LangModelProviderElem::API {
            schema: LangModelAPISchema::Bedrock,
            url: runtime_url(region.as_ref())?,
            api_key: Some(api_key),
        })
    }
}

/// Runtime base for a region; the registered `url` for both Bedrock schemas.
///
/// The region is interpolated into the host, so it is checked rather than
/// trusted, and the two shapes that matter fail differently. `"us east 1"` does
/// not parse, which unwrapping would surface as an anonymous parse error with
/// nothing naming the variable to fix. `"a@evil.com"` parses *successfully*,
/// into the host `evil.com.amazonaws.com`, which would send the bearer token
/// there — a fallible parse alone would not catch that one.
///
/// Region ids are alphanumerics and hyphens, so anything else is rejected
/// before either outcome. Case is not checked because the URL parser lowercases
/// the host, which makes `US-EAST-1` work today.
fn runtime_url(region: &str) -> anyhow::Result<Url> {
    if region.is_empty()
        || !region
            .bytes()
            .all(|b| b.is_ascii_alphanumeric() || b == b'-')
    {
        bail!("invalid AWS region {region:?}");
    }
    Ok(Url::parse(&format!(
        "https://bedrock-runtime.{region}.amazonaws.com"
    ))?)
}
```

Only this constructor needs it — the others parse a literal, so their `unwrap` can't fail. Six existing call sites take a `.unwrap()`: five in this file's tests and `marshal_request_bedrock_envelope_and_guard` in `rt.rs`.

src/lang_model/provider.rs
```diff
@@ -53,27 +56,44 @@ pub struct LangModelProvider {
...
+        if let Some(key) = env_key("KIMI_API_KEY") {
             p.insert("moonshotai/*".into(), Self::kimi(key));
         }
+        if let Some(key) = env_key("AWS_BEARER_TOKEN_BEDROCK") {
+            // Same precedence the AWS SDKs use, so a shell already configured
+            // for AWS needs no extra variable.
+            let region = env_key("AWS_REGION")
+                .or_else(|| env_key("AWS_DEFAULT_REGION"))
+                .unwrap_or_else(|| "us-east-1".to_string());
+            p.insert("bedrock/*".into(), Self::bedrock(region, key));
```
selloriwoo: With `bedrock()` fallible, could we `expect` here with the variable named? Only a caller who set `AWS_BEARER_TOKEN_BEDROCK` reaches this line, so failing immediately hits the misconfigured process rather than registering a `Url` that points somewhere else.

```suggestion
            let elem = Self::bedrock(region, key)
                .expect("AWS_REGION / AWS_DEFAULT_REGION is not a valid AWS region id");
            p.insert("bedrock/*".into(), elem);
```

selloriwoo (commented):
Two on region handling in `runtime_url`: the parse of an interpolated value is unwrapped, so a malformed `AWS_REGION` either panics with nothing naming the cause, or — with `a@evil.com` — parses fine and aims the request at another host. Validating the region covers both and lets the failure name the variable to fix.

Follow-ups, not blocking: a truncated event stream is delivered as a successful turn (`Framing::flush` drops the partial frame and `run_stream` synthesizes `Stop`), and a turn whose parts all filter out marshals to `content: []`, which Converse rejects.

Applied both suggestions locally: 94 passed, clippy/fmt clean.

nuri-yoo (commented):
Read at 0e856e2 on develop 2b795e0. `cargo test --lib lang_model`: 101 passed, 0 failed, 7 ignored (the live tests; no keys here). `cargo clippy --lib --tests` clean. rustfmt clean on bedrock.rs, framing/*, rt.rs, provider.rs. Nothing below overlaps with selloriwoo's two threads or the two follow-ups in that review body.

Inline: the `response_format` guard, the `stopReason` map, the image `format` value, and a few prose leftovers from the two-schema version. Four things with no diff line to anchor on:

**PR body vs head.** The Tests bullet "`chunk` base64 unwrap" and the Notes paragraph "The event stream decoder's `chunk` unwrap ... is unit-tested and kept for that follow-up" describe code that is not on this branch: `framing/eventstream.rs` has no base64 handling, and `chunk` only appears as a header value in `skips_non_string_headers`. Presumably it went to the `bedrock-anthropic-wire` tag with the InvokeModel schema. Also "89 passed": on this head I count 101, probably from the develop rebase, but the number in the body is stale either way.

**`env_key` (provider.rs:64).** This is a behaviour change for every provider (blank keys skipped) and the body mentions it in one clause under the Bedrock section; worth its own line. One latent nit: the check trims but the registered value does not, so `OPENAI_API_KEY=" sk-... "` passes the blank check and reaches the wire as `Bearer  sk-... `. dotenvy trims unquoted values, so this only bites a shell export with literal spaces. Either trim the returned value or leave it; flagging so the choice is deliberate.

**SSE prose after the framing split.** The second commit makes SSE one of two framings, but these still describe it as the only one: rt.rs:29 "requests a streaming (SSE) response", rt.rs:116 "requests an SSE response", rt.rs:120 "after the terminal SSE event", api/mod.rs:57 "per-event (SSE) unmarshaling", api/mod.rs:60 "for one SSE event".

**README.md:183-189** "Supported AI Models" lists OpenAI / Anthropic / Google / xAI. DeepSeek and Kimi are already missing there, so this is either a Bedrock row here plus the two missing ones, or a separate docs PR. Not a condition for this one.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+    }
+    header
+}
+
+/// Rejects requests Bedrock would 400 on, before anything is sent. Marshals
+/// return a bare `Value`, so this is where a Bedrock-only limitation becomes an
+/// error instead of a silently dropped field.
+pub(in crate::lang_model) fn validate_request(req: &LangModelRequest<'_>) -> anyhow::Result<()> {
+    if req.options.response_format.is_some() {
+        bail!("Bedrock does not support response_format (structured output)");
```
nuri-yoo: Converse does accept a JSON-schema output constraint, so this message states a limit Bedrock does not have. The Converse request syntax has `outputConfig.textFormat`; `OutputFormat.type` valid values are `json_schema`, and `JsonSchemaDefinition.schema` is "Type: String, Required: Yes" with `name` and `description` optional (API_runtime_OutputFormat, API_runtime_JsonSchemaDefinition). The user guide says "For Converse APIs, use the `outputConfig.textFormat` request field." Spec only on my side: I have no Bedrock key to send one.

What it costs: `AgentBuilder::response_format` (agent/builder.rs:184) works with `anthropic/claude-sonnet-5` (anthropic.rs:249 maps it to `output_config.format`) and fails before the request with `bedrock/global.anthropic.claude-sonnet-5`. Since the Notes say Converse is the only route to Bedrock-hosted Claude, the same model loses that feature by moving providers.

Two ways to settle it. (a) Map it: implement `ResponseSchemaMarshal` for `BedrockMarshal` and send `{"outputConfig": {"textFormat": {"type": "json_schema", "structure": {"jsonSchema": {"schema": serde_json::to_string(&wire_schema)}}}}}`. The default `additionalProperties: false` insertion fits, since Bedrock rejects any other value for it. Trade-off: Bedrock's accepted subset is narrower than Anthropic's (no `minimum`/`maximum`/`minLength`/`maxLength`, no recursive schemas, per the structured-output guide), so a schema that works on `anthropic/*` can 400 here. (b) Keep the guard for now but word it as not implemented for this schema rather than unsupported by Bedrock, and list it in the PR Notes next to the InvokeModel follow-up.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+// Bedrock throttling is a 429 `ThrottlingException` and always transient.
+impl super::QuotaClassifier for BedrockUnmarshal {}
+
+impl BedrockUnmarshal {
+    fn parse_finish_reason(reason: &str) -> FinishReason {
+        match reason {
+            "end_turn" | "stop_sequence" => FinishReason::Stop {},
+            "max_tokens" => FinishReason::Length {},
+            "tool_use" => FinishReason::ToolCall {},
+            other => FinishReason::Refusal {
```
nuri-yoo, line 472: Documented `stopReason` values are `end_turn | tool_use | max_tokens | stop_sequence | guardrail_intervened | content_filtered | malformed_model_output | malformed_tool_use | model_context_window_exceeded` (API_runtime_Converse, Response Elements). The `other => Refusal` fallback is the same convention as anthropic.rs:288 and is right for the guardrail/filter pair. `model_context_window_exceeded` is the one that lands wrong: it is a length stop, so `Length {}`. Latent today: nothing outside `message.rs` and the api marshals branches on `Length` vs `Refusal`, so only a caller reading `finish_reason` sees the difference.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+    Ok(())
+}
+
+/// Converse request marshal. Model-agnostic, so nothing here depends on the
+/// model family behind the id.
+#[derive(Clone, Debug, Default)]
+pub struct BedrockMarshal;
+
+/// Converse image `format` from a MIME type; Converse accepts png/jpeg/gif/webp.
+fn image_format(mime_type: &str) -> &str {
```
nuri-yoo, line 215: `ImageBlock.format` valid values are `png | jpeg | gif | webp` (API_runtime_ImageBlock). `image_format` strips the prefix and passes anything else through, and `Part::image_embedded` (part.rs:112) does not check the MIME type either, so `image/svg+xml` marshals as `"format": "svg+xml"` and is a 400 from Bedrock, which is the class the doc on `validate_request` says this module catches before sending. Option: whitelist the four (plus the `jpg` alias) in `validate_request`, where the image parts are already walked for the URL check. Trade-off: if Converse adds a format, the whitelist blocks it before Bedrock would accept it; letting Bedrock's own 400 through is the alternative.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
+//! Amazon Bedrock: the endpoint shared by every schema served there, plus the
```
nuri-yoo: With the InvokeModel schema on the tag, this branch registers one Bedrock schema. "every schema served there" here, and "both Bedrock schemas" at line 33 (also carried into the suggestion in the thread above; noted there), describe the two-schema version.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+    },
+    to_value,
+    tool::ToolDesc,
+};
+
+impl LangModelProvider {
+    /// Bedrock over the Converse API, which speaks to every model family in
+    /// the region with one body format.
+    ///
+    /// `model` must be an id Bedrock accepts for on-demand throughput, e.g. the
```
nuri-yoo: `model` reads as a parameter, but the constructor takes only `region` and `api_key`. If it means the id after `bedrock/` in the registered name, saying so avoids the hunt.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+}
+
+/// Runtime base for a region; the registered `url` for both Bedrock schemas.
+fn runtime_url(region: &str) -> Url {
+    Url::parse(&format!("https://bedrock-runtime.{region}.amazonaws.com")).unwrap()
+}
+
+/// `<base>/model/<model>/<action>`. `/` in the model id (application
+/// inference-profile ARNs) is escaped so it stays one path segment.
+pub(super) fn model_url(base: &Url, model: &str, action: &str) -> String {
```
nuri-yoo: `model_url` and `headers` (line 50) are `pub(super)` with no caller outside this file. If that is a hold for the tagged schema, a word saying so; otherwise private.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,926 @@
...
+        }
+        let out = acc.finish().unwrap();
+        assert_eq!(out.finish_reason, FinishReason::Stop {});
+        assert_eq!(out.message.contents[0].as_text(), Some("Hello"));
+        assert_eq!(out.message.thinking.as_deref(), Some("hmm"));
+        assert_eq!(out.message.signature.as_deref(), Some("sig"));
+    }
+
+    #[test]
+    fn constructors_register_the_runtime_base() {
```
nuri-yoo: One constructor on this branch; `constructors_register_...` is plural from the two-schema version. This is also one of the five call sites that gets `.unwrap()` when the region suggestion lands, so the rename may fit the same pass.

src/lang_model/impl/api/bedrock.rs
nuri-yoo: One line to carry along when applying this: the suggestion keeps `/// Runtime base for a region; the registered `url` for both Bedrock schemas.` from the current code, and this branch has one Bedrock schema (the InvokeModel one is on the `bedrock-anthropic-wire` tag). Applied both suggestions locally on 0e856e2 with the six `.unwrap()` call sites: 101 passed, clippy and rustfmt clean.

grf53: Thought the suggestion is not wanted a bit, cuz I thought all the other constructors returns no `Result` but `LangModelProvider::chat_completion()`.

The point is existence of input as a free string that can be fail(`url` in `chat_completion()` and `region` in `bedrock()`).

There are two options:
1. `Result`-returning `bedrock()`
2. Ensure the `region` parameter does not fail (= Make `region` an enum)

2 gives us a few more advantages, so I chose it.

src/lang_model/provider.rs
grf53: `bedrock()` is not fallible with more durable structure. Plz check https://github.com/brekkylab/ailoy/pull/448#discussion_r3965028119.

src/lang_model/impl/api/bedrock.rs
grf53: I missed something there. I’ve adjusted it to support structured output, just like the other schemas and providers.

src/lang_model/impl/api/bedrock.rs
grf53: In 9d1c634d9964a2f4246a7b8ad3bd90ebcee70c33, I added a comment specifying that `model_context_window_exceeded` is parsed as `Length`, while other cases are categorized as `Refusal`.
I did not include a test for this part as it is extremely straightforward.

src/lang_model/impl/api/bedrock.rs
grf53: Since other providers do not handle MIME types separately—meaning such issues cannot be prevented beforehand—it is natural for the blocking to occur at the API level with a 400 error.
This discrepancy appears to arise because Bedrock does not accept standard MIME types, requiring instead specific, predefined formats. I believe it is best to keep the conversion logic focused solely on formatting, avoiding the need for mechanisms like allowlists.
I am simply revising the wording of a comment to prevent potential misunderstanding.

src/lang_model/impl/api/bedrock.rs
selloriwoo: The enum is more explicit and easier to maintain when something goes wrong. LGTM.

src/lang_model/provider.rs
selloriwoo: LGTM.

src/lang_model/impl/api/bedrock.rs
```diff
@@ -0,0 +1,1147 @@
...
+    pub fn bedrock(region: BedrockRegion, api_key: String) -> LangModelProviderElem {
+        LangModelProviderElem::API {
+            schema: LangModelAPISchema::Bedrock,
+            url: region.runtime_url(),
+            api_key: Some(api_key),
+        }
+    }
+}
+
+/// AWS regions with a Bedrock runtime. The id is the host label in
```
nuri-yoo: The 35 ids here are exactly the "Amazon Bedrock control plane APIs" table on the endpoints page (docs.aws.amazon.com/general/latest/gr/bedrock.html). The "Amazon Bedrock runtime APIs" table on the same page has 22, and these 13 appear only in the former: `us-west-1`, `ap-east-2`, `ap-southeast-3`/`4`/`5`/`6`/`7`, `ca-west-1`, `il-central-1`, `me-central-1`, `me-south-1`, `mx-central-1`, `af-south-1`. So "regions with a Bedrock runtime" names the wrong table. Spec only: all 35 `bedrock-runtime.<id>.amazonaws.com` hosts resolve in DNS today, and that page can lag actual launches, so I would not drop entries on its word alone; it is the doc sentence that needs to match its source.

One design note while the type is new: without `#[non_exhaustive]`, each region added later is a semver break for anyone matching on this enum exhaustively, and a closed set means a user in a new region waits for a release (the `Default` path then logs "unsupported AWS region" and skips `bedrock/*`). Both are fine as decisions; noting so they are made on purpose.

selloriwoo (approved):

src/lang_model/impl/api/bedrock.rs
grf53: already covered.

src/lang_model/impl/api/bedrock.rs
grf53: removed. thanks.

src/lang_model/impl/api/bedrock.rs
grf53: It must be a residue from the first version of the constructor. thanks.

src/lang_model/impl/api/bedrock.rs
grf53: revised.

nuri-yoo (commented):
LGTM

src/lang_model/impl/api/bedrock.rs
grf53: 6995f422d5e66ebd558a485227d4d9f070957b94
1. The expression is edited.
2. That was on purpose, but not strongly insisted. I decided to consider that as a known issue, cuz quite a messy handler is need to open it up for free configuration. If there is any other opinion, I am open to follow that.

nuri-yoo (approved):

